PR #32, opened 2017-01-24
ipfs-cluster-ctl improvements

### LINKED ISSUE #21: CLI tooling (flags)
* ipfscluster-server
  * See if it is possible to have `-config PATH` and not `-config string` (not possible with vanilla flag I think)
  * Study use of https://github.com/urfave/cli
  * Make `ipfscluster-server -init` a subcommand too `ipfscluster-server init`
  * Add -f to overwrite config file rather than asking the user to do it

* ipfscluster
  * Longer general help
  * See also about urfave/cli
  * Probably should be able to launch the server
  * `ipfscluster id`

### LINKED ISSUE #22: Configuration files location and format
* Study configuration location change to `$(pwd)/.ipfs-cluster`

I dont like this. It's relative, prone to user error, departs from IPFS convention. If anything configurations live in `/etc/` in any standard system. Local configs are usually in `.config/<app-name>` these days.

* See if others care about using Go CamelCased keys, as IPFS does, for configuration entries.

I prefer the javascript style because that's the usual with JSON. Also related to how API responses. Look.

* Handle panics/errors when configuration is really invalid (empty node ID etc).

* Rename `ipfs_port` to `ipfs_node_port` and so on

* Use multiaddress format like: 

      ```
      "ipfs_cluster_api": "/ip4/127.0.0.1/tcp/9095/http",
      "ipfs_node_api":    "/ip4/127.0.0.1/tcp/5001/http",
      ```

hsanjuan:
Fixes #21 

* Use urfave/cli which handles nicer formatting and help in commands
* Add an /ID endpoint to cluster and support from it

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,9 +39,9 @@
     },
     {
       "author": "urfave",
-      "hash": "QmTMU3gTfgpBK4dQ8pSPdH4DpCR2NUNWYTRoXzz9X4ecFg",
+      "hash": "Qmc1AtgBdoUHP8oYSqU81NRYdzohmF45t5XNwVMvhCxsBA",
       "name": "cli",
-      "version": "2017.1.24"
+      "version": "1.19.1"
     }
   ],
   "gxVersion": "0.10.0",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,12 @@
       "hash": "QmZvXudUXTc2G3o2SYPU7uQX1gzbeQkAZXJ6bnN8C4rcXE",
       "name": "go-libp2p-rpc",
       "version": "0.0.10"
+    },
+    {
+      "author": "hsanjuan",
+      "hash": "QmSNNKmo6LtoiSLjaVJxi76QFRfKJvJUME96EmtLEtFPdq",
+      "name": "cli",
+      "version": "2017.1.24"
     }
   ],
   "gxVersion": "0.10.0",
--- a/package.json
+++ b/package.json
@@ -38,8 +38,8 @@
       "version": "0.0.10"
     },
     {
-      "author": "hsanjuan",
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -36,6 +36,12 @@
       "hash": "QmZvXudUXTc2G3o2SYPU7uQX1gzbeQkAZXJ6bnN8C4rcXE",
       "name": "go-libp2p-rpc",
       "version": "0.0.10"
+    },
+    {
+      "author": "hsanjuan",
+      "hash": "QmSNNKmo6LtoiSLjaVJxi76QFRfKJvJUME96EmtLEtFPdq",
+      "name": "cli",
+      "version": "2017.1.24"
```
mateon1: I'm not sure this should be using Y.M.D instead of semver. Is there any particular reason?

mateon1: Just looked, and the upstream package is using semver, but for some reason the gx package.json uses the date.

hsanjuan: Yes, I can't find a better way to allow the package.json description of a package to take me back to the original time a package was gx'ed so it can be rebuilt again (shall ipfs ever lose the hash).

Actually I should set author to the original one. If I have an author, a project name and a date I can at least know when more or less this external dependency was packed in case I ever need to rebuild it or try to produce something similar.

hsanjuan: (also the fact that gx compains if the version field is not a SemVer)

mateon1: How about using versions matching the release tags on https://github.com/urfave/cli?

hsanjuan: Yes indeed that's better. These were actually my excuses for other projects which do not provide release tags and forgot to check this time. I'll fix it.

PR #38, opened 2017-02-02
Fix #24: Add Cluster.Join() operation to join an existing cluster

### LINKED ISSUE #24: Autodiscovery - autosetup of cluster members

Note the trickiness of this:

hsanjuan:
Join() takes a peer's multiaddress. It then sends a "PeerAdd()"
RPC request with that peer and the local multiaddress by which
the remote peer is reached. The the remote peer uses the PeerAdd()
functionality to a) Tell the rest of the cluster about the new member
and b) send us the list of cluster members.

Join() has no REST API support, but it is provided instead via
"ipfs-cluster-service --join". A "--leave" flag has been included too,
by which a peer triggers "PeerRemove()" on itself when shutting down.

This allows a node to join an existing cluster and leave it by the
end of its life.

#24 

License: MIT

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -120,7 +120,7 @@ func NewCluster(cfg *Config, api API, ipfs IPFSConnector, state State, tracker P
 // Ready returns a channel which signals when this peer is
 // fully initialized (including consensus).
 func (c *Cluster) Ready() <-chan struct{} {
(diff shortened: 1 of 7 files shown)


hsanjuan:
This had some shortcomings and bugs which are addressed in #39

PR #39, opened 2017-02-03
Fix #24: Add Cluster.Join() operation to join an existing cluster

### LINKED ISSUE #24: Autodiscovery - autosetup of cluster members

hsanjuan:
This reworks PeerAdd() and adds a Join() operation.

Join(multiaddress) uses the given bootstrap multiaddress to register
itself with it and fetch its list of cluster peers. Then it proceeds
to register itself with every peer and fetch their lists of cluster peers.
If a cluster peer is not known, then we register ourselves with it again
and so on.

PeerAdd() is now "simply" registering the new node and sending a
Join() request to it. The new node then will fetch the list
of peers and register itself with everyone like explained above.

#24 

License: MIT

### DIFF
--- a/architecture.md
+++ b/architecture.md
@@ -88,15 +88,23 @@ This sections explains how some things work in Cluster.
 
 * The `RESTAPI` component receives a `PeerAdd` request on the respective endpoint. It makes a `PeerAdd` RPC request.
 * The local RPC server receives it and calls `PeerAdd` method in the main `Cluster` component.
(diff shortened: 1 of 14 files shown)


hsanjuan:
Should be good to go. I did manual testing on several nodes and also between nodes in different networks. Letting it rest til monday.

hsanjuan:
After giving it thought over the weekend. #40 has an more solid implementation. Closing this.

PR #40, opened 2017-02-07
Fix #24: Auto-join and auto-leave operations for Cluster

### LINKED ISSUE #24: Autodiscovery - autosetup of cluster members

hsanjuan:
This is the third implementation attempt. This time, rather than
broadcasting PeerAdd/Join requests to the whole cluster, we use the
consensus log to broadcast new peers joining.

This makes it easier to recover from errors and to know who exactly
is member of a cluster and who is not. The consensus is, after all,
meant to agree on things, and the list of cluster peers is something
everyone has to agree on.

Raft itself uses a special log operation to maintain the peer set.

The tests are almost unchanged from the previous attempts so it should
be the same, except it doesn't seem possible to bootstrap a bunch of nodes
at the same time using different bootstrap nodes. It works when using
the same. I'm not sure this worked before either, but the code is
simpler than recursively contacting peers, and scales better for
larger clusters.

Nodes have to be careful about joining clusters while keeping the state
from a different cluster (disjoint logs). This may cause problems with
Raft.

#24

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,28 +81,37 @@ $ ipfs-cluster-service -init
 
 The configuration will be placed in `~/.ipfs-cluster/service.json` by default.
 
-You can add the multiaddresses for the other cluster peers the `cluster_peers` variable. For example, here is a valid configuration for a cluster of 4 peers:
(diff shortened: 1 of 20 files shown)


hsanjuan:
Will let it rest tonight and merge tomorrow if I'm still happy.

Issue #41, opened 2017-02-07
Pins should support a replication factor


hsanjuan:
A Pin should not need to be pinned in every cluster member. We should be able to say that a pin needs to be pinned in 2, 3 cluster members.

We will start with a general replication factor for all pins, then maybe transition to replication factor per-pin.

These are thoughts for the first approach.

Replication factor `-1` means Pin everywhere. If replication factor is larger than the number of clusters then it is assumed to be as large.

### Pinning

We need a `PeerMonitor` component which is able to decide, when a pin request arrives, which peer comes next. The decision should be based on pluggable modules: for a start, we will start with one which attempts to evenly distribute the pins, although it should easily support other metrics like disk space etc.

Every commit log entry asking to Pin something must be tagged with the peers which are in charge. The Pin Tracker will receive the task and if it is tagged itself on a pin it will pin. Alternatively it will store the pin and mark it as `remote`.

If the `PinTracker` receives a Pin which is already known, it should unpin if it is no longer tagged among the hosts that are in charge of pinning. Somewhere in the pipeline we probably should detect re-pinnings and not change pinning peers stupidly.

### Unpinning

Unpinning works as usual removing the pin only where it is pinned.

### Re-pinning on peer failure

The peer monitor should detect hosts which are down (or hosts whose `ipfs` daemon is down).Upon a certain time threshold ( say 5 mins, configurable). It should grep the status for pins assigned to that host and re-pin them to new hosts.

The peer monitor should also receive updates from the peer manager and make sure that there are no pins assigned to hosts that are no longer in the cluster.

For the moment there is no re-rebalancing when a node comes back online.

This assumes there is a single peer monitor for the whole cluster. While monitoring the local ipfs daemon could be done by each peer (and triggering rebalances for that), if all nodes watch eachothers this will cause havoc when triggering rebalances. The Raft cluster leader should probably be in charge then. But this conflicts with being completely abstracted from the consensus algorithm below. If we had a non-leader-based consensus we could assume a distributed lottery to select someone. It makes no sense to re-implement code to choose a peer from the cluster when Raft has it all. Also, running the rebalance process in the Raft leader saves redirection for every new pin request.

### UX

We need to attack `ipfs-cluster-ctl` to provide more human readable outputs as the API formats are more stable. `status` should probably show succinctly which pins are underreplicated or peers in error, 1 line per pin.

20zinnm:
[...]

Issue #45, opened 2017-02-14
Cluster should detect peer failure and re-allocate affected content


hsanjuan:
Currently Cluster can allocate content to a number of peers but will not detect failures and re-allocate in that case.

Allocation is based on metrics which are regularly pushed to the Leader. If the last metric from a peer is expired or invalid then the peer is not considered as an available allocation when pinning. When re-pinning content, this situation is also detected and a new allocation will be found, so that part is done. 

The idea is then give the PeerMonitor the task of producing Alerts on a channel on which the main component is listening. When the PeerMonitor detects that a component is down (because, i.e. last metric has expired), it sends an alert. Cluster will then find which Cids are allocated to the problematic peer and re-trigger Pin operations for each.

This implies PeerMonitors should be made aware of current clusterPeers (or be aware themselves with the RPC API to the pinManager).

PR #46, opened 2017-02-14
Add replication factor support

hsanjuan:
New PeerManager, Allocator, Informer components have been added along
with a new "replication_factor" configuration option.

First, cluster peers collect and push metrics (Informer) to the Cluster
leader regularly. The Informer is an interface that can be implemented
in custom wayts to support custom metrics.

Second, on a pin operation, using the information from the collected metrics,
an Allocator can provide a list of preferences as to where the new pin
should be assigned. The Allocator is an interface allowing to provide
different allocation strategies.

Both Allocator and Informer are Cluster Componenets, and have access
to the RPC API.

The allocations are kept in the shared state. Cluster peer failure
detection is still missing and re-allocation is still missing (#45), although manually
re-pinning something when a node is down/metrics missing does re-allocate
the pin somewhere else.

Plus: The ipfs-cluster-ctl tool has nicer text output now.

Fixes: #41

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,7 @@ install:
 - go get github.com/mattn/goveralls
 - make deps
 script:
-- make service
-- make ctl
-- ./coverage.sh
+- make service && make ctl && ./coverage.sh
 env:
   global:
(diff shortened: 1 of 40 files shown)


hsanjuan:
The architecture and readme docs need an update. Will do that out of this PR.

PR #47, opened 2017-02-15
Small fixes

hsanjuan:
This fixes two small issues I watched during the demo the other day, mostly cosmetic:

  * duplicate listen multiaddresses in the peer information
  * timestamps shown in different timezones in the pin status information

Just checking that tests pass


hsanjuan:
Error is caused by not waiting long enough and is not critical. will fix in the future.

PR #52, opened 2017-02-28
Issue #45: Detect expired metrics and trigger re-pins

hsanjuan:
An initial, simple approach to this. The PeerMonitor will
check its metrics, compare to the current set of peers and put
an alert in the alerts channel if the metrics for a peer have expired.

Cluster reads this channel looking for "ping" alerts. The leader
(because that's the monitor here metrics are pushed from everywhere)
then reads the alerts and triggers repins in all the Cids allocated to
a given peer.

License: MIT

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -191,6 +191,23 @@ func (c *Cluster) stateSyncWatcher() {
 	}
 }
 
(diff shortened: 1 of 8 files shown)


hsanjuan:
note: s3 outage affecting travis

PR #61, opened 2017-03-10
Add docker and dist.ipfs.io install methods to README

hsanjuan:
As the time of this PR, the install locations are not yet ready, but will be when I merge.

@brainstorm how do the docker instructions look to you? Anything you would improve?

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,9 @@
 
 - [Maintainers and Roadmap](#maintainers-and-roadmap)
 - [Install](#install)
+  - [Pre-compiled binaries](#pre-compiled-binaries)
+  - [Docker](#docker)
+  - [Install from sources](#install-from-sources)
 - [Usage](#usage)
   - [`ipfs-cluster-service`](#ipfs-cluster-service)
   - [`ipfs-cluster-ctl`](#ipfs-cluster-ctl)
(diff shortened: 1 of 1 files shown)


brainstorm:
Looks great! I'm not sure if `/data` is very [LFS-friendly](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) as a mountpoint. I would use `/srv` instead since for instance the Amazon ECS AMIs and many other distros already have that folder created, so we save provisioning/deployment steps.

hsanjuan:
@brainstorm that's a good point, but I'm following the road set by the ipfs docker container (which puts its configuration and repo stuff in /data/ipfs), so it seemed natural to put cluster data there too, despite it's all against LFS.

And thinking of it, this doesn't belong in `/srv`... if we were to be strict, cluster configuration would go in `/etc/` and the data somewhere in `/var/lib`, since, after all, we don't serve anything (or better, `~/.config/ipfs-cluster/` and `~/.local/share/ipfs-cluster`)

But overall the IPFS ecosystem does not adhere very well to these things and I'm always caught between being consistent with Linux tradition or with IPFS/go-ipfs ways. And up to now I've been going with the latter because it is what is familiar to IPFS users.

brainstorm:
Good analysis and points, let's go the ipfs way then :+1: 

I am now trying to figure out the best mountpoint strategy for several docker instances and containers on ECS... currently looking at [REX-RAYS](https://github.com/codedellemc/rexray) for scalable volume management/deployment (which then IPFS data folders will mount to).

I'm pointing this out here since I'm [trying to catchup with how you guys are planning to do that](https://github.com/ipfs/test-lab/issues/5) in order to not reinvent the wheel :-S

@ohofmann

hsanjuan:
@brainstorm we have https://hub.docker.com/r/ipfs/ipfs-cluster/ now

PR #80, opened 2017-03-29
Fix #57: Hijack /add requests and pin items in cluster

### LINKED ISSUE #57: IPFS proxy should handle /add requets and pin the added items in cluster
It would be great to have option for auto pinning objects that are added/pinned with ipfs daemon.

Steps: 
- ipfs add
- ipfs pin
- ipfs-cluster auto pins and therefore all items are propagated automatically

hsanjuan:
Should work, but will re-look again. I'm tired and there are many small things to take care of.

### DIFF
--- a/ipfsconn/ipfshttp/ipfshttp.go
+++ b/ipfsconn/ipfshttp/ipfshttp.go
@@ -3,6 +3,7 @@
 package ipfshttp
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -11,6 +12,8 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"net/url"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
@@ -34,10 +37,12 @@ var ConnectSwarmsDelay = 7 * time.Second
 
 // IPFS Proxy settings
 var (
-	// maximum duration before timing out read of the request
-	IPFSProxyServerReadTimeout = 5 * time.Second
+	// maximum duration before timing out reading a full request
+	IPFSProxyServerReadTimeout = 10 * time.Minute
+	// maximum duration before timing out reading the headers of a request
+	IPFSProxyServerReadHeaderTimeout = 5 * time.Second
 	// maximum duration before timing out write of the response
-	IPFSProxyServerWriteTimeout = 10 * time.Second
+	IPFSProxyServerWriteTimeout = 10 * time.Minute
 	// server-side the amount of time a Keep-Alive connection will be
 	// kept idle before being reused
 	IPFSProxyServerIdleTimeout = 60 * time.Second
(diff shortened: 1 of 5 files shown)


ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -323,9 +357,166 @@ func (ipfs *Connector) pinLsHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	respBytes, _ := json.Marshal(pinLs)
+	resBytes, _ := json.Marshal(pinLs)
 	w.WriteHeader(http.StatusOK)
-	w.Write(respBytes)
+	w.Write(resBytes)
+}
+
+func (ipfs *Connector) addHandler(w http.ResponseWriter, r *http.Request) {
```
hsanjuan: @Kubuxu I wouldn't mind a second pair of eyes on the stuff happening in this function...

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -323,9 +357,166 @@ func (ipfs *Connector) pinLsHandler(w http.ResponseWriter, r *http.Request) {
...
+	// Usually when pinning a directory, the recursive pin comes last.
+	// But we may just be pinning different files and no directories.
+	// In that case, we need to recursively pin them separately.
+	// Decide recursivePins takes a conservative approach. It
+	// works on the regular use-cases. Otherwise, it might pin
+	// more things than it should.
+	pinHashes := decideRecursivePins(ipfsAddResps, r.URL.Query())
+
+	logger.Debugf("proxy /add request and will pin %s", pinHashes)
+	for _, pin := range pinHashes {
```
Kubuxu: Does this mean that it will separately pin all files in case of big directory?

Is that how cluster works?

ipfsconn/ipfshttp/ipfshttp.go
Kubuxu: Looked at it, if there is no better way, let's do it like that. I would be for adding additional message to go-ipfs (containing what is getting pinned) it could be used for verbose mode of ipfs add and here.

hsanjuan: I thought of that too. It would be useful, but then, I don't want ipfs to pin because cluster needs to decide where to pin, and it could be a completely different node(s). I could let ipfs pin first and then unpin after it is repinned somewhere else too if this feature existed. In the end what I really need to know is which of the returned things are top-level hashes.

I could force that cluster allocates the pin to this node and some other, but then, there is no point in having metrics and stuff if content gets forcefully pinned where you send it regardless of the pinning strategy.

Kubuxu: There is benefit of pinning at the source as the source already has the content. Not only HDD is resource but bandwidth is one too.

hsanjuan: Yes, gathered thoughts in #81.

Issue #84, opened 2017-03-31
Add a 'running cluster FAQs' to address common questions when using cluster


hsanjuan:
* What happens when a node is shutdown?
* How do I upgrade my cluster?
* How do I manage my existing cluster and what considerations need to be made?
* How is the cluster state stored?
* What are the security implications when running cluster?
* What do I do if a node crashes?
* Can cluster lose my pins?

etc.

PR #90, opened 2017-04-06
A guide to running ipfs cluster

hsanjuan:
This should fix what I meant in #84

License: MIT

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,7 @@
   - [Docker](#docker)
   - [Install from sources](#install-from-sources)
 - [Usage](#usage)
-  - [`ipfs-cluster-service`](#ipfs-cluster-service)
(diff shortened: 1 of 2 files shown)


hsanjuan:
Would be good to get feedback here.

@bgadmin, would you like to read this guide and provide some feedback? Like things that are missing, or not clear enough and can be better explained...

bgadmin:
@hsanjuan sure thing, sorry for late reply will get back to you today :)

VictorBjelkholm:
Would be nice to have a quickstart in the top, similar to what was in the top-level readme.md before, where you just list the minimal set of commands to get up and running.

hsanjuan:
I have added the quickstart to the main README.. The guide is meant for a more in-depth approach.

PR #95, opened 2017-04-28
Updating dockerfile and entrypoint script to have the same behavior t…

### LINKED ISSUE #94: Docker container data volume permissions error
Because of this https://github.com/ipfs/go-ipfs/pull/3744#issuecomment-287241122

The ipfs data volume does not come with right permissions anymore. We need to adapt our entrypoint in the same way that the `run_ipfs.sh` script was adapted (chown the folder and re-execute itself with the right user).

ZenGround0:
This closes #94 .  I have tried to emulate the change in go-ipfs PR #3477.  I ran a docker build to verify that it still runs correctly.  The resulting image is at zenground0/ipfs-cluster-legit.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,14 +10,12 @@ ENV PATH       /go/bin:$PATH
 ENV SRC_PATH   /go/src/github.com/ipfs/ipfs-cluster
 ENV IPFS_CLUSTER_PATH /data/ipfs-cluster
 
-USER root
+VOLUME $IPFS_CLUSTER_PATH
 
 COPY . $SRC_PATH
 
 RUN apk add --no-cache --virtual cluster-deps make musl-dev go git \
     && apk add --no-cache jq \
-    && mkdir -p $IPFS_CLUSTER_PATH \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -31,10 +29,6 @@ RUN apk add --no-cache --virtual cluster-deps make musl-dev go git \
     && apk del --purge cluster-deps \
     && cd / && rm -rf /go/src /go/bin/gx /go/bin/gx-go
 
-USER ipfs
-
-VOLUME $IPFS_CLUSTER_PATH
-
 ENTRYPOINT ["/usr/local/bin/start-daemons.sh"]
```
hsanjuan: Can we try tini like this https://github.com/ipfs/go-ipfs/blob/master/Dockerfile#L65 ?

ZenGround0:
I Updated entrypoint to use tini and squashed commits.  I again built locally and ran on kubernetes ipfs and the image checks out.

hsanjuan:
Thank you!

PR #96, opened 2017-04-28
Docker image experiment

ZenGround0:
WIP Not ready to merge.  Includes new build for stoppable cluster docker image.


docker/entrypoint.sh
```diff
@@ -19,23 +35,28 @@ else
...
     sed -i 's/127\.0\.0\.1\/tcp\/9094/0.0.0.0\/tcp\/9094/' "$IPFS_CLUSTER_PATH/service.json"
     sed -i 's/127\.0\.0\.1\/tcp\/9095/0.0.0.0\/tcp\/9095/' "$IPFS_CLUSTER_PATH/service.json"
 fi
 
-exec ipfs-cluster-service $@
+ipfs-cluster-service $@ &
+
+# Parent process blocks forever. Least ugly solution (still ugly)
+echo "Daemons launched"
+exec tail -f /dev/null
```
hsanjuan: You may redirect the output of cluster to a file and then tail it. That way cluster logs are not lost.

docker/entrypoint.sh
```diff
@@ -19,23 +35,28 @@ else
...
+if [ -e "$cluster_repo/service.json" ]; then
     echo "Found IPFS cluster configuration at $IPFS_CLUSTER_PATH"
 else
     ipfs-cluster-service init
     sed -i 's/127\.0\.0\.1\/tcp\/9094/0.0.0.0\/tcp\/9094/' "$IPFS_CLUSTER_PATH/service.json"
     sed -i 's/127\.0\.0\.1\/tcp\/9095/0.0.0.0\/tcp\/9095/' "$IPFS_CLUSTER_PATH/service.json"
 fi
 
-exec ipfs-cluster-service $@
+ipfs-cluster-service $@ &
```
hsanjuan: From the description of `tini`, it sounded like it may solve the problems you had when sending signals to the ipfs-cluster-service without the need  to launch in the background. Once we merge the other PR, you should test that signaling the process still doesn't work.

docker/entrypoint.sh
hsanjuan: Even better, make a pipe (`mkfifo`), redirect and and tail from it. Should work, but I can't remember the last time I used this...

ZenGround0: I'm happy to use a pipe, however I am looking at the logs for a container running this image and I do see the cluster output.  Should we include a pipe anyway?  It does seem less arbitrary to tail cluster output than dev/null

hsanjuan: You mean seeing them with `docker logs` ?

ZenGround0: Yes, and with kubectl logs.

hsanjuan: ah right! that makes sense, output still going to stdout. Sorry then, forget about this!

hsanjuan:
Closing this as it was just to get feedback on some code

PR #108, opened 2017-06-22
Private Network impl

dgrisham:
Currently working through dependency issues -- here's a write-up on that as requested by @hsanjuan (also tagging @Kubuxu and @whyrusleeping in case they have any immediate suggestions):

1. I included the imports for `go-libp2p-pnet` and `go-libp2p-interface-pnet` as:

```go
import (
    ...
    pnet "github.com/libp2p/go-libp2p-pnet"
    ipnet "github.com/libp2p/go-libp2p-interface-pnet"
    ...
)
```

Then ran `gx-go rewrite --fix && gx-go rewrite`. `go-libp2p-interface-pnet` resolved to a hash, but `go-libp2p-pnet` did not:

```go
import (
    ...
    pnet "github.com/libp2p/go-libp2p-pnet"
    ipnet "gx/ipfs/QmUxRRPqCRmjgZajYGDhUt4MNZFvT8sgry7YkA4ap7qLUP/go-libp2p-interface-pnet"
    ...
)
```

2. I tried to `go build` at this point, but I got the error:

```go
# github.com/ipfs/ipfs-cluster
./cluster.go:874: cannot assign "github.com/libp2p/go-libp2p-interface-pnet".Protector to protec (type "gx/ipfs/QmUxRRPqCRmjgZajYGDhUt4MNZFvT8sgry7YkA4ap7qLUP/go-libp2p-interface-pnet".Protector) in multiple assignment:
	"github.com/libp2p/go-libp2p-interface-pnet".Protector does not implement "gx/ipfs/QmUxRRPqCRmjgZajYGDhUt4MNZFvT8sgry7YkA4ap7qLUP/go-libp2p-interface-pnet".Protector (wrong type for Protect method)
		have Protect("github.com/libp2p/go-libp2p-transport".Conn) ("github.com/libp2p/go-libp2p-transport".Conn, error)
		want Protect("gx/ipfs/QmcYnysCkyGezY6k6MQ1yHHdrRiZaU9x3M9Y1tE9qZ5hD2/go-libp2p-interface-conn".Conn) ("gx/ipfs/QmcYnysCkyGezY6k6MQ1yHHdrRiZaU9x3M9Y1tE9qZ5hD2/go-libp2p-interface-conn".Conn, error)
```

I think (based on my very incomplete understanding of `go` and `gx`) that the issue is with [this import in `go-libp2p-pnet`](https://github.com/libp2p/go-libp2p-pnet/blob/master/protector.go#L6), which doesn't use the `gx` hash to import `go-libp2p-interface-pnet`.

3. I tried various fixes from here, but I now think they were quite ill-informed and probably not too helpful to write-up. I think one way to fix this would be to remove all of the `gx` imports from this project and make them `github.com/<repo>/<project>` imports, as they were before. Another solution might be to change the import in the `go-libp2p-pnet` project that I mentioned above to be the corresponding `gx` import rather than the `github.com/libp2p/go-libp2p-interface-pnet` import (but this might be a rabbit hole into doing this for other repos as well -- maybe it's preferred to get them all changed to `gx` imports anyway, and this would be a good time to do it?).

### DIFF
--- a/allocator/ascendalloc/ascendalloc.go
+++ b/allocator/ascendalloc/ascendalloc.go
@@ -10,10 +10,10 @@ import (
 
 	"github.com/ipfs/ipfs-cluster/api"
 
-	rpc "github.com/hsanjuan/go-libp2p-gorpc"
-	cid "github.com/ipfs/go-cid"
-	logging "github.com/ipfs/go-log"
-	peer "github.com/libp2p/go-libp2p-peer"
+	logging "gx/ipfs/QmSpJByNKFX1sCsHBEp3R73FL4NF6FnQTEGyNAXHm2GS52/go-log"
+	rpc "gx/ipfs/QmYqnvVzUjjVddWPLGMAErUjNBqnyjoeeCgZUZFsAJeGHr/go-libp2p-gorpc"
+	peer "gx/ipfs/QmZcUPvPhD1Xvk6mwijYF8AfR3mG31S1YsEfHG4khrFPRr/go-libp2p-peer"
+	cid "gx/ipfs/QmcTcsTvfaeEBRFo1TkFgT8sRmgi1n1LTZpecfVP8fzpGD/go-cid"
 )
 
 var logger = logging.Logger("ascendalloc")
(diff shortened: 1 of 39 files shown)


hsanjuan:
@dgrisham moved this to #109  so I can make commits to the branch too. You have been added as collaborator to ipfs-cluster so you can write to branches in the project too from now on.

PR #109, opened 2017-06-23
Private Network impl

hsanjuan:
Copied from #108 by @dgrisham :

Currently working through dependency issues -- here's a write-up on that as requested by @hsanjuan (also tagging @Kubuxu and @whyrusleeping in case they have any immediate suggestions):

1. I included the imports for `go-libp2p-pnet` and `go-libp2p-interface-pnet` as:

```go
import (
    ...
    pnet "github.com/libp2p/go-libp2p-pnet"
    ipnet "github.com/libp2p/go-libp2p-interface-pnet"
    ...
)
```

Then ran `gx-go rewrite --fix && gx-go rewrite`. `go-libp2p-interface-pnet` resolved to a hash, but `go-libp2p-pnet` did not:

```go
import (
    ...
    pnet "github.com/libp2p/go-libp2p-pnet"
    ipnet "gx/ipfs/QmUxRRPqCRmjgZajYGDhUt4MNZFvT8sgry7YkA4ap7qLUP/go-libp2p-interface-pnet"
    ...
)
```

2. I tried to `go build` at this point, but I got the error:

```go
# github.com/ipfs/ipfs-cluster
./cluster.go:874: cannot assign "github.com/libp2p/go-libp2p-interface-pnet".Protector to protec (type "gx/ipfs/QmUxRRPqCRmjgZajYGDhUt4MNZFvT8sgry7YkA4ap7qLUP/go-libp2p-interface-pnet".Protector) in multiple assignment:
	"github.com/libp2p/go-libp2p-interface-pnet".Protector does not implement "gx/ipfs/QmUxRRPqCRmjgZajYGDhUt4MNZFvT8sgry7YkA4ap7qLUP/go-libp2p-interface-pnet".Protector (wrong type for Protect method)
		have Protect("github.com/libp2p/go-libp2p-transport".Conn) ("github.com/libp2p/go-libp2p-transport".Conn, error)
		want Protect("gx/ipfs/QmcYnysCkyGezY6k6MQ1yHHdrRiZaU9x3M9Y1tE9qZ5hD2/go-libp2p-interface-conn".Conn) ("gx/ipfs/QmcYnysCkyGezY6k6MQ1yHHdrRiZaU9x3M9Y1tE9qZ5hD2/go-libp2p-interface-conn".Conn, error)
```

I think (based on my very incomplete understanding of `go` and `gx`) that the issue is with [this import in `go-libp2p-pnet`](https://github.com/libp2p/go-libp2p-pnet/blob/master/protector.go#L6), which doesn't use the `gx` hash to import `go-libp2p-interface-pnet`.

3. I tried various fixes from here, but I now think they were quite ill-informed and probably not too helpful to write-up. I think one way to fix this would be to remove all of the `gx` imports from this project and make them `github.com/<repo>/<project>` imports, as they were before. Another solution might be to change the import in the `go-libp2p-pnet` project that I mentioned above to be the corresponding `gx` import rather than the `github.com/libp2p/go-libp2p-interface-pnet` import (but this might be a rabbit hole into doing this for other repos as well -- maybe it's preferred to get them all changed to `gx` imports anyway, and this would be a good time to do it?).


hsanjuan:
@dgrisham as this is now, build fails because of changes in the code (not more because of deps, I think)...

you can run `make build` as short-hand to build and `make clean` before pushing changes (don't commit rewritten paths). Check the makefile to see the steps.

What I have done is to bump go-libp2p in this branch of ipfs-cluster, libp2p-gorpc, and libp2p-raft (they need to use the same libp2p). Not sure if that was necessary. Then I have `gx import`ed libp2p-pnet.

dgrisham:
@hsanjuan awesome, thanks Hector! I'll take a look

hsanjuan:
As discussed:

* I'm happy if the private key is part of the configuration. We should set some safeguards to make sure that it is not too short or anything like that.
* The `NewDefaultConfig` configuration function should generate a random key.
* The init command in `ipfs-cluster-service` should offer the option of setting the pnets key (or disabling pnets -leaving it empty-). Disabled pnets should show a warning on start (show it from `makeHost`). When doing `init` I'd propose something like a `-pnetkey` which optionally reads the value from an environment variable (https://github.com/urfave/cli#values-from-the-environment).
* We need to print some Info message on start explaining the ipfs-cluster is configured with a Private Network key and that all members should share it.
* We need to update all documentation referencing cluster intialization, as well as the guide (I'll try to point out places).

On a separate PR, to be further discussed:

* PeerAdd-ing a new member which is not configured with the shared key will not work anymore.
* PeerAdd should support adding new members which are started without a key by sending them the key.
* Still, the Join (bootstrap) operation should not work unless the peer joining has the valid PNet key.

Kubuxu:
PNet has also `fingerprint` function which you can use to print which PNet the network is on without revealing the PNet key itself. Fingerprint is safe to share.

hsanjuan:
@dgrisham I have thought the UX for Pnets:

Let's have a simple text password in the configuration and let's call it `cluster_secret`. Hide PNet implementation-specific details to the user, basically automatically turn the password from whatever it is to the format required by PNets:

```
/key/swarm/psk/1.0.0/
/base16/
<pw>
```

Rationale is: as a user, I just want to deploy a ipfs+cluster. I don't want to become familiar with IPFS nomenclatures (pnet == "private networks" == "locking the swarm to peers that have a pre-shared-key"). As a user I don't want to have tio figure out why this is base16 and not base64 (or why it's not using multibase), if there is anything alternative to psk/1.0.0 etc. or to run a separate script to generate this. I just want my cluster to be locked down.

`cluster_secret` captures much better the sense that it's something private and shared for the whole cluster. The dependencies are pinned, any changes in key format for pnet will make tests fail. Since the configuration already contains sensitive material (the peer's private key), it's not worse to include more secrets in it.

In the future we can decide if we read these from file, or offer alternatives to providing secrets to ipfs-cluster.

config.go
```diff
@@ -409,12 +409,26 @@ func (cfg *Config) unshadow() {
...
+	const charEnd = 126
+	const numChars = charEnd - charStart + 1
+
+	// modify random chars to be in desired range
+	secret := make([]byte, secretLength)
+	for j := range rand {
+		secret[j] = byte(int(rand[j])%numChars + charStart)
+	}
+
+	return string(secret), nil
```
Kubuxu: Why are you decided to change it from `hex.EncodeToString()`, encode is for sure not biased.
Your function will be biased as for doing modulo 94 over `0-255` range which will have slight bias for values `0-67`. Which will translate in bias for decoded numbers too.

Better solution is either `hex.EncodeToString` or using `/bin/` coded and passing raw seed in the string.

dgrisham: Ah, yes, very very good point. Thanks for catching that! Was going to have @hsanjuan check this out, but you got there first ;)

ipfs-cluster-service/main.go
```diff
@@ -190,8 +197,15 @@ func main() {
 		{
 			Name:  "init",
 			Usage: "create a default configuration and exit",
+			Flags: []cli.Flag{
+				cli.BoolFlag{
+					Name:  "generate-secret",
+					Usage: "automatically generate swarm secret",
```
hsanjuan: cluster secret

ipfs-cluster-service/main.go
```diff
@@ -190,8 +197,15 @@ func main() {
 		{
 			Name:  "init",
 			Usage: "create a default configuration and exit",
+			Flags: []cli.Flag{
+				cli.BoolFlag{
+					Name:  "generate-secret",
```
hsanjuan: let's short this a bit for faster typing: `gen-secret,g`

ipfs-cluster-service/main.go
```diff
@@ -225,7 +239,8 @@ func main() {
 
 func run(c *cli.Context) error {
 	if c.Bool("init") {
-		initConfig(c.Bool("force"))
+		initConfig(c.Bool("force"), c.Bool("generate-secret"),
```
hsanjuan: generate-secret and env-swarm-secret are not flags in the main application so just use `false, ""`.

ipfs-cluster-service/main.go
```diff
@@ -337,12 +352,33 @@ func setupDebug() {
...
 
-func initConfig(force bool) {
+func initConfig(force bool, generateSecret bool, envSecret string) {
 	if _, err := os.Stat(configPath); err == nil && !force {
 		err := fmt.Errorf("%s exists. Try running with -f", configPath)
 		checkErr("", err)
 	}
-	cfg, err := ipfscluster.NewDefaultConfig()
+
+	var swarmSecret *string
```
hsanjuan: use string, not pointer. rename to clusterSecret

ipfs-cluster-service/main.go
```diff
@@ -337,12 +352,33 @@ func setupDebug() {
...
+		swarmSecret = &envSecret
+	} else {
+		// get swarm secret from user
+		scanner := bufio.NewScanner(os.Stdin)
+		fmt.Print("Enter swarm secret (to automatically generate, rerun with --generate-secret): ")
+		scanner.Scan()
+		// QuoteToASCII leaves outer double quotes on the resulting string,
+		// Unqoute removes them, and hence there *should* never be an error, so
+		// we ignore it
+		input, _ := strconv.Unquote(strconv.QuoteToASCII(scanner.Text()))
```
hsanjuan: not sure this is needed. in the end the string should be just casted to []byte, and go handles that well for non-ascii chars.

ipfs-cluster-service/main.go
```diff
@@ -337,12 +352,33 @@ func setupDebug() {
...
+		swarmSecret = &input
+	}
+
+	cfg, err := ipfscluster.NewDefaultConfig(swarmSecret)
```
hsanjuan: as mentioned above, just do cfg.ClusterSecret = []byte(clusterSecret) after this

hsanjuan: Along these lines:
```
cfg, err := ipfscluster.NewDefaultConfig(swarmSecret)
checkErr("creating default configuration", err)
if envSecret != "" {
   fmt.Println("Reading swarm secret from SWARM_SECRET environment variable.")
   cfg.ClusterSecret = []byte(envSecret)
} else if !generateSecret {
   cfg.ClusterSecret = []byte(promptUser("Enter swarm secret..."))
}
```

Thinking it would be good to have a promptUser function that we can re-use to ask things interactively from the user in the future.

config.go
```diff
@@ -394,9 +408,32 @@ func (cfg *Config) unshadow() {
...
+	key := make([]byte, 32)
+	_, err := crand.Read(key)
+	if err != nil {
+		return "", fmt.Errorf("Error reading from rand: %v", err)
+	}
+	return hex.EncodeToString(key), nil
+}
+
+func swarmSecretToKey(secret string) (string, error) {
+	hexSecret := hex.EncodeToString([]byte(secret))[:68]
```
Kubuxu: I am open to PR to go-libp2p-pnet that adds direct handling of `[]byte` secret, it would simplify this part quite a bit.

config.go
```diff
@@ -394,9 +408,32 @@ func (cfg *Config) unshadow() {
...
+	key.WriteString("/base16/\n")
+	key.WriteString(hexSecret)
+
+	return key.String(), nil
+}
+
 // NewDefaultConfig returns a default configuration object with a randomly
 // generated ID and private key.
-func NewDefaultConfig() (*Config, error) {
+func NewDefaultConfig(swarmSecret *string) (*Config, error) {
```
dgrisham: Regarding your comment on pointers, I've actually been curious about whether it's considered good form to use a pointer in place of what Haskell/Rust would refer to as `Nothing` or `None`, resp. E.g. if we want an empty string (`""`) to be distinguished from 'no input', it seems that the best option we have in Go is to use a pointer to a string and use `nil` to mean 'no input' (which I definitely think is bad practice (subverting the type system), but is sort of made to be the common solution based on the design of Go (and many languages)). Maybe there's a better alternative though?

Also, regarding this case in particular, is there any good reason in allowing an empty swarm secret/key (`""`)?

config.go
dgrisham: @hsanjuan Thanks for catching the 68, that was purely for testing purposes. And @Kubuxu yeah that would be good I think, I'll make a PR soon if no one else gets to it first.

config.go
```diff
@@ -34,6 +37,11 @@ type Config struct {
 	ID         peer.ID
 	PrivateKey crypto.PrivKey
 
+	// Swarm key for private network
+	SwarmSecret string
+	// Fingerprint for private network
+	PNetFingerprint []byte
```
dgrisham: Right, I had this in my notes to address at some point, and I agree that it doesn't belong here. It was originally a minor convenience for passing it from the `makeHost()` function.

config.go
```diff
@@ -394,9 +408,32 @@ func (cfg *Config) unshadow() {
 	cfg.shadow = nil
 }
 
+func generateSwarmSecret() (string, error) {
+	key := make([]byte, 32)
+	_, err := crand.Read(key)
+	if err != nil {
+		return "", fmt.Errorf("Error reading from rand: %v", err)
+	}
+	return hex.EncodeToString(key), nil
```
dgrisham: Good idea, I like this solution.

dgrisham:
Just pushed to address the last round of review comments.

config.go
```diff
@@ -408,22 +410,21 @@ func (cfg *Config) unshadow() {
...
-	if len(hexSecret) < 10 {
-		return "", fmt.Errorf("Swarm secret must be >= 10 chars, but is %d\n",
-			len(hexSecret))
-	}
+func clusterSecretToKey(secret []byte) (string, error) {
+	hexSecret := hex.EncodeToString(secret)
+
 	var key bytes.Buffer
 	key.WriteString("/key/swarm/psk/1.0.0/\n")
 	key.WriteString("/base16/\n")
```
Kubuxu: As you are storing your secret in base64, just pass it in there.

dgrisham: Nice, meant to check whether that would be an option.

dgrisham:
The next push will include updates to the docs, as well as error handling in the event that the user does not input a base64 `cluster_secret` when initializing a configuration.

dgrisham:
Remaining Todo:

- [ ] Enforce 32-byte key on `ipfs-cluster-service init`
- [ ] Enforce 32-byte key when starting cluster
- [ ] Update `ClusterSecret` comment in config to reflect 32-byte key requirement

dgrisham:
@hsanjuan @Kubuxu this is ready for review (again)

config.go
```diff
@@ -394,6 +416,38 @@ func (cfg *Config) unshadow() {
...
+	var key bytes.Buffer
+	key.WriteString("/key/swarm/psk/1.0.0/\n")
+	key.WriteString("/base64/\n")
+	key.Write(secret)
+
+	return key.String(), nil
+}
+
+func IsValidSecret(secret []byte) (bool, error) {
+	var isBase64 = regexp.MustCompile(`^[0-9A-Za-z+/]+$`).Match
```
hsanjuan: Base64 encoding has `=` sign for padding

Just do a `parseSecret` function that does https://golang.org/pkg/encoding/base64/#Encoding.DecodeString (and returns an error if it doesn't work, and the checks that the length is 32 before returning the byte slice. No need to regexp stuff ourselves.

ipfs-cluster-service/main.go
```diff
@@ -337,13 +350,34 @@ func setupDebug() {
...
 	}
+
 	cfg, err := ipfscluster.NewDefaultConfig()
 	checkErr("creating default configuration", err)
+
+	if !generateSecret {
+		if len(envSecret) != 0 {
+			// read cluster secret from env variable
+			fmt.Println("Reading cluster secret from CLUSTER_SECRET environment variable.")
+			cfg.ClusterSecret = []byte(envSecret)
```
hsanjuan: base64.decode envSecret

config.go
```diff
@@ -394,6 +416,38 @@ func (cfg *Config) unshadow() {
 	cfg.shadow = nil
 }
 
+func generateClusterSecret() ([]byte, error) {
+	encodedSecretLength := 64
+	secret := make([]byte, base64.StdEncoding.DecodedLen(encodedSecretLength))
```
hsanjuan: This is weird. The secret is 32 bytes. This is not something that you should infer from the length of the encoded base64 representation of the secret (which by the way, is 48). You are making a longer than 32 byte secret here.

ipfs-cluster-service/main.go
hsanjuan: and check the result is 32-bytes

ipfs-cluster-service/main.go
```diff
@@ -337,13 +350,34 @@ func setupDebug() {
...
+		} else {
+			// get cluster secret from user
+			isValid := false
+			for !isValid {
+				cfg.ClusterSecret = []byte(promptUser("Enter cluster secret (to automatically generate, rerun with --gen-secret): "))
```
hsanjuan: base64.decode the secret and check the result is 32byte

config.go
hsanjuan: I think we can expose a `func ParseClusterSecret(b64secret string) ([]byte, error) {}`. You can re-use it when parsing the JSON configuration and in ipfs-cluster-service to validate user-input.

ipfs-cluster-service/main.go
```diff
@@ -337,13 +350,34 @@ func setupDebug() {
...
+
+	if !generateSecret {
+		if len(envSecret) != 0 {
+			// read cluster secret from env variable
+			fmt.Println("Reading cluster secret from CLUSTER_SECRET environment variable.")
+			cfg.ClusterSecret = []byte(envSecret)
+		} else {
+			// get cluster secret from user
+			isValid := false
+			for !isValid {
```
hsanjuan: You can directly abort (`checkErr()`) the first time the user screws up. Makes no differencein our case (we only have 1 question) and facilitates testing.

dgrisham:
@hsanjuan I appreciate the thorough feedback, dealing with multiple encodings is fairly new territory to me (which I think shows in my code). I'll make these changes and let you know if I have any questions!

hsanjuan:
@dgrisham thanks!

config_test.go
```diff
@@ -2,13 +2,13 @@ package ipfscluster
 
 import "testing"
 
-var testingClusterSecret = []byte("2951539a3737c06a5aee55834c27145ca1783bdc7daeaa92f9712b3ff6e9fa25")
+var testingClusterSecret, _ = DecodeClusterSecret("2588b80d5cb05374fa142aed6cbb047d1f4ef8ef15e37eba68c65b9d30df67ed")
```
dgrisham: Should be fine to ignore error here, but is this frowned upon still?

hsanjuan: You can ignore the err

PR #111, opened 2017-06-29
Docker test image add cluster respawn

ZenGround0:
@hsanjuan one last set of changes to enable functionality needed by tests in kubernetes-ipfs PR # 27.


hsanjuan:
Thanks! The automatic build should be ready in a few minutes.

Issue #86, opened 2017-04-04
Content from manually removed peers is not re-allocated


hsanjuan:
I think this happens because the monitor only throws alerts for the current set of peers so it will not complain for a peer that has been removed from the cluster. Should not be like this. Remember to add a test.

PR #113, opened 2017-07-03
Re-allocate pins on peer removal

hsanjuan:
PeerRm now triggers re-pinning of all the Cids allocated to
the removed peer.

#86 

License: MIT

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -305,7 +305,7 @@ func (c *Cluster) alertsHandler() {
 	}
 }
 
-// find all Cids pinned to a given peer and triggers re-pins on them
+// find all Cids pinned to a given peer and triggers re-pins on them.
 func (c *Cluster) repinFromPeer(p peer.ID) {
 	cState, err := c.consensus.State()
 	if err != nil {
@@ -317,7 +317,7 @@ func (c *Cluster) repinFromPeer(p peer.ID) {
 		for _, alloc := range pin.Allocations {
 			if alloc == p { // found pin allocated to node
 				logger.Infof("repinning %s out of %s", pin.Cid, p.Pretty())
-				c.Pin(pin)
+				c.pin(pin, []peer.ID{p}) // pin blacklisting this peer
 			}
 		}
 	}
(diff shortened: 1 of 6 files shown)


cluster.go
```diff
@@ -1135,8 +1124,9 @@ func (c *Cluster) allocate(hash *cid.Cid, repl int) ([]peer.ID, error) {
 
 		// we don't have enough peers to pin
 		if len(candidateAllocs) < needed {
```
dgrisham: If `candidateAllocs` is just `candidates` sorted by priority, can't we check `len(candidates) < needed` before we call `Allocate`? Would save a sorting step if we don't have enough candidates anyway.

hsanjuan: It's great that you saw that! I saw it too :) You can shortcut the call to allocator.Allocate, but, after calling it, we need to check it anyways.

This is because the `allocator` is a black box (in the future we'd want to support multiple strategies). So even with enough candidates (as in "peers reporting valid metrics"), it could decide against using them (think of an allocator which must allocate to peers in different regions, but all candidates are from the same region). So the output of that call could have less peers than the candidates we give it.

Seeing that, I dropped the `len(candidates) < needed` case because it is a condition that would be caught after calling the allocator anyways.

That said, since you wondered about it, and since the allocator would be a really slow black box, I'll re-add it. Probably makes code more explicit (and marginally more efficient).

dgrisham: Gotcha, thought it might have to do with keeping the 'allocate' strategy space open :) Yeah, hard to say whether it has much gain for an arbitrary allocation strategy/situation, but seems worth checking to catch potentially costly cases.

cluster.go
```diff
@@ -1135,8 +1127,9 @@ func (c *Cluster) allocate(hash *cid.Cid, repl int) ([]peer.ID, error) {
...
-			err = logError("cannot find enough allocations for this CID: needed: %d. Got: %s",
-				needed, candidateAllocs)
+			err = logError(
+				"cannot find enough allocations for %s. Needed: %d. Got: %d",
+				hash, needed, candidateAllocs)
```
dgrisham: `s/candidateAllocs/len(candidateAllocs)/` ?

dgrisham: Might be nice to print out the number as well (e.g. `Got <len(candidateAllocs>: <candidateAllocs>`). But otherwise, this all lgtm :+1:

PR #114, opened 2017-07-03
improve ipfs-cluster-ctl command help.

hsanjuan:
It seems that urfave/cli acts differently handling Description and Usage
depending on whether it is dealing with a command or a subcommand.

This fixes most of these issues, as well as the commands output which was
printing some blank lines.

License: MIT

### DIFF
--- a/ipfs-cluster-ctl/formatters.go
+++ b/ipfs-cluster-ctl/formatters.go
@@ -83,7 +83,7 @@ func textFormatPrintIDSerial(obj *api.IDSerial) {
 	}
 
 	fmt.Printf("%s | %d peers\n", obj.ID, len(obj.ClusterPeers))
-	addrs := sort.StringSlice{}
+	addrs := make(sort.StringSlice, 0, len(obj.Addresses))
 	for _, a := range obj.Addresses {
 		addrs = append(addrs, string(a))
 	}
@@ -97,7 +97,7 @@ func textFormatPrintIDSerial(obj *api.IDSerial) {
 		return
 	}
 
-	ipfsAddrs := sort.StringSlice{}
+	ipfsAddrs := make(sort.StringSlice, 0, len(obj.Addresses))
 	for _, a := range obj.IPFS.Addresses {
 		ipfsAddrs = append(ipfsAddrs, string(a))
 	}
@@ -110,7 +110,7 @@ func textFormatPrintIDSerial(obj *api.IDSerial) {
 
 func textFormatPrintGPinfo(obj *api.GlobalPinInfoSerial) {
 	fmt.Printf("%s :\n", obj.Cid)
-	peers := sort.StringSlice{}
+	peers := make(sort.StringSlice, 0, len(obj.PeerMap))
 	for k, _ := range obj.PeerMap {
 		peers = append(peers, k)
 	}
--- a/ipfs-cluster-ctl/formatters.go
+++ b/ipfs-cluster-ctl/formatters.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/ipfs/ipfs-cluster/api"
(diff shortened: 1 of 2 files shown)


ipfs-cluster-ctl/main.go
```diff
@@ -128,19 +129,17 @@ This command will print out information about the cluster peer used
...
 					Name:  "ls",
 					Usage: "list the nodes participating in the IPFS Cluster",
-					UsageText: `
-This commands provides a list of the ID information of all the peers in the Cluster.
+					Description: `
+This command provides a list of the ID information of all the peers in the Cluster.
 `,
-					Flags: []cli.Flag{parseFlag(formatID)},
+					Flags:     []cli.Flag{parseFlag(formatID)},
+					ArgsUsage: " ",
```
dgrisham: Probably doesn't make a lot of difference, but `ArgsUsage: ""` instead of `" "`? Or is there a reason for the space?

Hm, after looking further on it seems you do this a few times. I won't mark all of them, it's starting to look intentional.

ipfs-cluster-ctl/main.go
```diff
@@ -271,14 +266,14 @@ although unpinning operations in the cluster may take longer or fail.
...
-					UsageText: `
+					Description: `
 This command will list the CIDs which are tracked by IPFS Cluster and to
 which peers they are currently allocated. This list does not include
 any monitoring information about the IPFS status of the CIDs, it
 merely represents the list of pins which are part of the shared state of
 the cluster. For IPFS-status information about the pins, use "status".
 `,
-					ArgsUsage: "[cid]",
+					ArgsUsage: "[CID]",
```
dgrisham: Should this be blank (`ArgsUsage: ""`), or does `pin ls` take in a list for filtering results or something?

ipfs-cluster-ctl/main.go
```diff
@@ -412,10 +409,17 @@ func parseFlag(t int) cli.IntFlag {
...
 }
 
-func walkCommands(cmds []cli.Command) {
+func walkCommands(cmds []cli.Command, parentHelpName string) {
 	for _, c := range cmds {
-		fmt.Println(c.HelpName)
-		walkCommands(c.Subcommands)
+		var h string
+		// Sometimes HelpName is empty :S
+		if h = c.HelpName; h != "" {
```
dgrisham: This block is a bit confusing to me, one of the below might be more clear --

```go
h := c.HelpName
if h == "" {
    h = fmt.Sprintf("%s %s", parentHelpName, c.FullName())
}
fmt.Println(h)
```

or, if you prefer

```go
var h string
if c.HelpName != "" {
    h = c.HelpName
} else {
    h = fmt.Sprintf("%s %s", parentHelpName, c.FullName())
}
fmt.Println(h)
```

Up to you of course, just my preference.

ipfs-cluster-ctl/formatters.go
```diff
@@ -82,23 +83,41 @@ func textFormatPrintIDSerial(obj *api.IDSerial) {
 	}
 
 	fmt.Printf("%s | %d peers\n", obj.ID, len(obj.ClusterPeers))
-	fmt.Println("  > Addresses:")
+	addrs := sort.StringSlice{}
```
dgrisham: Hm, that link is acting funny for me, I'm meaning to link to line 351 (not 340).

ipfs-cluster-ctl/main.go
hsanjuan: It's intentional "", is the default value, and it will print '<command arguments>' when not set. " " removes that.

ipfs-cluster-ctl/main.go
hsanjuan: pin ls takes an optional [CID] as filter.

ipfs-cluster-ctl/main.go
hsanjuan: thanks that's def clearer

ipfs-cluster-ctl/formatters.go
hsanjuan: I'm going to do addrs := make(sort.StringSlice, 0, len(obj.Addresses)).

The problem is not the append per se, but the slice capacity which forces appends to re-allocate.

dgrisham: Ah, good point, forgot slices worked that way. Good fix too, I think it's better than what I suggested (because the `sort.Strings()` has an implicit cast that you're not doing now, not sure how costly that is but still).

dgrisham:
lgtm!

PR #118, opened 2017-07-12
Fix #97: Assume default DataFolder as subfolder to config folder when empty

### LINKED ISSUE #97: Permission denied when mounting host directory with Docker
This might be due to my poor knowledge of Docker but:

`docker run -d --name ipfs-cluster -v ~/.ipfs-cluster:/data/ipfs-cluster ipfs/ipfs-cluster`
yields
```
ERROR  consensus: creating file snapshot store: snapshot path not accessible: mkdir /Users: permission denied raft.go:107
[...]
error starting cluster: snapshot path not accessible: mkdir /home/me: permission denied
```

hsanjuan:
We no longer set ConsensusDataFolder. We leave it empty (and ommited from the
configuration). When not set, it will take the path from which the configuration
file was read and use an "ipfs-cluster-data" subfolder in that path.

When set, the behaviour is just as before (ensures backwards compatiblity).

This will facilitate re-use of configuration files, for example, when mounting
them inside docker.

#97 

License: MIT

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -159,10 +159,15 @@ func (c *Cluster) setupConsensus() error {
 		startPeers = peersFromMultiaddrs(c.config.Bootstrap)
 	}
 
+	dataFolder := c.config.ConsensusDataFolder
+	if dataFolder == "" {
+		dataFolder = filepath.Join(filepath.Dir(c.config.path), DefaultDataFolder)
+	}
+
 	consensus, err := raft.NewConsensus(
 		append(startPeers, c.id),
 		c.host,
-		c.config.ConsensusDataFolder,
+		dataFolder,
 		c.state)
 	if err != nil {
 		logger.Errorf("error creating consensus: %s", err)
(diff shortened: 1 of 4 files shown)


cluster.go
```diff
@@ -156,10 +156,15 @@ func (c *Cluster) setupConsensus() error {
 		startPeers = peersFromMultiaddrs(c.config.Bootstrap)
 	}
 
+	dataFolder := c.config.ConsensusDataFolder
```
dgrisham: Would it make more sense to check this value/change it when loading the config instead?

dgrisham:
@hsanjuan Just had one question, otherwise lgtm!

cluster.go
hsanjuan: Yeah, the reason I did not do it when loading the config, is that if you set ConsensusDataFolder there (or anywhere else) to a expected value  when it's empty, then you need to remove it when you save the configuration (configuration is saved sometimes, for example when peers are added) and make sure it's exactly as it was loaded. So I went the simpler way.

This way configuration stays the same.

dgrisham: Gotcha, I noticed that the config wasn't updated so this makes sense.

PR #123, opened 2017-07-14
Allow empty cluster secret, update test to reflect this.

dgrisham:
Small update to allow empty cluster secret for starting cluster on unprotected network.


hsanjuan:
@dgrisham a related test is segfaulting. See if you can fix it.

Issue #120, opened 2017-07-13
Support HTTPs API endpoints


hsanjuan:
Allow to configure SSL certificate and key for cluster REST API endpoints.

PR #124, opened 2017-07-19
WIP: HTTPS API endpoint support

dgrisham:
This PR addresses the issue https://github.com/ipfs/ipfs-cluster/issues/120. The current implementation builds, but has not yet been tested. The default TLS configuration used was copied from https://github.com/denji/golang-tls#perfect-ssl-labs-score-with-go and has not yet been verified outside of that source.

### DIFF
--- a/api/restapi/restapi.go
+++ b/api/restapi/restapi.go
@@ -4,6 +4,7 @@ package restapi
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"net"
 	"net/http"
@@ -77,7 +78,25 @@ func NewRESTAPI(apiMAddr ma.Multiaddr) (*RESTAPI, error) {
 	if err != nil {
 		return nil, err
 	}
+	return newRESTAPI(apiMAddr, l)
+}
 
+// NewTlsRESTAPI creates a new REST API component that uses TLS for security
+// (authentication, encryption). It receives the multiaddress on which the API
+// listens, as well as paths to certificate and private key files
+func NewTLSRESTAPI(apiMAddr ma.Multiaddr, tlsCfg *tls.Config) (*RESTAPI, error) {
+	n, addr, err := manet.DialArgs(apiMAddr)
+	if err != nil {
+		return nil, err
+	}
+	l, err := tls.Listen(n, addr, tlsCfg)
+	if err != nil {
+		return nil, err
+	}
+	return newRESTAPI(apiMAddr, l)
+}
+
+func newRESTAPI(apiMAddr ma.Multiaddr, l net.Listener) (*RESTAPI, error) {
 	router := mux.NewRouter().StrictSlash(true)
 	s := &http.Server{
 		ReadTimeout:  RESTAPIServerReadTimeout,
--- a/config.go
+++ b/config.go
@@ -66,6 +66,18 @@ type Config struct {
 	// Listen parameters for the the Cluster HTTP API component.
 	APIAddr ma.Multiaddr
 
+	// TLSCertFile is a path to a certificate file used for identifying the
+	// Cluster's HTTP(S) API. TLSKeyFile must also be set in order to establish
+	// a TLS server. If both this and TLSKeyFile are empty, then HTTP (without
+	// TLS) will be used
+	TLSCertFile string
+
+	// TLSKeyFile is a path to a private key used for setting up TLS sessions for
+	// the HTTP(S) API. TLSCertFile must also be set in order to establish a TLS
+	// server. If both this and TLSCertFile are empty, then HTTP (without
+	// TLS) will be used
+	TLSKeyFile string
+
 	// Listen parameters for the IPFS Proxy. Used by the IPFS
 	// connector component.
 	IPFSProxyAddr ma.Multiaddr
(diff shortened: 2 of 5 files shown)


pnet_test.go
```diff
@@ -3,31 +3,31 @@ package ipfscluster
 import "testing"
 
 func TestClusterSecretFormat(t *testing.T) {
-    goodSecret := "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
-    emptySecret := ""
-    tooShort := "0123456789abcdef"
-    tooLong := "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0"
-    unsupportedChars := "0123456789abcdef0123456789!!!!!!0123456789abcdef0123456789abcdef"
+	goodSecret := "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
```
hsanjuan: Hm if we are to be scrict, this belongs on a different PR...

api/restapi/restapi.go
```diff
@@ -66,14 +67,39 @@ type peerAddBody struct {
...
+		tls.TLS_RSA_WITH_AES_256_CBC_SHA,
+	},
+}
+
 // NewRESTAPI creates a new REST API component. It receives
-// the multiaddress on which the API listens.
-func NewRESTAPI(apiMAddr ma.Multiaddr) (*RESTAPI, error) {
+// the multiaddress on which the API listens, as well as (potentially empty)
+// paths to certificate and private key files for TLS usage
+func NewRESTAPI(apiMAddr ma.Multiaddr, certFile, keyFile string) (*RESTAPI, error) {
```
hsanjuan: It's not a good practice to change the API (I mean, as in the signature of a public function of a module). So better to add a `NewTlsRESTAPI` or something like that.

A plus is that then you can make sure the user is not giving you an empty certificate path when he really means to use certificates.

dgrisham: Right, you told me that before too (about not changing the API). That's one of the big differences between writing code for production vs. personal use/school haha.

pnet_test.go
dgrisham: Yeah, this was the result of me running `go fmt` (which I'd forgotten to do before). Easy enough to change, I'll make another PR with this fix.

dgrisham:
@hsanjuan Just added a few tests for this feature, most of the functionality should be tested via the `NewRESTAPI()` tests (which shares a common helper function with `NewTLSRESTAPI()` that does most of the work). The tests I wrote primarily test the `tls.LoadX509KeyPair()` function, which obviously isn't really testing our code. There's a bit more happening in the `ipfs-cluster-service/main.go:run()` function that might be good to test in some way, not quite sure how far to go here.

api/restapi/restapi.go
```diff
@@ -77,7 +92,30 @@ func NewRESTAPI(apiMAddr ma.Multiaddr) (*RESTAPI, error) {
 	if err != nil {
 		return nil, err
 	}
+	return newRESTAPI(apiMAddr, l)
+}
+
+// NewTlsRESTAPI creates a new REST API component that uses TLS for security
+// (authentication, encryption). It receives the multiaddress on which the API
+// listens, as well as paths to certificate and private key files
+func NewTLSRESTAPI(apiMAddr ma.Multiaddr, certFile, keyFile string) (*RESTAPI, error) {
```
hsanjuan: Let this function receive a tls.Config directly: cluster integrator's responsability to set it up.

For ipfs-cluster-service, we are the integrator, so you have to set it up there, :)

api/restapi/restapi.go
```diff
@@ -36,6 +38,19 @@ var (
 	RESTAPIServerIdleTimeout = 120 * time.Second
 )
 
+// copied from https://github.com/denji/golang-tls
+var tlsCfg = &tls.Config{
```
hsanjuan: Move to ipfs-cluster-service/main.go (see comment below).

config.go
```diff
@@ -66,6 +66,21 @@ type Config struct {
 	// Listen parameters for the the Cluster HTTP API component.
 	APIAddr ma.Multiaddr
 
+	// If APIUseTLS is `true`, then Cluster will use TLS (configured via the
```
hsanjuan: Redundant. Let ipfs-cluster-service use TLS if either TLSCertFile or TLSKeyFile != "".

ipfs-cluster-service/main.go
```diff
@@ -263,7 +263,15 @@ func run(c *cli.Context) error {
 		cfg.AllocationStrategy = a
 	}
 
-	api, err := restapi.NewRESTAPI(cfg.APIAddr)
+	var api *restapi.RESTAPI
+	if cfg.APIUseTLS {
```
hsanjuan: Ok, following from above, here you can check if the CertFile or KeyFile are defined. Attempt to create the TLS Config and then pass it to NewTLSRESTAPI.

dgrisham:
@hsanjuan This last push should address all of the comments you made, assigned from the self-signed/sharness points. I tested with `ipfs-cluster-ctl --https`and everything that I could test behaved as expected (basically, either client/server forcing TLS while other wasn't would result in a failed connection). But you're right, I couldn't test both using TLS since I'm using a self-signed key.

I looked a bit and found this: https://golang.org/src/crypto/tls/common.go#L441. Might be the way to go, but we can also talk about this/look around for other approaches.

ipfs-cluster-service/main.go
dgrisham: Awesome, this is much cleaner.

ipfs-cluster-ctl/main.go
```diff
@@ -85,6 +87,10 @@ func main() {
 			Name:  "https, s",
 			Usage: "use https to connect to the API",
 		},
+		cli.BoolFlag{
+			Name:  "skip-verify-cert",
```
hsanjuan: let's use `--no-check-certificate`, that's what `wget` calls it.

dgrisham: Fixed, and yeah we can do sharness in a different PR.

PR #125, opened 2017-07-20
Ran `go fmt` to format pnet_test.go

dgrisham:


Kubuxu:
You might want to enable codeclimate with `go fmt`, it would catch that. For sample config see `go-ipfs` but it has some other functions enabled.

PR #127, opened 2017-07-24
Release fixes

hsanjuan:
A couple of changes after playing with cluster a little bit, before tagging a new release.

### DIFF
--- a/ipfs-cluster-service/main.go
+++ b/ipfs-cluster-service/main.go
@@ -194,13 +194,13 @@ This command will initialize a new service.json configuration file
 for %s.
 
 By default, %s requires a cluster secret. This secret will be
-automatically generated, but can be manually provided with -custom-secret
+automatically generated, but can be manually provided with --custom-secret
 (in which case it will be prompted), or by setting the CLUSTER_SECRET
 environment variable.
 
 The private key for the libp2p node is randomly generated in all cases.
 
-Note that the -force first-level-flag allows to overwrite an existing
+Note that the --force first-level-flag allows to overwrite an existing
 configuration.
 `, programName, programName),
 			ArgsUsage: " ",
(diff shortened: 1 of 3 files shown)


ipfs-cluster-service/main.go
```diff
@@ -356,15 +371,14 @@ func initConfig(force bool, generateSecret bool, envSecret string) {
...
-		}
+	if len(envSecret) != 0 {
+		// read cluster secret from env variable
+		fmt.Println("Reading cluster secret from CLUSTER_SECRET environment variable.")
+		cfg.ClusterSecret, err = ipfscluster.DecodeClusterSecret(envSecret)
+	} else if !generateSecret {
+		// get cluster secret from user
+		cfg.ClusterSecret, err = ipfscluster.DecodeClusterSecret(promptUser("Enter cluster secret (to automatically generate, rerun with --gen-secret): "))
+
 		checkErr("parsing cluster secret", err)
```
dgrisham: Move this below bottom brace, currently doesn't check the `err` from the first `if` case.

ipfs-cluster-service/main.go
```diff
@@ -356,15 +371,14 @@ func initConfig(force bool, generateSecret bool, envSecret string) {
...
-			// get cluster secret from user
-			cfg.ClusterSecret, err = ipfscluster.DecodeClusterSecret(promptUser("Enter cluster secret (to automatically generate, rerun with --gen-secret): "))
-		}
+	if len(envSecret) != 0 {
+		// read cluster secret from env variable
+		fmt.Println("Reading cluster secret from CLUSTER_SECRET environment variable.")
+		cfg.ClusterSecret, err = ipfscluster.DecodeClusterSecret(envSecret)
+	} else if !generateSecret {
+		// get cluster secret from user
+		cfg.ClusterSecret, err = ipfscluster.DecodeClusterSecret(promptUser("Enter cluster secret (to automatically generate, rerun with --gen-secret): "))
```
dgrisham: `--gen-secret` --> `--cluster-secret` (also, we should be consistent with `--long-option` vs. `-long-option` usage in the docs/printed output. I'm a fan of the former because the latter could be ambiguous, but consistency is the main thing :) )

ipfs-cluster-service/main.go
```diff
@@ -194,15 +189,35 @@ func main() {
...
+
+Note that the -force first-level-flag allows to overwrite an existing
+configuration.
+`, programName, programName),
+			ArgsUsage: " ",
 			Flags: []cli.Flag{
 				cli.BoolFlag{
-					Name:  "gen-secret, g",
-					Usage: "automatically generate cluster secret",
+					Name:  "custom-secret, s",
```
dgrisham: Commit message says that the short flag here is `m`, change either this or commit message to be consistent.

ipfs-cluster-service/main.go
hsanjuan: Yeah good point. I'm not sure. `--option` feels more natural to me (for long options) and `-option` feels more Go-ish. I might have been using both forms depending on the feeling of the day. I have used `--option` for these changes. Happy to settle on that and keep an eye on next doc reviews.

ipfs-cluster-service/main.go
hsanjuan: changed commit message

hsanjuan:
comments addressed

PR #129, opened 2017-07-25
Rename TLS configuration options and set JSON key

hsanjuan:
Also, omit these keys in JSON generation when they are empty.

License: MIT


hsanjuan:
I noticed that we forgot to set the meta-tag for the json encoding in these config options. And since I was at it, I renamed them. "SSLCert" naming is more usual for these things (i.e. apache,nginx).

Issue #136, opened 2017-07-26
Empty CLUSTER_SECRET does not result in empty cluster_secret when initializing configuration


hsanjuan:
it's ignored when the variable is defined but with an empty value...

hsanjuan:
@dgrisham gonna give this one to you

dgrisham:
`urfave/cli` doesn't seem to have support for distinguishing between unset and empty environment variables. I'll use [os.LookupEnv](https://golang.org/pkg/os/#LookupEnv) for this.

PR #140, opened 2017-07-28
Distinguish unset/empty cluster secret; also rename generateSecret -> customSecret

dgrisham:
This resolves https://github.com/ipfs/ipfs-cluster/issues/136. It also renames the `generateSecret` variable used in `ipfs-cluster-service/main.go:initConfig()` to `customSecret` to keep it consistent with the way `initConfig()` is called (less confusing to understand for newcomers, I think).

### DIFF
--- a/ipfs-cluster-service/main.go
+++ b/ipfs-cluster-service/main.go
@@ -212,12 +212,8 @@ configuration.
 				},
 			},
 			Action: func(c *cli.Context) error {
-				envSecret, envSecretDefined := os.LookupEnv("CLUSTER_SECRET")
-				if envSecretDefined {
-					initConfig(c.GlobalBool("force"), c.Bool("custom-secret"), &envSecret)
-				} else {
-					initConfig(c.GlobalBool("force"), c.Bool("custom-secret"), nil)
-				}
+				userSecret, userSecretDefined := userProvidedSecret(c.Bool("custom-secret"))
+				initConfig(c.GlobalBool("force"), userSecret, userSecretDefined)
 				return nil
 			},
 		},
@@ -250,7 +246,7 @@ configuration.
 
 func run(c *cli.Context) error {
 	if c.Bool("init") {
-		initConfig(c.Bool("force"), false, nil)
+		initConfig(c.Bool("force"), nil, false)
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


ipfs-cluster-service/main.go
```diff
@@ -379,11 +378,11 @@ func initConfig(force bool, generateSecret bool, envSecret string) {
 	cfg, err := ipfscluster.NewDefaultConfig()
 	checkErr("creating default configuration", err)
 
-	if len(envSecret) != 0 {
+	if envSecret != nil {
 		// read cluster secret from env variable
 		fmt.Println("Reading cluster secret from CLUSTER_SECRET environment variable.")
-		cfg.ClusterSecret, err = ipfscluster.DecodeClusterSecret(envSecret)
-	} else if !generateSecret {
+		cfg.ClusterSecret, err = ipfscluster.DecodeClusterSecret(*envSecret)
```
hsanjuan: I think this will break when envSecret = "" ? Will the Decode throw an error then?

ipfs-cluster-service/main.go
```diff
@@ -210,15 +210,14 @@ configuration.
...
-					Name:   "env-cluster-secret",
-					EnvVar: "CLUSTER_SECRET",
-					Hidden: true,
-				},
 			},
 			Action: func(c *cli.Context) error {
-				initConfig(c.GlobalBool("force"), !c.Bool("custom-secret"),
-					c.String("env-cluster-secret"))
+				envSecret, envSecretDefined := os.LookupEnv("CLUSTER_SECRET")
+				if envSecretDefined {
```
hsanjuan: Should we have a function which takes:

```
func userProvidedSecret(customSecret bool) (secret []byte, defined bool) {
var secret string
if customSecret {
  secret =  prompt user
} else if CLUSTER_SECRET is defined
  secret = value
} else {
  return empty, false
}

if secret = "" {
  return []byte{}, true
} else {
  decode
  return decoded secret, true
}
```

And then if defined == true, overwrite the configuration-generated one. Otherwise do nothing. ?

ipfs-cluster-service/main.go
dgrisham: It doesn't, decode just returns an empty result as well. Works as expected when I test it with `CLUSTER_SECRET=""`.

ipfs-cluster-service/main.go
dgrisham: Nice, yeah I like that because it cleans up the logic around `initConfig()` and takes care of grabbing a user-specified secret in one place (instead of doing part of it outside of `initConfig()` and part of it inside).

hsanjuan: yeah, also more clear about what takes priority

dgrisham:
Just pushed the commit that implements the discussed `userSecretDefined()` function.

hsanjuan:
THANKS

PR #141, opened 2017-07-28
Sharness tests for API SSL support

dgrisham:
Simple sharness tests for the HTTPS API support implemented in https://github.com/ipfs/ipfs-cluster/pull/124.

I also modified `sharness/lib/test-lib.sh:test_cluster_init` to make it easy to copy config files relevant to a specific test (used to move my `service.json` and ssl key/cert into the cluster config), we may want to consider whether there's a cleaner way to do that (`getopt`?).


sharness/lib/test-lib.sh
```diff
@@ -66,6 +68,9 @@ test_cluster_init() {
         echo "error initializing ipfs cluster"
         exit 1
     fi
+    if [[ -n "$custom_config_files" ]]; then
```
hsanjuan: Big thing to remember: we use `sh`, not `bash`. Bash features are not available (`[[`). This is why travis build fails.

(I'm not sure why we don't use `/bin/bash` everywhere either, it might be ok but Wyatt used `/bin/sh`. I like that `sh` prevents from doing a lot of magic, as shell scripts + magic is a bad combination though)

dgrisham: Ah, nice, I noticed `[` was used throughout and meant to test the difference (but forgot it was an `sh` thing). Thanks, about to push that change.

dgrisham: And `sh` seems sufficient for this, and is also used in the chriscool/sharness example (maybe just to be shell-agnostic though, not sure).

hsanjuan:
sharness still failing

hsanjuan:
error creating TLS config: : Error loading TLS certficate/key: open server.crt: no such file or directory

hsanjuan:
Hey!

So there were some problems:

* The test that inspired you for the new ones was probably buggy. I think it was doing `| egrep ""` so it suceeded all the time. I have fixed that.
* ipfs-cluster was not starting. It's a bit difficult to check what starts and what not since it runs in the background. Problem was that the certificate path is relative to the running folder of ipfs-cluster, not to the configuration folder. To fix this I have changed the temp folder into a permanent one. I pushed for temp folders initially, but it's more confusing, made the custom config thing more difficult and gets cleaned up anyway.

My tricks for testing:

* There is a IPFS_OUPUT variable in the test_lib.sh, which can be set to /dev/stderr, which shows the logs of the things starting
* You can redirect output to files within the tests. If the test fail, a `trash something` directory is left with everything that test wrote. If a test passes it's removed. If you still wanna check it out, you can comment the `test_done` bit.
* `set -x` might helps in cases (for initialization code for example), but since tests run inside evals it's sometimes very useless or confusing.

Now, crossing fingers that travis becomes green.

PR #142, opened 2017-08-03
`FreeSpace` allocation metric impl

dgrisham:
Currently a WIP, but looking for feedback on implementation. I've added a `disk.NewInformerWithMetric(metric string)` function that supports `"disk-reposize"` and `"disk-freespace"` metrics (values checked in `setupAllocation()`, may want to check in `disk.NewInformerWithMetric()` instead). I think the disk Informer impl is fairly clean and makes it easy to add new metrics (simply by adding an entry to the `nameToRPC` map), but let me know what you think @hsanjuan.

TODO:
- [x] implement `descendalloc`
- [x] Tests for freespace metric

### DIFF
--- a/allocator/ascendalloc/ascendalloc.go
+++ b/allocator/ascendalloc/ascendalloc.go
@@ -5,8 +5,6 @@
 package ascendalloc
 
 import (
-	"sort"
-
 	"github.com/ipfs/ipfs-cluster/allocator/util"
 	"github.com/ipfs/ipfs-cluster/api"
 
@@ -38,7 +36,5 @@ func (alloc AscendAllocator) Shutdown() error { return nil }
 // candidates based on their metric values (smallest to largest).
 func (alloc AscendAllocator) Allocate(c *cid.Cid, current, candidates map[peer.ID]api.Metric) ([]peer.ID, error) {
 	// sort our metrics
-	sortable := util.NewMetricSorter(candidates, false)
-	sort.Sort(sortable)
-	return sortable.Peers, nil
+	return util.SortNumeric(candidates, false), nil
 }
--- a/allocator/ascendalloc/ascendalloc.go
+++ b/allocator/ascendalloc/ascendalloc.go
@@ -1,13 +1,13 @@
-// Package ascendalloc implements an ipfscluster.Allocator returns allocations
-// based on sorting the metrics in ascending order. Thus, peers with smallest
-// metrics are first in the list. This allocator can be used with a number
-// of informers, as long as they provide a numeric metric value.
+// Package ascendalloc implements an ipfscluster.PinAllocator, which returns
+// allocations based on sorting the metrics in ascending order. Thus, peers with
+// smallest metrics are first in the list. This allocator can be used with a
+// number of informers, as long as they provide a numeric metric value.
 package ascendalloc
 
 import (
 	"sort"
-	"strconv"
 
+	"github.com/ipfs/ipfs-cluster/allocator/util"
 	"github.com/ipfs/ipfs-cluster/api"
 
 	rpc "github.com/hsanjuan/go-libp2p-gorpc"
@@ -18,79 +18,27 @@ import (
 
 var logger = logging.Logger("ascendalloc")
 
-// Allocator implements ipfscluster.Allocate.
-type Allocator struct{}
+// AscendAllocator extends the SimpleAllocator
+type AscendAllocator struct{}
 
-// NewAllocator returns an initialized Allocator
-func NewAllocator() *Allocator {
-	return &Allocator{}
+// NewAscendAllocator returns an initialized AscendAllocator
+func NewAllocator() AscendAllocator {
+	return AscendAllocator{}
 }
 
 // SetClient does nothing in this allocator
-func (alloc *Allocator) SetClient(c *rpc.Client) {}
+func (alloc AscendAllocator) SetClient(c *rpc.Client) {}
 
 // Shutdown does nothing in this allocator
-func (alloc *Allocator) Shutdown() error { return nil }
+func (alloc AscendAllocator) Shutdown() error { return nil }
 
 // Allocate returns where to allocate a pin request based on metrics which
 // carry a numeric value such as "used disk". We do not pay attention to
-// the metrics of the currently allocated peers and we just sort the candidates
-// based on their metric values (from smallest to largest).
-func (alloc *Allocator) Allocate(c *cid.Cid, current, candidates map[peer.ID]api.Metric) ([]peer.ID, error) {
+// the metrics of the currently allocated peers and we just sort the
+// candidates based on their metric values (smallest to largest).
+func (alloc AscendAllocator) Allocate(c *cid.Cid, current, candidates map[peer.ID]api.Metric) ([]peer.ID, error) {
 	// sort our metrics
-	sortable := newMetricsSorter(candidates)
+	sortable := util.NewMetricSorter(candidates, false)
 	sort.Sort(sortable)
-	return sortable.peers, nil
-}
-
-// metricsSorter attaches sort.Interface methods to our metrics and sorts
-// a slice of peers in the way that interest us
-type metricsSorter struct {
-	peers []peer.ID
-	m     map[peer.ID]int
-}
-
-func newMetricsSorter(m map[peer.ID]api.Metric) *metricsSorter {
-	vMap := make(map[peer.ID]int)
-	peers := make([]peer.ID, 0, len(m))
-	for k, v := range m {
-		if v.Discard() {
-			continue
-		}
-		val, err := strconv.Atoi(v.Value)
-		if err != nil {
-			continue
-		}
-		peers = append(peers, k)
-		vMap[k] = val
-	}
-
-	sorter := &metricsSorter{
-		m:     vMap,
-		peers: peers,
-	}
-	return sorter
-}
-
-// Len returns the number of metrics
-func (s metricsSorter) Len() int {
-	return len(s.peers)
-}
-
-// Less reports if the element in position i is less than the element in j
-func (s metricsSorter) Less(i, j int) bool {
-	peeri := s.peers[i]
-	peerj := s.peers[j]
-
-	x := s.m[peeri]
-	y := s.m[peerj]
-
-	return x < y
-}
-
-// Swap swaps the elements in positions i and j
-func (s metricsSorter) Swap(i, j int) {
-	temp := s.peers[i]
-	s.peers[i] = s.peers[j]
-	s.peers[j] = temp
+	return sortable.Peers, nil
 }
--- a/allocator/descendalloc/descendalloc_test.go
+++ b/allocator/descendalloc/descendalloc_test.go
@@ -5,7 +5,6 @@ import (
 	"time"
 
 	"github.com/ipfs/ipfs-cluster/api"
-	"github.com/ipfs/ipfs-cluster/informer/numpin"
 
 	cid "github.com/ipfs/go-cid"
 	peer "github.com/libp2p/go-libp2p-peer"
@@ -31,25 +30,25 @@ var testCases = []testcase{
 	{ // regular sort
 		candidates: map[peer.ID]api.Metric{
 			peer0: api.Metric{
-				Name:   numpin.MetricName,
+				Name:   "some-metric",
 				Value:  "5",
 				Expire: inAMinute,
 				Valid:  true,
 			},
 			peer1: api.Metric{
-				Name:   numpin.MetricName,
+				Name:   "some-metric",
 				Value:  "1",
 				Expire: inAMinute,
 				Valid:  true,
 			},
 			peer2: api.Metric{
-				Name:   numpin.MetricName,
+				Name:   "some-metric",
 				Value:  "3",
 				Expire: inAMinute,
 				Valid:  true,
 			},
 			peer3: api.Metric{
-				Name:   numpin.MetricName,
+				Name:   "some-metric",
 				Value:  "2",
 				Expire: inAMinute,
 				Valid:  true,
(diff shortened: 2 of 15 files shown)


hsanjuan:
I think there are files missing (util)

dgrisham:
@hsanjuan Thanks, was in a hurry when I pushed. Just added the missing files

allocator/allocator.go
```diff
@@ -0,0 +1,91 @@
...
+	"strconv"
+
+	"github.com/ipfs/ipfs-cluster/api"
+
+	rpc "github.com/hsanjuan/go-libp2p-gorpc"
+	cid "github.com/ipfs/go-cid"
+	peer "github.com/libp2p/go-libp2p-peer"
+)
+
+type Allocator interface {
```
hsanjuan: Dup from the PinAllocator interface? I'm not sure why it's needed

informer/disk/disk.go
```diff
@@ -12,13 +12,21 @@ import (
 	"github.com/ipfs/ipfs-cluster/api"
 )
 
+// TODO: switch default to disk-freespace
```
hsanjuan: dont forget

informer/disk/disk.go
```diff
@@ -28,6 +36,14 @@ type Informer struct {
 
 // NewInformer returns an initialized Informer.
 func NewInformer() *Informer {
+	MetricName = DefaultMetric
```
hsanjuan: On the line of previous comment, I think you can simply make the metric name a property of the informer...

informer/disk/disk.go
```diff
@@ -12,13 +12,21 @@ import (
...
+const DefaultMetric = "disk-reposize"
+
 var logger = logging.Logger("diskinfo")
 
 // MetricTTL specifies how long our reported metric is valid in seconds.
 var MetricTTL = 30
 
 // MetricName specifies the name of our metric
-var MetricName = "disk"
+var MetricName string
```
hsanjuan: Before this was more like a configuration option for the module, but now it's a property of the informer defined on creation, so it should not exist anymore as global variable.

informer/disk/disk.go
```diff
@@ -49,30 +65,28 @@ func (disk *Informer) Name() string {
...
 		}
 	}
 
-	var repoSize int
+	var metric int
 	valid := true
 	err := disk.rpcClient.Call("",
 		"Cluster",
-		"IPFSRepoSize",
+		nameToRPC[MetricName],
```
hsanjuan: either handle the metricName -> rpcname conversion via a switch statement (doing everything here in the function) with a `default` section or add code to make sure that the metric name is in the nameToRPC map. If not, log the problem clearly and return an invalid metric.

Right now the error logged would probably be very cryptic (rpc method missing i guess) and probably difficult to figure out it's just from a typo in the metric name.

informer/disk/disk.go
```diff
@@ -12,13 +12,21 @@ import (
 	"github.com/ipfs/ipfs-cluster/api"
```
hsanjuan: I can't place comment before this line: The godoc description of the module needs to be updated to explain the things this informer disk informer can do.

informer/disk/disk.go
```diff
@@ -28,6 +36,14 @@ type Informer struct {
 
 // NewInformer returns an initialized Informer.
```
hsanjuan: mention that it will use the default metric

informer/disk/disk.go
```diff
@@ -28,6 +36,14 @@ type Informer struct {
...
 func NewInformer() *Informer {
+	MetricName = DefaultMetric
+	return &Informer{}
+}
+
+// NewInformer returns an initialized Informer.
```
hsanjuan: mention which are the valid metrics that can be used

allocator/allocator.go
```diff
@@ -0,0 +1,91 @@
...
+type SimpleAllocator struct{}
+
+// SetClient does nothing in this allocator
+func (alloc SimpleAllocator) SetClient(c *rpc.Client) {}
+
+// Shutdown does nothing in this allocator
+func (alloc SimpleAllocator) Shutdown() error { return nil }
+
+// Allocate does nothing in this allocator (important to override this)
+func Allocate(c *cid.Cid, current, candidates map[peer.ID]api.Metric) ([]peer.ID, error) {
```
hsanjuan: I think this was meant to be  `func (alloc SimpleAllocator) Allocate()?`

allocator/allocator.go
```diff
@@ -0,0 +1,91 @@
...
+	// Allocate returns where to allocate a pin request based on metrics which
+	// carry a numeric value such as "used disk". We do not pay attention to
+	// the metrics of the currently allocated peers and we just sort the
+	// candidates based on their metric values.
+	Allocate(c *cid.Cid, current, candidates map[peer.ID]api.Metric) ([]peer.ID, error)
+}
+
+// SimpleAllocator implements the Allocator interface. Other allocators can
+// extend SimpleAllocator and should at least override the Allocate() function.
+type SimpleAllocator struct{}
```
hsanjuan: I'm not very sure why a `SimpleAllocator` which does nothing is necessary...

allocator/allocator.go
```diff
@@ -0,0 +1,91 @@
...
+func (alloc SimpleAllocator) Shutdown() error { return nil }
+
+// Allocate does nothing in this allocator (important to override this)
+func Allocate(c *cid.Cid, current, candidates map[peer.ID]api.Metric) ([]peer.ID, error) {
+	return nil, nil
+}
+
+// MetricsSorter attaches sort.Interface methods to our metrics and sorts
+// a slice of peers in the way that interest us
+type MetricsSorter interface {
```
hsanjuan: hmm this is just re-defining https://golang.org/pkg/sort/#Interface

allocator/allocator.go
```diff
@@ -0,0 +1,91 @@
...
+	Len() int
+	Swap(i, j int)
+	Less(i, j int) bool
+}
+
+// SimpleSorter implements the MetricsSorter interface. Other sorters may extend
+// this and should at least override the default Less() function.
+type SimpleSorter struct {
```
hsanjuan: Probably remove the MetricSorter interface and rename this to MetricSorter (a type that implements the sort.Interface for metrics). The type can have a property `Reverse` which would allow it to sort in reverse order.

allocator/ascendalloc/ascendalloc.go
```diff
@@ -1,96 +1,50 @@
...
 
 var logger = logging.Logger("ascendalloc")
 
-// Allocator implements ipfscluster.Allocate.
-type Allocator struct{}
-
-// NewAllocator returns an initialized Allocator
-func NewAllocator() *Allocator {
-	return &Allocator{}
+// AscendAllocator extends the SimpleAllocator
```
hsanjuan: SimpleAllocator does nothing so I don't quite get the benefit of extending. I think it is simpler to implement the ipfscluster.PinAllocator interface directly and use the common MetricSorter.

allocator/allocator.go
dgrisham: Right, I realized that before but then apparently forgot it completely while playing with Go's type system.

informer/disk/disk.go
dgrisham: Ah, okay I see. I was a bit uncertain if I should change that or not, your explanation makes sense though.

ipfs-cluster-service/main.go
```diff
@@ -342,8 +344,12 @@ func setupLogging(lvl string) {
 
 func setupAllocation(strategy string) (ipfscluster.Informer, ipfscluster.PinAllocator) {
 	switch strategy {
-	case "disk", "reposize":
-		return disk.NewInformer(), ascendalloc.NewAllocator()
+	case "disk", "disk-freespace":
+		informer := disk.NewInformer()
+		return informer, descendalloc.NewAllocator()
```
dgrisham: @hsanjuan It might be nice to have a way for the `Informer` to specify which `Allocator` it should use based on the metric (not sure whether there's a good reason to keep those decoupled like they are now).

hsanjuan: They're meant to be fully decoupled. You might gather metrics for a bunch of  reasons not related to an allocator. Then, there might be metrics which make sense with several allocators.

dgrisham: Gotcha, that makes sense.

hsanjuan:
Tests related to metric/allocator are failing...

cluster_test.go
```diff
@@ -79,6 +79,7 @@ func (ipfs *mockConnector) PinLs(filter string) (map[string]api.IPFSPinStatus, e
 
 func (ipfs *mockConnector) ConnectSwarms() error                          { return nil }
 func (ipfs *mockConnector) ConfigKey(keypath string) (interface{}, error) { return nil, nil }
+func (ipfs *mockConnector) FreeSpace() (int, error)                       { return 0, nil }
```
hsanjuan: Probably this makes the tests fail, as you report 0 free space, nothing can be allocated

informer/disk/disk.go
```diff
@@ -12,23 +13,44 @@ import (
...
+		metricName: DefaultMetric,
+	}
+}
+
+// NewInformerWithMetric returns an initialized Informer with a specified metric
+// name, if that name is valid, or nil. A metric name is valid if it is a key in
+// the nameToRPC map.
+func NewInformerWithMetric(metric string) *Informer {
+	// check whether specified metric is supported
+	if _, valid := nameToRPC[metric]; valid {
```
hsanjuan: There something here I don't fully like.

I think we need constants for allowed metric types for the disk informer:

```go
const (
  MetricFreeSpace = iota
  MetricRepoSize
)
```

`NewInformerWithMetric(metric int)` should panic and inform the user that given metric is invalid. It should also log a critical error (try to see if logger.Critical works, otherwise logger.Error will do).

`nameToRPC` becomes a map from int -> string

`Informer` can store directly the RPC name (that way there is no need to read it from the map on every GetMetric).

allocator/descendalloc/descendalloc_test.go
```diff
@@ -0,0 +1,116 @@
...
+	testCid, _ = cid.Decode("QmP63DkAFEnDYNjDYBpyNDfttu1fvUw99x1brscPzpqmmq")
+)
+
+var inAMinute = time.Now().Add(time.Minute).Format(time.RFC3339Nano)
+
+var testCases = []testcase{
+	{ // regular sort
+		candidates: map[peer.ID]api.Metric{
+			peer0: api.Metric{
+				Name:   numpin.MetricName,
```
hsanjuan: Change numpin.Metric name to DefaultMetric or something.. this was probably an old copypasta bug, with the side effect that it imports a totally unrelated module

informer/disk/disk.go
hsanjuan: Actually `type MetricType int` with NewInformerWithMetric(metric MetricType)`

dgrisham:
@hsanjuan Just made an intermediate push addressing some of your review, but not all of it. As of now, I'm treating the `Informer`'s name as completely separate from the `MetricType` -- the `Type` (`MetricType`) determines how the `Informer` behaves, and the `name` is simply the user-facing name used to identify the `Informer` (so it can be anything).

dgrisham:
Hm, lot of failed tests on that last round. I'll look at this closer soon.

informer/disk/disk.go
```diff
@@ -12,23 +13,57 @@ import (
...
-	return &Informer{}
+// NewInformer returns an initialized Informer that uses the DefaultMetric
+func NewInformer(name string) *Informer {
+	return &Informer{
+		name:    name,
+		rpcName: metricToRPC[DefaultMetric],
+	}
+}
+
+func NewInformerWithMetric(metric MetricType, name string) *Informer {
```
hsanjuan: godoc comment missing

informer/disk/disk.go
```diff
@@ -44,35 +79,28 @@ func (disk *Informer) Shutdown() error {
...
+		&metric)
 	if err != nil {
 		logger.Error(err)
 		valid = false
 	}
 
 	m := api.Metric{
-		Name:  MetricName,
-		Value: fmt.Sprintf("%d", repoSize),
+		Name:  disk.rpcName,
```
hsanjuan: disk.name is perhaps better?

informer/disk/disk.go
```diff
@@ -12,23 +13,61 @@ import (
...
+func NewInformer(name string) *Informer {
+	return &Informer{
+		name:    name,
+		rpcName: metricToRPC[DefaultMetric],
+	}
+}
+
+// NewInformerWithMetric returns an Informer that uses the input MetricType. The
+// name argument has the same purpose as in NewInformer.
+func NewInformerWithMetric(metric MetricType, name string) *Informer {
```
hsanjuan: Can we return (*Informer, error) and set the error of the metric type is invalid here?

hsanjuan:
@dgrisham ops I didn't see your last comment on irc about wanting to debug the issue yourself, so I went ahead and fixed it. I found it pretty quickly and it was something silly (replace ascendalloc with descendalloc in cluster initialization).

Went over things again and left one last comment. I think I can merge after. Good job!

informer/disk/disk.go
dgrisham: Ah, right, good idea. Definitely want to do that.

dgrisham:
@hsanjuan No worries, glad it was relatively simple -- figured it had to be. I'll make that change and push it up!

dgrisham:
@hsanjuan Hm, so right now the default metric used within cluster is determined within `setupAllocation()`, and the `NewInformer()` method is never used -- do you think it's worth keeping the `NewInformer()` separate from `NewInformerWithMetric()`? The way I see it, we could handle the default metric in two ways:

1. Initialize with the name `"disk"`, like `NewInformer("disk")`. This would indicate that the user was using the default `"disk"` metric, but would not clarify which one.
2. Initialize with a specific name, like "`disk-freespace"` (which is what we're doing now). In this case we never call `NewInformer()`, which I think makes sense because if we did it could be confusing -- if we initialize with `NewInformer("disk-freespace")` but later the `DefaultMetric` changes to something else, then we might forget to change the name in `NewInformer("disk-freespace")` which could result in something confusing/difficult to debug.

The way we're doing it now seems best to me, but then I'm not sure it makes sense to keep a `NewInformer()` that uses a default metric.

dgrisham:
@hsanjuan Decided to keep `NewInformer` but remove the `name` argument (automatically set to to `"disk-default"`).

hsanjuan:
THANKS @dgrisham !

Issue #121, opened 2017-07-13
Support basic auth in REST API endpoints


hsanjuan:
Password-protecting these might be useful. Must include integration from ipfs-cluster-ctl.

hsanjuan:
Nice to have:

* Cluster should support an array of { username: <user>, password: <pass> } which are used by the API component to grant or deny access
* Probably a boolean to activate it (disabled by default)
* `ipfs-cluster-ctl` takes a -u flag for the username:password. (also allow to take from some env variable)
* Introduce a `.ipfs-cluster/ctl.json` configuration file used by `ipfs-cluster-ctl` which allows to permanently configure:
  * Basic authorization user and password
  * ipfs-cluster host
  * https (and ignore)
  * Output format

hsanjuan:
Implementation in #147

PR #147, opened 2017-09-07
Issue #121: BasicAuth REST API

dgrisham:
@hsanjuan Here's a start to the BasicAuth implementation. Todos:

- Server-side
    - [x] Default to HTTPS when BasicAuth active
    - [x] Better response to bad credentials.
- CLI flags for server
    - [x] (De?)activate BasicAuth
    - [x] Force BasicAuth w/ HTTP (HTTPS by default)
- Client-side
    - [x] BasicAuth impl
    - [x] `-c <username>:<password>` flag
    - [x] Env var input for username/password
        - `CLUSTER_CREDENTIALS="<username>:<password>"`
    - [x] Default to HTTPS with BasicAuth (flag to override) -- this mirrors the server behavior
- Tests
    - [x] Sharness


api/restapi/restapi.go
```diff
@@ -70,33 +75,26 @@ type peerAddBody struct {
...
+	return NewRESTAPIWithConfig(apiMAddr, &Config{})
 }
 
-// NewTlsRESTAPI creates a new REST API component that uses TLS for security
-// (authentication, encryption). It receives the multiaddress on which the API
-// listens, as well as paths to certificate and private key files
-func NewTLSRESTAPI(apiMAddr ma.Multiaddr, tlsCfg *tls.Config) (*RESTAPI, error) {
+// NewRESTAPI creates a new REST API component. It receives
+// the multiaddress on which the API listens.
+func NewRESTAPIWithConfig(apiMAddr ma.Multiaddr, cfg *Config) (*RESTAPI, error) {
```
hsanjuan: Should include the apiMaddr as part of the configuration I guess

api/restapi/restapi.go
```diff
@@ -56,6 +56,11 @@ type RESTAPI struct {
 	wg           sync.WaitGroup
 }
 
+type Config struct {
```
hsanjuan: Don't forget to document public types

api/restapi/restapi.go
dgrisham: Sounds good -- the only difference that I see so far is that the `apiMAddr` is require, while everything in the `cfg` is optional. Not sure if that difference matters much, but for now I'm combining them into a single config (so we'll have `NewRESTAPI(cfg *Config)`).

dgrisham:
@hsanjuan @ZenGround0 Just made a push, the current state should be consistent with the Todo list in the OP.

dgrisham:
@hsanjuan I'm a bit confused about how to handle the error response that cluster-service sends when bad BasicAuth credentials are provided. As of now, I'm using `http.Error(w, "Unauthorized", 401)` in [this handler function](https://github.com/ipfs/ipfs-cluster/blob/2bf672ad6666d2c5215fbcc0b428758aa073afad/api/restapi/restapi.go#L143), also at line [153](https://github.com/ipfs/ipfs-cluster/blob/2bf672ad6666d2c5215fbcc0b428758aa073afad/api/restapi/restapi.go#L153).

However, when cluster-ctl provides bad BA credentials, the CLI tool just says:

```
error decoding JSON: invalid character 'U' looking for beginning of value
```

Which comes from [here](https://github.com/ipfs/ipfs-cluster/blob/2bf672ad6666d2c5215fbcc0b428758aa073afad/ipfs-cluster-ctl/formatters.go#L55) and the fact that the response body starts with 'Unauthorized'.

Obviously I need to format the response with JSON somehow, but it's not immediately clear to me how. I've only put a modest amount of time into looking around the codebase for other examples of this, but there don't seem to be many (at least that I can find with `grep -r http.Error`).

sharness/t0042-basic-auth.sh
```diff
@@ -0,0 +1,41 @@
...
+test_cluster_init "$config"
+cleanup test_clean_cluster
+
+test_expect_success "prerequisites" '
+    test_have_prereq IPFS && test_have_prereq CLUSTER
+'
+
+test_expect_success "BasicAuth fails without credentials" '
+    id=`cluster_id`
+    ipfs-cluster-ctl id | grep -q Unauthorized
```
dgrisham: I'm not convinced checking for 'Unauthorized' in the response is the best approach to this -- e.g. we could accidentally be leaking info in some case but the presence of the word 'Unauthorized' would allow this test to pass anyway.

dgrisham: @hsanjuan

dgrisham:
@hsanjuan Two questions on the remaining Todos I have listed:

- **Config file to hold cluster-ctl config** -- Should I leave that to https://github.com/ipfs/ipfs-cluster/issues/162 and just support CLI/env var input for now?
- **Go tests** -- sharness seems sufficient/the best way to test this (think we did the same for TLS), but lmk if you think we need Go tests here too

sharness/t0042-basic-auth.sh
hsanjuan: @dgrisham check that the command did not return 0 for exit code and then check for Unauthorized (should be ok for the moment).

ipfs-cluster-ctl/main.go
```diff
@@ -105,11 +107,30 @@ func main() {
 			Name:  "debug, d",
 			Usage: "set debug log level",
 		},
+		cli.StringFlag{
+			Name: "credentials, c",
+			Usage: "specify BasicAuth credentials for server that requires " +
```
hsanjuan: Would need a hint that they are passed as "user:password"

ipfs-cluster-ctl/main.go
```diff
@@ -105,11 +107,30 @@ func main() {
 			Name:  "debug, d",
 			Usage: "set debug log level",
 		},
+		cli.StringFlag{
+			Name: "credentials, c",
```
hsanjuan: perhaps rename to --basic-auth (or even `-u, --user USER[:PASSWORD]  Server user and password` like `curl` does) . Not sure if it deserves a short-hand alias. I'd like to keep -c for config (in line with server).

hsanjuan:
> Config file to hold cluster-ctl config -- Should I leave that to #162 (or some other PR) and just support CLI/env var input for now?

Yes, let's leave it like that for the moment

>    Go tests -- sharness seems sufficient/the best way to test this (think we did the same for TLS), but lmk if you think we need Go tests here too

Sharness is ok

hsanjuan:
Implements #141

ipfs-cluster-ctl/main.go
dgrisham: Hm, so right now (I think) an empty password would work by passing `--basic-auth <username>:` (note the `:` at the end) but you'd get an error if you just passed `--basic-auth <username>` (no `:`). Might be nice to support the latter instead, I'll make that change.

hsanjuan: sounds good

ZenGround0:
@dgrisham I had the same Travis problem yesterday.  Rebasing on top of master should do the trick.

ipfs-cluster-ctl/main.go
```diff
@@ -447,8 +469,15 @@ func request(method, path string, body io.Reader, args ...string) *http.Response
 	checkErr("creating request", err)
 	r.WithContext(ctx)
 
+	if len(defaultUsername) != 0 {
+		r.SetBasicAuth(defaultUsername, defaultPassword)
+	}
+
 	client := &http.Client{Transport: defaultTransport}
 	resp, err := client.Do(r)
+	if err == nil && resp.StatusCode >= 400 {
```
hsanjuan: Now that this has been brought to my attention :)

I think the best place to detect error and set exit code in such case is the end of `formatResponse`. This ensures the received data/error is shown to the user in the desired format. Only then you can check resp.StatusCode and exit with 2 when it was not a successful request.

dgrisham: Gotcha, yeah I was definitely planning on discussion/correction on this approach. And out of curiosity, why exit code 2 specifically? Is there a good rule of thumb for the first few exit codes that I can keep in mind?

dgrisham:
@ZenGround0 Ah, dope, thanks for the tip :)

dgrisham:
@hsanjuan Tests passing! Let me know if there are any other changes you think I should make.

ipfs-cluster-ctl/main.go
hsanjuan: Nah, I said 2 because 1 is used for the rest of errors, so just to use a different one for errors which are from responses. It's completely arbitrary...

dgrisham:
@hsanjuan Don't merge this quite yet, there are a couple of sharness tests I want to add first.

ipfs-cluster-ctl/main.go
dgrisham: Gotcha, good to know. We should document that somewhere for any users who might want to distinguish between to the two.

Issue #150, opened 2017-09-11
Update tests to work with private network default


ZenGround0:
Docker entrypoints and respawning scripts need to explicitly use the empty secret.  Alternatively the configs of different cluster nodes could all share a secret.

hsanjuan:
@ZenGround0 is this a problem at the moment or did we fix it?

ZenGround0:
PR #156 addressed so closing issue.

PR #156, opened 2017-09-21
set cluster secret to empty in test image

ZenGround0:
One-line fix to make testing image compatible with cluster secret feature.  Addresses #150


hsanjuan:
thanks

Issue #112, opened 2017-06-30
Cluster consistently fails to propogate pins to newly added peers


ZenGround0:
This issue was first found when running test 0015-add_pin_rm_peers.yml.  For concreteness I will discuss results with 4 peers and 3 files pinned, though the problem occurs for all peer and file-pin-numbers seen so far.  This test begins by removing two peers from the cluster.  It then pins 3 files to the cluster.  Finally it restarts the cluster daemon in the two removed peers and adds them back to the cluster with newly generated peer ids.  

The expectation is that after being added back the two new peers will replicate all 3 files (rep factor is default -1).  Consistently one peer will not pin all files.  In the instance of the failure I am currently documenting the first peer to depart and re-enter the cluster only replicates one of the three expected pinned files by the end of the test.  As I recall the peer erroneously not replicating is usually the first node of the departing group (and so the first to depart and reenter) though I believe this is not always the case.  Also to my recollection it is usually only one or two pins that the erroring peer fails to replicate, I believe in all instances of the failure the peer replicates at least one pin correctly.

In the past I have seen the erroring peer display incorrect information about the status of the other departing peer, saying that this peer also does not replicate the pins, though in the current failure that I am documenting this is not the case: here all peers agree that 1 pin is replicated everywhere and that 2 pins are only replicated in 3 nodes.  This does not appear to be a timeout issue.  In this failure instance I query cluster status 15 minutes after launching the test and the failure to replicate is still present.

Looking through the logs one thing to note is that in the departing peer that successfully pinned all 3 files, there are three lines of the form `issuing pin call for <cid>`, whereas in the departing peer that only pins 1 file there is only one line of this form.  It seems that this peer is not being informed of all the pins it should be replicating upon reentry.  Right before the pins are issued in the log of the peer in error, there are a few raft warnings that aren't present in the correctly replicating peer that reenters the cluster.  These warnings are:
`[WARN] raft: Failed to get previous log: 14 log not found (last: 11) raft.go:45`
`[WARN] raft: Previous log term mis-match: ours: 3 remote: 1 raft.go:45`
`[WARN] raft: Previous log term mis-match: ours: 2 remote: 1 raft.go:45`
`[WARN] raft: Clearing log suffix from 10 to 11 raft.go:45`

hsanjuan:
[...]

PR #170, opened 2017-10-11
Fix/raft rejoin errors

ZenGround0:
Addresses #112 by storing peers as bootstrappers when a node is shutdown.  This makes rejoining the cluster pain-free (no chance for bad raft state to error during rejoin) for the departing node.  This strengthens the implicit assumption that departing nodes will rejoin a cluster by default, as starting the departing node in its own cluster will now require modifying the config.

@hsanjuan I may be using `Shadow` in an unconventional way, feedback is welcome.

This PR exposed a flaw in the tests that is still a WIP.  Specifically in k8s-ipfs restarting the cluster is currently handled by reinitializing ipfs-cluster-service before starting the daemon.  This clears listeners bound to sockets that are left behind by the old daemon.  Using the saved config between restarts means reinitializing is not an option, and the network state needs to be cleaned up elsewhere.  I will comment when this is resolved and the full k8s test suite correctly preserves configs.

**edit** 
Upon further inspection it looks like this is just a timing consideration (the re-init just gave more time for system to clean things up).  Newest commit doubles restart timing buffer to help with this.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -423,9 +423,8 @@ func (c *Cluster) Shutdown() error {
 		} else {
 			time.Sleep(2 * time.Second)
 		}
-		/* set c.Config.Bootstrap to current peers */
+		c.config.unshadow()
 		c.config.Bootstrap = c.peerManager.peersAddrs()
-		c.config.Shadow()
 		c.config.Save("")
 		c.peerManager.resetPeers()
 	}
--- a/cluster.go
+++ b/cluster.go
@@ -423,6 +423,9 @@ func (c *Cluster) Shutdown() error {
 		} else {
 			time.Sleep(2 * time.Second)
 		}
+		/* set c.Config.Bootstrap to current peers */
+		c.config.Bootstrap = c.peerManager.peersAddrs()
+		c.config.Save("")
 		c.peerManager.resetPeers()
 	}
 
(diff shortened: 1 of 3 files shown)


cluster.go
```diff
@@ -423,6 +423,10 @@ func (c *Cluster) Shutdown() error {
 		} else {
 			time.Sleep(2 * time.Second)
 		}
+		/* set c.Config.Bootstrap to current peers */
+		c.config.Bootstrap = c.peerManager.peersAddrs()
+		c.config.Shadow()
```
hsanjuan: Hmm yeah, unconventional :)

There is a behaviour change here. Before, if the user passed in a flag like `--leave` but `leave_on_shutdown` was false in the configuration, the configuration would stay as false when the cluster is shudown. With this, however, it will switch to true when this code runs.

I think the cleanest here is to do `c.config.unshadow()` before editing the Bootstrap key. In both cases you are saving right afterwards, and that's when unshadow is supposed to happen anyway. Since you've done it before, whatever you write on `.Bootstrap` will stay.

hsanjuan:
You may have broken sharness too

hsanjuan:
LGTM now, but tests should pass

hsanjuan:
Maybe sharness will be fixed upon rebasing.

@ZenGround0 note that you can work directly on ipfs-cluster branches and don't need to do it on your own fork. This would allow me to push to them too more easily.

ZenGround0:
@hsanjuan ack

PR #171, opened 2017-10-11
Sharness Test Fix: Moving exec call to sh as bash is deprecated in go-ipfs

ZenGround0:
Additionally I found that my system need `ps -a` in order to register that ipfs container was already running and prevent an initialization failure.


ZenGround0:
Hmm sharness now passes but we get "The command "make service && make ctl && ./coverage.sh" exited with 1."

ZenGround0:
Hooray Travis passes!

PR #175, opened 2017-10-12
Adding swagger build + basic annotations

ZenGround0:
This PR gently introduces ipfs-cluster to swagger (OpenAPI) documentation support.  Swagger annotations are added directly in the source and then with the help of the go-swagger tool: https://github.com/go-swagger/go-swagger, they are compiled into a json document.  

This PR only adds basic route descriptions to the rest api endpoints; there is plenty of room for more annotation.  These changes also don't define a sophisticated user interface for the swagger spec.  Currently the spec is generated with the makefile target `make swagger` and can be found in swagger/swagger.json after building.  In the future we could extend our interface to serve a copy over an api endpoint, or probably do all kinds of things I haven't yet thought of.

@hsanjuan I took a stab at this without prior exposure to swagger, so in all likelihood I am missing things here.  Let me know what else we need to get this into mergeable condition.

#documentationFridays

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ gx_bin=$(deptools)/$(gx)
 gx-go_bin=$(deptools)/$(gx-go)
 bin_env=$(shell go env GOHOSTOS)-$(shell go env GOHOSTARCH)
 sharness = sharness/lib/sharness
+swagger = swagger/swagger
 
 # For debugging
 problematic_test = TestClustersReplicationRealloc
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -81,10 +82,26 @@ clean_sharness:
...
+$(swagger):
+	@echo "Installing swagger"
+	@rm -rf swagger
+	@mkdir swagger
+	@go get -u github.com/go-swagger/go-swagger/cmd/swagger
+	@cd $$GOPATH/src/github.com/go-swagger/go-swagger && go install .
+
+clean_swagger:
+	@rm -rf swagger
+	@rm -rf $$GOPATH/src/github.com/go-swagger
```
hsanjuan: This messes too much with the system. It removes swagger, which could have been installed outside cluster. I think probably Makefile should not remove things outside its own project that he cannot be sure are owned by it.

ZenGround0:
Thanks for the feedback.  I'll switch this to add a static spec to the repo.

ZenGround0:
I'll reopen this at a later date when this is a higher priority.  Right now it is just cluttering up the PRs.

Issue #184, opened 2017-10-19
Dockerfile is broken.


countzen2000:
`FROM ipfs/go-ipfs:release`
The line above assumes that go-ipfs is  based on alpine. It is not, therefore the `apk add` no longer works.

`COPY . $SRC_PATH` assumes that the person building the docker image has a git clone of ipfs-cluster already, which may not be true.

The enttypoint.sh assumes existence of `su-exec`.

hsanjuan:
We'll have fixes very soon

PR #185, opened 2017-10-20
Fix #154: Update Dockerfiles

### LINKED ISSUE #154: Study how glibc ipfs docker image affects cluster docker images
See: https://github.com/ipfs/go-ipfs/pull/4219

hsanjuan:
After changes on how the final docker image for go-ipfs is generated,
our Dockerfiles failed to build.

The new version is inspired on go-ipfs Dockerfile. It takes advantage
of docker >= v17.06 which allows multi-stage Dockerfiles with
several FROM directives and COPY --from. This allows build freely on a
first stage, without thinking of cleaning up, and then just copy the
final artifacts to a second, minimal image.

The Dockerfile-test is similarly updated. Minor changes have been
applied to the entrypoints.

License: MIT

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,38 @@
-FROM ipfs/go-ipfs:release
+FROM golang:1.9-stretch AS builder
 MAINTAINER Hector Sanjuan <hector@protocol.ai>
+
+# This build state just builds the cluster binaries
+
+ENV GOPATH     /go
+ENV SRC_PATH   $GOPATH/src/github.com/ipfs/ipfs-cluster
+
+COPY . $SRC_PATH
+WORKDIR $SRC_PATH
+RUN make install
+
+#------------------------------------------------------
+FROM ipfs/go-ipfs
+MAINTAINER Hector Sanjuan <hector@protocol.ai>
+
+# This is the container which just puts the previously
+# built binaries on the go-ipfs-container.
+
+ENV GOPATH     /go
+ENV SRC_PATH   /go/src/github.com/ipfs/ipfs-cluster
+ENV IPFS_CLUSTER_PATH /data/ipfs-cluster
 
 EXPOSE 9094
 EXPOSE 9095
 EXPOSE 9096
 
-ENV GOPATH     /go
-ENV PATH       /go/bin:$PATH
-ENV SRC_PATH   /go/src/github.com/ipfs/ipfs-cluster
-ENV IPFS_CLUSTER_PATH /data/ipfs-cluster
+COPY --from=builder $GOPATH/bin/ipfs-cluster-service /usr/local/bin/ipfs-cluster-service
+COPY --from=builder $GOPATH/bin/ipfs-cluster-service /usr/local/bin/ipfs-cluster-ctl
+COPY --from=builder $SRC_PATH/docker/entrypoint.sh /usr/local/bin/start-daemons.sh
+
+RUN mkdir -p $IPFS_CLUSTER_PATH && \
+    chown 1000:100 $IPFS_CLUSTER_PATH
 
 VOLUME $IPFS_CLUSTER_PATH
-
-COPY . $SRC_PATH
-
-RUN apk add --no-cache --virtual cluster-deps make musl-dev go git \
-    && apk add --no-cache jq \
-    && go get -u github.com/whyrusleeping/gx \
-    && go get -u github.com/whyrusleeping/gx-go \
-    && cd $SRC_PATH \
-    && gx install --global \
-    && gx-go rewrite \
-    && go build \
-    && make -C ipfs-cluster-service install \
-    && make -C ipfs-cluster-ctl install \
-    && cp docker/entrypoint.sh /usr/local/bin/start-daemons.sh \
-    && chmod +x /usr/local/bin/start-daemons.sh \
-    && apk del --purge cluster-deps \
-    && cd / && rm -rf /go/src /go/bin/gx /go/bin/gx-go
-
 ENTRYPOINT ["/sbin/tini", "--", "/usr/local/bin/start-daemons.sh"]
 
 CMD ["$IPFS_CLUSTER_OPTS"]
--- a/Dockerfile-test
+++ b/Dockerfile-test
@@ -1,42 +1,68 @@
 FROM golang:1.9-stretch AS builder
 MAINTAINER Hector Sanjuan <hector@protocol.ai>
 
-# This build state just builds the cluster binaries
+# The testing container needs bash, so we use alpine rather than
+# busybox. That means we need to mimic the whole ipfs-dockerfile rather
+# than just re-use it.
 
 ENV GOPATH     /go
 ENV SRC_PATH   $GOPATH/src/github.com/ipfs/ipfs-cluster
 
+# Install IPFS, su-exec
+ENV SUEXEC_VERSION v0.2
+RUN go get github.com/ipfs/ipfs-update \
+    && ipfs-update install latest \
+    && cd /tmp \
+    && git clone https://github.com/ncopa/su-exec.git \
+    && cd su-exec \
+    && git checkout -q $SUEXEC_VERSION \
+    && make
+
+# install cluster
 COPY . $SRC_PATH
 WORKDIR $SRC_PATH
 RUN make install
-
 #------------------------------------------------------
-FROM ipfs/go-ipfs
+FROM debian:stretch-slim
 MAINTAINER Hector Sanjuan <hector@protocol.ai>
-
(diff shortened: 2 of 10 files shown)


hsanjuan:
#154

hsanjuan:
#184

Dockerfile
```diff
@@ -1,34 +1,38 @@
...
-    && chmod +x /usr/local/bin/start-daemons.sh \
-    && apk del --purge cluster-deps \
-    && cd / && rm -rf /go/src /go/bin/gx /go/bin/gx-go
+RUN mkdir -p $IPFS_CLUSTER_PATH && \
+    chown 1000:100 $IPFS_CLUSTER_PATH
 
+VOLUME $IPFS_CLUSTER_PATH
 ENTRYPOINT ["/sbin/tini", "--", "/usr/local/bin/start-daemons.sh"]
 
 CMD ["$IPFS_CLUSTER_OPTS"]
```
ZenGround0: Out of curiousity, where does $IPFS_CLUSTER_OPTS get set?

docker/entrypoint.sh
```diff
@@ -2,20 +2,20 @@
...
 set -e
 user=ipfs
-repo="$IPFS_PATH"
 
 if [ `id -u` -eq 0 ]; then
+    echo "Changing user to $user"
     # ensure directories are writable
-    su-exec "$user" test -w "$repo" || chown -R -- "$user" "$repo"
-    su-exec "$user" test -w "$IPFS_CLUSTER_PATH" || chown -R -- "$user" "$IPFS_CLUSTER_PATH"
+    su-exec "$user" test -w "${IPFS_PATH}" || chown -R -- "$user" "${IPFS_PATH}"
```
ZenGround0: Again a reminder from countzen2000's comments, we are not changing the assumption of existence of su-exec.

docker/entrypoint.sh
```diff
@@ -2,20 +2,20 @@
...
+    su-exec "$user" test -w "${IPFS_CLUSTER_PATH}" || chown -R -- "$user" "${IPFS_CLUSTER_PATH}"
```
ZenGround0: Out of curiousity, why add the braces?

Dockerfile
```diff
@@ -1,34 +1,38 @@
...
 
-EXPOSE 9094
-EXPOSE 9095
-EXPOSE 9096
+# This build state just builds the cluster binaries
+
+ENV GOPATH     /go
+ENV SRC_PATH   $GOPATH/src/github.com/ipfs/ipfs-cluster
+
+COPY . $SRC_PATH
```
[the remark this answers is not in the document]
hsanjuan: Your understanding is correct. Could it be improved? Possibly. Happy to accept further PRs on that direction.

Dockerfile
```diff
@@ -1,34 +1,38 @@
...
+EXPOSE 9094
+EXPOSE 9095
+EXPOSE 9096
 
-COPY . $SRC_PATH
+COPY --from=builder $GOPATH/bin/ipfs-cluster-service /usr/local/bin/ipfs-cluster-service
+COPY --from=builder $GOPATH/bin/ipfs-cluster-service /usr/local/bin/ipfs-cluster-ctl
+COPY --from=builder $SRC_PATH/docker/entrypoint.sh /usr/local/bin/start-daemons.sh
 
-RUN apk add --no-cache --virtual cluster-deps make musl-dev go git \
```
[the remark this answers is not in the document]
hsanjuan: Before we worried about installing the necessary things to build and removing them in the same RUN step so that Docker did not commit a new step (images are incremental, what's written in RUN cannot be removed later), so we saved size.

Now we create a temporary image (builder) with all the stuff, and copy out the binaries to the actual final one which is a minimal thing. This way:

* We can get rid of all the code that installed/removed things
* We can call `make install` directly
* We don't need to manually add go/make etc bc the golang image on top of which we are building already has them

Dockerfile
hsanjuan: You could set them when running docker `docker run --env IPFS_CLUSTER_OPTS="-d"...`. But if you do `docker run ipfs-cluster <arguments>` the arguments will replace the CMD anyway. I think it just allows some more flexibility.

docker/entrypoint.sh
hsanjuan: But su-exec exists in alpine and go-ipfs image so... I am also assuming /bin/sh exists to run this script... it's not a script for a system where su-exec doesn't exist...

docker/entrypoint.sh
hsanjuan: Just good practice. Ensures no silly errors happen: `"$MY_PATHdefault"` would be substituted for `""` even if your code defines `MY_PATH`. Solution: `"${MY_PATH}default"`.

hsanjuan:
@ZenGround0:  deptools/gx_v0.12.0: 1: deptools/gx_v0.12.0: ����: not found

??? There is something with the encoding of a file or with some weird files.

Does it happen on a clean, freshly checked-out repository? What if you put the repo in a diferent filesystem (like a usb stick) and run it from there?

Dockerfile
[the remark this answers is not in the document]
ZenGround0: @countzen2000 FYI

docker/entrypoint.sh
ZenGround0: @countzen2000 FYI

Dockerfile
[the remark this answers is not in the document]
ZenGround0: Awesome this makes more sense now

ZenGround0:
I see the same error on a fresh repo.  I'm running `docker build -t zenground0/ipfs-cluster:official-test -f Dockerfile-test .`  I'm looking into what's going wrong.  It wouldn't be the first time my machine throws mysterious gx-related errors.

ZenGround0:
Ok, the error was something dumb on my part.  This builds fine from a fresh repo.  
HOWEVER, there is a big kubernetes-ipfs break: bash is no longer on the image so the tests are all broken.  We could add bash here, or we may just want to move kubernetes-ipfs to always use `sh`, as the go-ipfs images no longer support bash either.  What do you think?

docker/entrypoint.sh
hsanjuan: Actually, su-exec is built by hand in the go-ipfs image, but we build on top so it's fine as long as they don't change it. I haven't looked into the implications of not using it though (same as tini)

hsanjuan:
@ZenGround0 I'm going to update the Dockerfile-test back to alpine with bash. I'm not sure the tests can run with sh or ash. We should test that a bit better.

Dockerfile-test
```diff
@@ -32,11 +48,21 @@ COPY --from=builder $SRC_PATH/docker/random-stopper.sh /usr/local/bin/random-sto
...
+    chown 1000:100 $IPFS_PATH && \
     chown 1000:100 $IPFS_CLUSTER_PATH
 
+VOLUME $IPFS_PATH
 VOLUME $IPFS_CLUSTER_PATH
-ENTRYPOINT ["/sbin/tini", "--", "/usr/local/bin/start-daemons.sh"]
 
+ENV IPFS_LOGGING ""
+
+ENTRYPOINT ["/sbin/tini", "--", "/usr/local/bin/start-daemons.sh"]
```
ZenGround0: I didn't catch this earlier, but if we use tini as the entrypoint I believe it is going to intercept signals sent to the testing containers and break existing tests.  To be fair we should be reaping zombie processes in the test container, but I'll need to think about how to achieve this without blocking signals.

hsanjuan:
@ZenGround0 how about now?

Undertaciart:
> After changes on how the final docker image for go-ipfs is generated, our Dockerfiles failed to build.
> 
> The new version is inspired on go-ipfs Dockerfile. It takes advantage of docker >= v17.06 which allows multi-stage Dockerfiles with several FROM directives and COPY --from. This allows build freely on a first stage, without thinking of cleaning up, and then just copy the final artifacts to a second, minimal image.
> 
> The Dockerfile-test is similarly updated. Minor changes have been applied to the entrypoints.
> 
> License: MIT Signed-off-by: Hector Sanjuan [hector@protocol.ai](mailto:hector@protocol.ai)

docker/entrypoint.sh
koxon: Hi. 

I'm using the image ipfs/ipfs-cluster:latest and I just got this error:
/usr/local/bin/entrypoint.sh: line 13: su-exec: not found

su-exec is only used when this condition is met:
"if [ `id -u` -eq 0 ]; then"

Can someone check if this actually works correctly ?

Issue #153, opened 2017-09-12
Unique binary for ctl and server


ferranbt:
Create a unique ipfs-cluster binary to run both ipfs-cluster-ctl and ipfs-cluster-server as in go-ipfs.

hsanjuan:
I'm unconvinced, but maybe my opinion would change with a good list of reasons... right now I don't see that would bring significant advantages and I kind of like the strong division between components:

* `ipfs-cluster` is a go library
* `ipfs-cluster-service` is an application that uses the go library
* `ipfs-cluster-ctl` is a CLI to the REST API component.

ZenGround0:
Thanks @ferranbt, it's good to bring in a fresh perspective.  For what it's worth I also thought combining the binaries would be a good idea when I started working with cluster but have since come to appreciate the structure.  Not sure what that exactly means about the UX tradeoffs, so it would be helpful to understand your thought process a little better to add some data points.

ferranbt:
I am working on an automatic deployment of ipfs-cluster on the cloud. Whenever I deploy ipfs-cluster-service I necessary need to have ipfs-cluster-ctl to check the state of the node. I ended up having to install both binaries. 

Since the Api endpoint in service is right now binded to localhost (correct me if I am wrong) I cannot see the use case when you want to use ctl without having installed service.

Other projects like nomad or consul for example have a single binary to handle both the daemon and  other cli operations.

hsanjuan:
I am thinking that it should not be terribly difficult to have an `ipfs-cluster` binary which wraps and bundles both existing binaries (and then you would call `ipfs-cluster ctl bla` or `ipfs-cluster server bla`). Happy to support if anyone wants to try it out.

ZenGround0:
Same change independently suggested by @VictorBjelkholm in issue #172

hsanjuan:
Labeling this as easy, because compiling the existing binaries as plugins and loading them from a wrapper binary should be an easy exercise to prototype this if anyone wants to do it.

I am still unconvinced that merging an API client to the`service` provides huge benefits. I, for example, like to use `ipfs-cluster-ctl` to remote control cluster nodes and I like that it's lighter, easier to build, faster to load. As a separate application maintenance is clearer and provides a better example on how to interact with the API.

I know there are applications which merge client and servers, but there are also a bunch of applications which do not. Perhaps we can have both anyway.

PR #186, opened 2017-10-20
ipfs-cluster-service: Do not run with unknown subcommands

hsanjuan:
Shows an error when running cluster with an unknown subcommand.

Renames "ipfs-cluster-service run" to "ipfs-cluster-service daemon" which
is consistent with go-ipfs and paves ground for #153.

License: MIT

### DIFF
--- a/ipfs-cluster-service/main.go
+++ b/ipfs-cluster-service/main.go
@@ -227,9 +227,9 @@ configuration.
 			},
 		},
 		{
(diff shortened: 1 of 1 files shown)


ZenGround0:
There are some conflicts with the config PR that I will let you resolve before merging in.

hsanjuan:
ok, should be fixed

PR #188, opened 2017-10-21
Dockerfile-test: Add bash to busybox

hsanjuan:
Remove the debian-stretch Dockerfile-test and put the busybox one
with bash in its place.

Tested that kubernetes-ipfs runs with it.

License: MIT


hsanjuan:
The whole problem about busybox not having bash was this simple to solve:

```
# Add bash
COPY --from=builder /bin/bash /bin/bash
COPY --from=builder /lib/x86_64-linux-gnu/libtinfo.so.5 /lib64/libtinfo.so.5
```

PR #205, opened 2017-10-26
Fix #202: Fix informers and allocators for 32-bit architectures

### LINKED ISSUE #202: The `freespace` informer is broken for 32-bit architectures
It turns out that 10GB of `StorageMax` > than 32bit and we use `int` all around for metrics.

Fix upcoming...

hsanjuan:


ZenGround0:
LGTM

Issue #51, opened 2017-02-27
Upgrading and backups


hsanjuan:
There are a number of pain points if the consensus state format changes upon an upgrade.

Currently, persistence is obtained via Raft snapshots which are loaded on boot and written on shutdown (at least). Raft snapshots format is coming from `go-libp2p-raft` FSMSnapshot implementation, which is just a serialization of the state using msgpack.

If the state changes, loading the snapshot is likely to break. Also, this format is unreadable to the user and hard to work with.

A few thoughts about tackling this:

* [x] ipfs-cluster should store a readable, easy to work with, json backup of the state on shutdown. [DONE]
* [ ] There should be a pain-free procedure (at the very least documented) to load and old backup onto a newer `State` (a migration). This is potentially tricky on a large cluster:
  * Raft snapshots need to be removed [manually?] or Raft will attempt deserializing them on top of a newer version of the state.
  * Raft may also replay log entries on boot, which does not make sense if snapshots have been removed, so probably all Raft data needs to be cleaned up upon upgrade.
  * If Raft starts with a clean state after an upgrade, the new state could load the old backup and migrate it to the new format. With that state, we could directly run a rollback operation (which replaces the whole clean Raft state with the migrated version). However `go-libp2p-consensus` rollbacks are not very specific and this would work only because it's the way `go-libp2p-raft` does it at the moment.
  * An alternative it is to let cluster replay and commit every entry on the old state. This might be slightly cleaner. Replaying would need to make sure allocations remain the same as in the imported state.

ZenGround0:
I am comfortable adding this feature as a subcommand of ipfs-cluster-ctl ex: `ipfs-cluster-ctl migrate`, at least for now.  go-ipfs has a separate tool for this downloaded on the fly, but apparently this is to keep the binary size down (see screenshot) which I don't see as an issue with adding this functionality to ipfs-cluster.

![screen shot 2017-10-30 at 4 18 15 pm](https://user-images.githubusercontent.com/5515260/32193796-c6514518-bd8e-11e7-8a70-c756fa24e4df.png)

PR #220, opened 2017-11-01
Feat/backups upgrade path  -- WIP

ZenGround0:
The basic approach is to read the json backup and replace all raft state with a single snapshot of the new version, while preserving the necessary raft config meta-data.  I added `migration` as an ipfs-cluster-service rather than ipfs-cluster-ctl command because it reinforces to the user that the ipfs-cluster-service daemon should be stopped, reuses some of the config functions already built into this tool, and can't talk to an ipfs-cluster-service daemon endpoint so including it in ipfs-cluster-ctl would break the common pattern.

- Still testing this out
- Needs sharness, possibly other tests
- Still thinking about upgrading general migration framework

--Update--
@hsanjuan The core functionality is here and tests out manually.  When you have a chance your feedback would be awesome.  This is still WIP as I want to add a few things (sharness, cleaner development upgrade path and UX for trying to load bad raft state after an update without a migration).  Thanks!


ipfs-cluster-service/main.go
```diff
@@ -434,3 +531,42 @@ func makeConfigs() (*config.Manager, *ipfscluster.Config, *rest.Config, *ipfshtt
...
+	)
+	if err != nil {
+		return nil, err
+	}
+		
+	bhost := basichost.New(network)
+	return bhost, nil
+}
+	
+func encodeState(state mapstate.MapState) ([]byte, error) {
```
ZenGround0: This is just copy pasted from libp2p-raft/codec.  I'd like your opinion on whether this should be exported from the module or kept inline here.

util.go
```diff
@@ -52,6 +52,10 @@ func copyEmptyStructToIfaces(in []struct{}) []interface{} {
 	return ifaces
 }
 
+func MultiaddrSplit(addr ma.Multiaddr) (peer.ID, ma.Multiaddr, error) {
```
ZenGround0: I shadowed multiaddrSplit with an exported version but probably shouldn't keep it this way.  I wanted to check if you think it is OK to export.

ipfs-cluster-service/main.go
```diff
@@ -231,6 +242,74 @@ configuration.
...
+ 	                        checkErr("loading configuration", err)
+				backupFilePath := c.Args().First()
+				var raftDataPath string
+				if c.NArg() == 1 {
+					raftDataPath = consensusCfg.DataFolder
+				} else {
+					raftDataPath = c.Args().Get(1)
+				}
+				
+				//Migrate backup to new state
```
ZenGround0: If you think there is a better file (maybe I should make one?) to execute these things I am open to suggestions.

consensus/raft/raft.go
```diff
@@ -182,6 +182,10 @@ func makeDataFolder(baseDir, folder string) (string, error) {
 	return folder, nil
 }
 
+func MakeServerConf(peers []peer.ID) hraft.Configuration {
```
ZenGround0: I shadowed makeServerConf with an exported version but probably shouldn't keep it this way. I wanted to check if you think it is OK to export.

ipfs-cluster-service/main.go
```diff
@@ -231,6 +242,74 @@ configuration.
...
+				//Remove raft state
+				err = cleanupRaft(raftDataPath)
+				checkErr("cleaningup Raft", err)
+				
+				//Write snapshot of the migrated state
+				snapshotStore, err := hraft.NewFileSnapshotStoreWithLogger(raftDataPath, 5, nil)
+				checkErr("creating snapshot store", err)
+				dummyHost, err := makeDummyHost(clusterCfg)
+				checkErr("creating dummy host for snapshot store", err)
+				dummyTransport, err := libp2praft.NewLibp2pTransport(dummyHost, consensusCfg.NetworkTimeout)
```
hsanjuan: I think you can use `hraft.InmemTransport` rather than going through the trouble of creating a libp2p one?

ZenGround0:
@hsanjuan I am in agreement with you for the most part.  I've moved the version checking to the state module and moved the raft specific parts of migration to the consensus/raft module.  One place where I could use a little more advice is where you say "Instead of checking if the state needs a migration within the consensus component, you could do it in main.go, before you load anything."

I've been planning this out and if I try to validate the version of the state of existing raft snapshots before loading the consensus in main.go then I think I would end up needing to do a bunch of messy stuff like creating dummy raft components to call HasExistingState and traversing the raft data folder in search of snapshots files to load manually.

Would it be acceptable to check state validity after loading things, i.e. after calling NewCluster from main.go?  This could still be done from within main.go so that the only dependence introduced between consensus/raft and state/mapstate is in raft.Reset(mapstate) and would simplify the check.

Thanks!

consensus/raft/raft.go
```diff
@@ -427,6 +448,147 @@ func (rw *raftWrapper) Peers() ([]string, error) {
...
+	raftSnapVersion = 1         // As of v1.0.0 this is always 1                                     
+	raftIndex       := uint64(1) // We reset history to the beginning                                                    
+	raftTerm        := uint64(1) // We reset history to the beginning
+	configIndex     := uint64(1) // We reset history to the beginning
+	srvCfg := MakeServerConf(peers)
+	sink, err := snapshotStore.Create(raftSnapVersion, raftIndex, raftTerm, srvCfg, configIndex, dummyTransport)
+	if err != nil {
+		return err
+	}
+	newStateBytes, err := encodeState(newState)
```
hsanjuan: let this function receive an already encoded state, perhaps a https://godoc.org/github.com/hashicorp/raft#FSMSnapshot object, which is what the FSM.Snapshot() provides.

Also probably rename to `SnapshotReset`. `Reset` sounds like a thing which could be maybe used by `Consensus` for something less specific in the future.

consensus/raft/raft.go
```diff
@@ -427,6 +448,147 @@ func (rw *raftWrapper) Peers() ([]string, error) {
...
+        snapShotDir := filepath.Join(raftDataDir, "snapshots")
+        err := os.Remove(raftDB)
+        if err != nil {
+		return err
+        }
+        err = os.RemoveAll(snapShotDir)
+        return err
+}
+
+func encodeState(state interface{}) ([]byte, error) {
```
hsanjuan: `encodeState` and `decodeState` should go away (see comment above about using libp2p-raft for this)

ipfs-cluster-service/main.go
```diff
@@ -231,6 +233,57 @@ configuration.
...
+			Usage:  "migrate the IPFS Cluster state between versions",
+			Description: `
+
+This command migrates the internal state of the ipfs-cluster node to the state 
+specified in a backup file. The state format is migrated from the version 
+of the backup file to the version supported by the current cluster version. 
+To succesfully run a migration of an entire cluster, shut down each peer without
+removal, migrate using this command, and restart every peer.
+`,
+			ArgsUsage: "<BACKUP-FILE-PATH> [RAFT-DATA-DIR]",
```
hsanjuan: maybe downcase the args names ? For consistency with `ctl`...

ipfs-cluster-service/main.go
hsanjuan: Perhaps let's put most of this action in a `migrate()` function. `migrate.go` file won't hurt either.

ipfs-cluster-service/main.go
```diff
@@ -290,6 +343,27 @@ func daemon(c *cli.Context) error {
 	checkErr("creating IPFS Connector component", err)
 
 	state := mapstate.NewMapState()
+
+	r, err := raft.ExistingStateReader(consensusCfg) // Note direct dependence on raft here
```
hsanjuan: This would be slightly different (See comment above). I think we can move most of this logic to a function too.

ipfs-cluster-service/main.go
```diff
@@ -328,6 +402,17 @@ func daemon(c *cli.Context) error {
 	return nil
 }
 
+func cleanupRaft(raftDataDir string) error {
```
hsanjuan: oh, this seems dup. Perhaps export from `raft`?

sharness/t0050-service-migration.sh
```diff
@@ -0,0 +1,33 @@
...
+test_cluster_init
+test_create_v1State
+cleanup test_clean_cluster
+
+test_expect_success IPFS,CLUSTER "cluster-service state preserved by migration" '
+    cid=`docker exec ipfs sh -c "echo test | ipfs add -q"`
+    ipfs-cluster-ctl pin add "$cid" &> test4 && sleep 2 &&
+    kill -1 CLUSTER_D_PID && sleep 2 &&
+    backup-file=$( ls test-config/backups/ | head -n 1)
+    ipfs-cluster-service migration "test-config/backups/"$(backup-file) "test-config/ipfs-cluster-data"
```
hsanjuan: Are we forgetting to check if this command is successful ? It would seem that, in case of failure, it will go unnoticed

hsanjuan: if everything is not chained with `&&` and there are no returns, sharness will happily keep going forward.

sharness/t0050-service-migration.sh
```diff
@@ -0,0 +1,33 @@
...
+    ipfs-cluster-service --config "test-config" >"$IPFS_OUTPUT" 2>&1 &
+    ipfs-cluster-ctl pin ls "$cid" | grep -q "$cid" &&
+    ipfs-cluster-ctl status "$cid" | grep -q -i "PINNED"
+'
+
+test_expect_success IPFS,CLUSTER,V1STATE "cluster-service loads v1 state correctly" '
+    cid=`docker exec ipfs sh -c "echo test | ipfs add -q"` 
+    kill -1 CLUSTER_D_PID &&
+    ipfs-cluster-service migration "test-config/v1State" "test-config/ipfs-cluster-data"
```
hsanjuan: same as above

state/interface.go
```diff
@@ -27,4 +27,6 @@ type State interface {
 	Snapshot(w io.Writer) error
 	// Restore restores a snapshot from a reader
 	Restore(r io.Reader) error
+	// Check whether version of state in reader has correct version
+	VersionOk(r io.Reader) bool
```
hsanjuan: Probably it will be more versatile to have a `Version(r io.Reader) int` that tells us the version, and we can do comparison with mapstate.Version outside.

util.go
```diff
@@ -153,3 +161,32 @@ func containsPeer(list []peer.ID, peer peer.ID) bool {
 	}
 	return false
 }
+
+func BackupState(baseDir string, state state.State) error {
```
hsanjuan: I think we can pass *Config here. This function can probably be improved with a "backup retain number" which will be another Config key and who knows what else. So more versatile to just receive the Config.

sharness/t0050-service-migration.sh
```diff
@@ -0,0 +1,33 @@
+#! /bin/sh
```
hsanjuan: Sharness! 👍

ipfs-cluster-service/main.go
```diff
@@ -290,6 +343,27 @@ func daemon(c *cli.Context) error {
...
+		if !valid {
+			logger.Error("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
+                        logger.Error("Raft state is in a non-supported version")
+			err = state.Restore(r)
+			if err == nil {
+				err = ipfscluster.BackupState(clusterCfg.BaseDir, *state)
+			}
+			if err == nil {
+				logger.Error("An updated backup of this state has been saved")
+				logger.Error("to .ipfs-cluster/backups.  To setup state for use")
```
hsanjuan: actually, saved to baseDir/backups...

ipfs-cluster-service/main.go
```diff
@@ -231,6 +233,57 @@ configuration.
 			Usage:  "run the IPFS Cluster peer (default)",
 			Action: daemon,
 		},
+		{
+			Name:   "migration",
```
hsanjuan: migrate? upgrade? or `state upgrade`? Perhaps we should leave room for other state commands (`state inspect`, `state clean`, `state downgrade`, `state export`)...

ZenGround0: I like it!

consensus/raft/raft.go
ZenGround0: @hsanjuan Hmmm correct me if I'm wrong but if this function receives an encoded state that implies the caller (ipfs-cluster-service/main.go) will be creating a dummy libp2p-raft object, loading up its FSM with state and then calling Snapshot().  This is actually what I started with before refactoring after your first round of comments, and I think you were right then that that approach forces the main function to be too raft aware.  

I do get that we would rather keep the encode / decode functionality confined to libp2p raft.  I was thinking maybe we could export Encode and Decode?  Writing the first draft of this showed me that the current libp2p-raft interface makes you jump through hoops to get encodings of states but at the end of the day leaks them out anyway so we wouldn't be changing what users can do, just how easy they can do it.  

Let me know what you think or if I am missing something here.

ZenGround0: Agree with the name change 👍

hsanjuan: Hmm, fair enough, you've convinced me. Let's export Encode and Decode!

You will need to make the change in libp2p-raft, do `gx publish -f` there, then take the hash and `gx update go-libp2p-raft <newhash>` here. Once everything is good we'll merge there, make a release, import a new final hash here. Let me know if you have problems with it.

consensus/raft/raft.go
ZenGround0: I agree but am in favor of exporting them from libp2p raft, see my response above.

hsanjuan:
Heads up, I have published go-libp2p-raft 1.1.1 QmSrgyB4gDbaQnTCSaE9taRxhJkCUc9p5pYFqtXGyTaTbR. It has the exported functions. You should be able to `gx update go-libp2p-raft github.com/libp2p/go-libp2p-raft` to use it.

ZenGround0:
Thanks @hsanjuan, I'll ping you when I'm ready for another round of review.

ZenGround0:
@hsanjuan sharness is killing me today but apart from the failure I like where the code is at.  Could you give it another pass?  

If anything happens to jump out at you that I am doing wrong with Sharness, particularly the second test, I would be very grateful for any help.  These tests pass on my machine albeit with more delay than is required from other tests and with sharness.sh reporting something about Hangup Signals, but consistently fail on Travis.  I've ran the steps manually on a minikube cluster and migration always goes smoothly so I suspect the issue is related to cluster setup/teardown more than the state update.

ipfs-cluster-service/main.go
```diff
@@ -290,6 +315,11 @@ func daemon(c *cli.Context) error {
 	checkErr("creating IPFS Connector component", err)
 
 	state := mapstate.NewMapState()
+
+	if needsUpdate(clusterCfg, consensusCfg) {
+		return errors.New("unsupported state version")
```
hsanjuan: tell user what command to use or what to do in this case

ipfs-cluster-service/migrate.go
```diff
@@ -0,0 +1,91 @@
...
+	peer "github.com/libp2p/go-libp2p-peer"
+	
+	ipfscluster "github.com/ipfs/ipfs-cluster"
+	"github.com/ipfs/ipfs-cluster/state/mapstate"
+	"github.com/ipfs/ipfs-cluster/consensus/raft"
+)
+
+
+func upgrade(c *cli.Context) error {
+	if c.NArg() < 1 || c.NArg() > 2 {
```
hsanjuan: do the argument checking on the Action() and then call upgrade() with the right values

ipfs-cluster-service/migrate.go
```diff
@@ -0,0 +1,91 @@
...
+
+	defer backup.Close()
+	r := bufio.NewReader(backup)
+	newState := mapstate.NewMapState()
+	err = newState.Restore(r)
+	if err != nil {
+		return err
+	}
+	//Record peers of cluster                                                                  
+	var peers []peer.ID
```
hsanjuan: if SnapshotReset does as explained above, it should be happy just with current peer ID. Also this is going to break with my PR since we are going to support multiple addresses per peer ID.

ipfs-cluster-service/migrate.go
```diff
@@ -0,0 +1,91 @@
...
+		return fmt.Errorf("Usage: <BACKUP-FILE-PATH> [RAFT-DATA-DIR]")
+	}
+	//Load configs                                                             
+	cfg, clusterCfg, _, _, consensusCfg, _, _, _ := makeConfigs()
+	err := cfg.LoadJSONFromFile(configPath)
+	if err != nil {
+		return err
+	}
+	backupFilePath := c.Args().First()
+	var raftDataPath string
```
hsanjuan: I'm not super sold on letting the user provide a custom raftDataPath with all the logic it adds and needing to carry it down 3 levels of functions. perhaps just set consensusCfg.DataFolder to the value of raftDataPath?

 Note that consensusCfg.DataFolder is normally empty and that in that case, baseDir+defaultDataSubfolder should be used. That's mostly for the raft component's functions.

ipfs-cluster-service/migrate.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+
+func needsUpdate(cfg *ipfscluster.Config, cCfg *raft.Config) bool {
+	state := mapstate.NewMapState()
+	r, err := raft.ExistingStateReader(cCfg) // Note direct dependence on raft here
```
hsanjuan: following from comments above, this would be like `validSnap, err := raft.LastState(ccfg, &state)

ipfs-cluster-service/migrate.go
```diff
@@ -0,0 +1,91 @@
...
+	if err == nil { //err != nil no snapshots so skip check
```
hsanjuan: if validSnap == false && err == nil -> no snapshots / nothing to do
if validSnap == true && err == nil -> do GetVersion check and maybe tell user migration is needed
if validSnap == true && err != nil -> could not deserialize state from snapshot, migration very probably needed or snapshot corrupted
if validSnap == false && err != nil -> some error before trying to read snapshots

hsanjuan:
@ZenGround0 we have one problem here though, we don't snapshot anymore on raft.Shutdown.

This had problems when a node had been removed from the cluster (snapshotting would hang). In that situation we cannot check for peers, we cannot snapshot, we can only shutdown. Since we cannot request the peerset, it is hard to figure out that we have been removed (if we wanted to skip snapshotting on that case).

This means that:

* there might be a snapshot, but it does not have the latest state because there are some newer entries in the log
* there might not be a snapshot at all, only entries in the log

Either we find a way to raft.Snapshot() on every shutdown or the only other approach I see is to get inspiration from https://github.com/hashicorp/raft/blob/master/api.go#L307 and:

* Try to load last snapshot
* Load the logstore
* Restore snapshot with libp2p-raft.FSM (this does de-serialization)
* Run fsm.Apply for all missing log entries between the snapshotIndex and the log last index
* return (the state should be the last state then)

---

Maybe we are focusing this wrongly. We save a backup on every shutdown. We could directly load the last backup against the current mapstate and check the version. If the last backup has not the right version, then we suggest a migration from it. Then we don't have to load snapshots at all and we just have to make the migrate part (close to what it is now).

On the other side, our backups are inefficient and the snapshot store pretty much does exactly the same. Maybe it would also help if backup format for mapstate is the same as libp2p-raft-snapshot, so it's easier to convert a snapshot to a backup, and backwards, but it's also tying mapstate and raft together.

I'm going to keep giving some thought to this. My feeling is that having a "soft" RecoverCluster which takes the last snapshot or makes a new one, takes the log, commits everything on it and saves it as a newer snapshot, is a useful thing to have. But let's test if we can snapshot right before raft.Shutdown() again.

hsanjuan:
Our main problem here is that we have a mapstate implementation which does not know how consensus->raft->libp2p raft stores states, and a consensus implementation that does not know if the mapstate implementation has a new format.

This https://github.com/libp2p/go-libp2p-raft/pull/29 introduces the possibility for MapState to specify how it wishes to be serialized.

With this, I'm going to take another try at the workflow, assuming snapshots are saved in a format that mapstate can understand directly because it implements Marshal/Unmarhsal itself.

* Let's find a way to save a snapshot on shutdown. I think it might work with `ShutdownOnLeave` set to true (this way attempts will fail when peer has been removed, but not hang at least). We assume the snapshot is the final known state.
* To check if a migration is needed:
  * main.go calls LastSnapshot() which returns the raw snapshot bytes of the last raft snapshot.
  * main.go creates a temp mapstate and calls mapstate.VersionFromBytes(bytes) which just deserializes into a `{ Version int }` object (we assume all our state formats will have a Version key) and tells us the version/ Deserializing the whole thing might fail if the new state format is not backwards compatible. I think it's nicer to just check the version part.
 * if the version from the read snapshot does not match the last Mapstate.Version, a migration is needed. We tell the users that they have Version X but current cluster uses Y.
* To do the migration:
  * Mapstate.Restore from the last snapshot's bytes
  * raft.ResetSnapshot(newstate) - like now, but adding a new snapshot based on the last one - same peers, same indexes, just different contents)

Note that we can forget about backups, since fully rely on raft snapshot store. And we should not cleanRaft() so harshly anymore either on some shutdowns, we should just rename the ipfs-cluster-data folder to something else.

Does this make sense? It's not so different from existing approach, except the snapshot format is now in control of mapstate. and simplifies `LastSnapshot` a bit.

Let's have a chat about it on monday to make sure we are on the same page?

Sorry for the loong notes, this is a very tricky part in cluster and it should work very well so I want a very clean approach (and something that we can cleanly split in the future if we need to).

consensus/raft/raft.go
```diff
@@ -427,9 +434,88 @@ func (rw *raftWrapper) Peers() ([]string, error) {
...
+		return nil, err
+	}
+
+	// Package state from snapshot as json in a reader
+	var state p2pconsensus.State
+	err = p2praft.DecodeState(rawBytes, &state)
+	if err != nil {
+		return nil, err
+	}
+	stateBytes, err := json.Marshal(state)
```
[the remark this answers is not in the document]
ZenGround0: Thank you @hsanjuan for explaining.  To check my understanding, when you say this code is "using insider information" are you referring to this code importing the `consensus.State` type, or my implicit assumption that a consensus.State could be json-marshalled into bytes that themselves will work out to be json-unmarshalled into a valid mapState?  I understand why the second thing is a problem, is the first thing (importing the consensus module)  also a problem?

ipfs-cluster-service/migrate.go
ZenGround0: Thanks, given the above info I'm happy using the default or config value.

consensus/raft/raft.go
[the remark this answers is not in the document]
hsanjuan: > my implicit assumption that a consensus.State could be json-marshalled into bytes that themselves will work out to be json-unmarshalled into a valid mapState

That's the problem. Importing the consensus module is not

ZenGround0:
Thank you for the explanations @hsanjuan.  Some thoughts:
Overall I like it.  I am in full agreement about being less harsh on old raft data, giving mapstate (un)marshal powers, and incorporating info about old snapshots into new ones.

One thing we lose by making raft snapshots the one-and-only backup is the readability and ease of use of the json backups, which was one of the original goals of issue #51.  I think the way to make this happen would be through another `state` sub-command.  For example `state backup` could store a json backup in a different folder by reading from the latest snapshot, and `state load` could create a raft snapshot from the json backup (It seems like it could be useful to specify an entire state at once, for example while testing).

I need to look into why saving snapshots on shutdown is a problem more closely, but your comments above indicate that it is not going to be possible in all circumstances.  Right now it sounds like you are suggesting considering any state changes that is not snapshotted due to bad conditions on shutdown as something cluster should ignore when migrating that peer's state.  Is this correct?  In the first long comment you proposed the idea of a "Soft Recover cluster" function that addressed the problem of failed shutdown snapshots, but this does not seem to be in the proposed workflow.  It seems to me this "soft recover" functionality could happen lazily/in memory in calls to LastSnapshot, or possibly be called from its own `state` subcommand.  It seems to be important and achievable.  Do you think it belongs in this PR?

One final thing to remind ourselves is that by deciding that cluster's state backups are the same as raft snapshots we might be further entrenching ourselves into a version of cluster that becomes harder to generalize to other consensus models.

I really appreciate the long notes.  I'm excited that this is coming together, thanks again for all the feedback!

consensus/raft/raft.go
```diff
@@ -427,9 +434,88 @@ func (rw *raftWrapper) Peers() ([]string, error) {
...
+
+	serverAddr := hraft.ServerAddress(peer.IDB58Encode(peers[len(peers) -1]))
+	_, dummyTransport := hraft.NewInmemTransport(serverAddr)
+	var raftSnapVersion hraft.SnapshotVersion
+	raftSnapVersion = 1         // As of v1.0.0 this is always 1                                     
+	raftIndex       := uint64(1) // We reset history to the beginning
+	raftTerm        := uint64(1) // We reset history to the beginning
+	configIndex     := uint64(1) // We reset history to the beginning
+	srvCfg := makeServerConf(peers)
+	sink, err := snapshotStore.Create(raftSnapVersion, raftIndex, raftTerm, srvCfg, configIndex, dummyTransport)
```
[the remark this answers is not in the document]
ZenGround0: Thanks for this, all great points

ZenGround0:
@hsanjuan I am happy with this apart from the libp2p-raft blocker.  Let me know how it looks.  Thanks!

consensus/raft/raft.go
```diff
@@ -427,9 +435,134 @@ func (rw *raftWrapper) Peers() ([]string, error) {
...
+	err = p2praft.DecodeSnapshot(rawBytes, state)
+	if err != nil {
+		return true, err
+	}
+	return true, nil
+}
+
+// SnapshotMigrate saves a raft snapshot containing a migrated version of the
+// latest snapshot's state.  If no state yet exists there is nothing to migrate
+// and it returns an error.  Old snapshots are cleaned up.
```
hsanjuan: `old raft data is backed up and replaced by the new snapshot` right?

state/mapstate/map_state.go
```diff
@@ -92,22 +97,70 @@ func (st *MapState) Snapshot(w io.Writer) error {
...
+// to the current version in a later call to restore.  Note: Out of date
+// version is not an error
+func (st *MapState) Unmarshal(bs []byte) error {
+	// Check version byte
+	logger.Debugf("The incoming bytes to unmarshal: %x", bs)
+	v := int(bs[0])
+	logger.Debugf("The interpreted version: %d", v)
+	if v != Version { // snapshot is out of date
+		st.Version = v
+		st.toRestore = bs[1:]
```
hsanjuan: I find it a bit indirect to set `.toRestore` here. Is it possible to just return an error and then call:

`st.Restore(bs []byte)`. Or `st.Migrate(bs []byte)`

The "indirection" that bothers me is that it is not clear that Restore can only be called after a failed Unmarshal() attempt.

cluster_config.go
```diff
@@ -392,3 +395,33 @@ func clusterSecretToKey(secret []byte) (string, error) {
 
 	return key.String(), nil
 }
+
+// BackupState backs up a state according to this configuration's options
+func (cfg *Config) BackupState(state state.State) error {
```
hsanjuan: should we get rid if this and assume consensus backs up our state just fine?

consensus/raft/raft.go
```diff
@@ -427,9 +435,134 @@ func (rw *raftWrapper) Peers() ([]string, error) {
...
+	oldStateBytes, err := ioutil.ReadAll(r)
+	if err != nil {
+		return err
+	}
+
+	err = p2praft.DecodeSnapshot(oldStateBytes, newState)
+	if err != nil {
+		logger.Debug("error decoding snapsot")
+		return err
+	}
```
hsanjuan: the code here and up is very similar to LastState()

consensus/raft/raft.go
```diff
@@ -427,9 +435,134 @@ func (rw *raftWrapper) Peers() ([]string, error) {
...
+		return true, err
+	}
+	return true, nil
+}
+
+// SnapshotMigrate saves a raft snapshot containing a migrated version of the
+// latest snapshot's state.  If no state yet exists there is nothing to migrate
+// and it returns an error.  Old snapshots are cleaned up.
+// Only call when Raft is shutdown.
+func SnapshotMigrate(cfg *Config, newState state.State) error {
```
hsanjuan: Do you think it would make sense to do the migrate in main.go. I mean, instead of calling SnapshotMigrate() and let this (which is part of consensus) do everything, do instead:

* `consensus.LastState(cfg, st) []byte` to load the last snapshot
* `st.Restore([]byte)` to migrate it
* `consensus.SaveState(cfg,st)` to write a new snapshot

Or is there something strongly preventing this?

state/interface.go
```diff
@@ -25,6 +25,14 @@ type State interface {
 	Get(*cid.Cid) api.Pin
 	// Snapshot writes a snapshot of the state to a writer
 	Snapshot(w io.Writer) error
-	// Restore restores a snapshot from a reader
-	Restore(r io.Reader) error
+	// Restore restores an outdated state to the current version
+	Restore() error
```
hsanjuan: If Restore() is called from main (see comment above), it would be called on the mapstate directly and maybe doesn't need to be part of the interface (same as Snapshot). I think also GetLatestVersion() might be ok to remove too.

cluster_config.go
ZenGround0: Right now the output is unused and getting rid of this function and its one call won't break anything.  I am neutral about this.  Removing reduces clutter, keeping it around could be useful for future additions to state commands.

consensus/raft/raft.go
ZenGround0: Yes, I'll clear this up!

consensus/raft/raft.go
ZenGround0: Thanks for this.  For some reason I thought there was something strongly preventing this but there actually isn't anything.  I am pretty moving to this will address all of your other comments.

consensus/raft/raft.go
ZenGround0: Will be fixed while addressing previous comment

state/interface.go
ZenGround0: Hmm... fair enough, I am wondering though what kind of state would not want a restore method?  To me it makes sense to keep that one on the interface even if we don't need to.  I will be happy to remove GetLatestVersion() now that I am bringing more things to main.

state/mapstate/map_state.go
ZenGround0: Getting rid of this with the new approach suggested a few comments up 🎉

consensus/raft/raft.go
ZenGround0: This opens up a bit more the question of how unmarshal handles an old version of snapshot.  The current plan is to record the version, throw away the data and return no errors.  If you'd rather throw an error I am ok with that too.

state/interface.go
hsanjuan: Well, the interface is more, 'what are the requirements of the state components so it can be used by cluster'. If cluster does not use the Restore method... but yeah, it doesn't bother me much anyway.

sharness/t0050-service-state-upgrade-from-current.sh
```diff
@@ -0,0 +1,24 @@
...
+
+. lib/test-lib.sh
+
+test_ipfs_init
+cleanup test_clean_ipfs
+test_cluster_init
+cleanup test_clean_cluster
+
+test_expect_success IPFS,CLUSTER "cluster-service state preserved by migration" '
+    cid=`docker exec ipfs sh -c "echo testing | ipfs add -q"` &&
```
hsanjuan: These 3 first don't need to be &&-chained I guess, since whats below will fail if they didn't work

sharness/t0050-service-state-upgrade-from-current.sh
```diff
@@ -0,0 +1,24 @@
...
+    ipfs-cluster-ctl pin add "$cid" && sleep 2 &&
+    kill -1 $CLUSTER_D_PID && sleep 30 &&
+    ipfs-cluster-service --config "test-config" state upgrade &&
+    ipfs-cluster-service --config "test-config" >"$IPFS_OUTPUT" 2>&1 &
+    sleep 50 &&
+    ipfs-cluster-ctl pin ls "$cid" | grep -q "$cid" &&
+    ipfs-cluster-ctl status "$cid" | grep -q -i "PINNED" &&
+    pkill -1 -f "ipfs-cluster-service --config test-config" && sleep 2
```
hsanjuan: I have the feeling we should add a kill() start() functions to the lib so we can do this a bit more cleanly.

sharness/t0051-service-state-upgrade-from-old.sh
```diff
@@ -0,0 +1,35 @@
...
+cleanup test_clean_ipfs
+test_cluster_init
+test_confirm_v1State
+cleanup test_clean_cluster
+
+# Make a pin and shutdown to force a snapshot. Modify snapshot files to specify
+# a snapshot of v1 state pinning "test" (it's easier than taking a new one each
+# time with the correct metadata). Upgrade state, check that the correct
+# pin cid is in the state
+test_expect_success IPFS,CLUSTER,V1STATE,JQ "cluster-service loads v1 state correctly" '
```
hsanjuan: this must have been very hard to get working <3

consensus/raft/raft.go
```diff
@@ -427,9 +435,98 @@ func (rw *raftWrapper) Peers() ([]string, error) {
...
+	if err != nil {
+		return nil, meta, true, err
+	}
+	return rawBytes, meta, true, nil
+}
+
+// SnapshotSave saves the provided state to a snapshot in the
+// raft data path.  Old raft data is backed up and replaced
+// by the new snapshot
+func SnapshotSave(cfg *Config, newState state.State, meta *hraft.SnapshotMeta) error {
```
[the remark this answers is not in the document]
ZenGround0: It certainly could.  I was addressing your earlier comment that `SnapshotSave` unnecessarily duplicates a lot of `LastStateRaw`'s code.  If we don't pass in meta, `SnapshotSave` will duplicate most of `LastStateRaw`.

We could choose not to receive the meta return value from `LastStateRaw` in main, change `SnapshotSave` to not take in meta, and then call `LastStateRaw` this time receiving meta from within `SnapshotSave`.  It would still keep the meta return argument but meta wouldn't be exposed in main.  

I'm pretty neutral about any position in this tradeoff space so let me know what you prefer.

consensus/raft/raft.go
```diff
@@ -427,9 +435,98 @@ func (rw *raftWrapper) Peers() ([]string, error) {
...
+	}
+	dataFolder, err := makeDataFolder(cfg.BaseDir, cfg.DataFolder)
+	if err != nil {
+		return err
+	}
+
+	// make a new raft snapshot
+	var raftSnapVersion hraft.SnapshotVersion
+	raftSnapVersion = 1 // As of hraft v1.0.0 this is always 1
+	raftIndex := meta.Index
```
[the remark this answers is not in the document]
ZenGround0: SGTM, let me know your preference on my response above and I will handle this either when encountering a nil meta or when searching for a snapshot and finding none.

package.json
```diff
@@ -21,9 +21,9 @@
     },
     {
       "author": "hsanjuan",
-      "hash": "QmcjpDnmS6jGrYrTnuT4RDHUHduF97w2V8JuYs6eynbFg2",
+      "hash": "QmWKmeVQuGMKDeLC6BEYE6Xz6mSUKrzDZALfcdAaVpurtX",
```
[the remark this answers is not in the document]
ZenGround0: Thanks for the gx publish!

sharness/t0050-service-state-upgrade-from-current.sh
```diff
@@ -0,0 +1,24 @@
...
+test_cluster_init
+cleanup test_clean_cluster
+
+test_expect_success IPFS,CLUSTER "cluster-service state preserved by migration" '
+    cid=`docker exec ipfs sh -c "echo testing | ipfs add -q"` &&
+    ipfs-cluster-ctl pin add "$cid" && sleep 2 &&
+    kill -1 $CLUSTER_D_PID && sleep 30 &&
+    ipfs-cluster-service --config "test-config" state upgrade &&
+    ipfs-cluster-service --config "test-config" >"$IPFS_OUTPUT" 2>&1 &
+    sleep 50 &&
```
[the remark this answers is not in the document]
ZenGround0: I have had a lot of trouble with cluster starting up slowly the second time around and needing this long sleep.  It may just be my local setup so I will try it out.

sharness/t0050-service-state-upgrade-from-current.sh
ZenGround0: Hmm yeah I agree with you.  I tried adapting the current clean_cluster lib function to take in a pid for this test, but I ran back into the problem that within sharness executable strings `$!` doesn't actually record the latest spawned process.  So `pkill` seems like the best I can do right now.  I am cool with adding a `start` and `kill` that uses `pkill`.

consensus/raft/raft.go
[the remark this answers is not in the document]
hsanjuan: maybe make a private LastStateRaw-like function which provides the meta, and call from both LastStateRaw and SnapshotSave ?

sharness/t0050-service-state-upgrade-from-current.sh
hsanjuan: start() can save the pid and kill can re-use that, if both are in the lib/test_lib.sh file. it works like that now (it kills on clean up)

ZenGround0:
Currently Travis always fails on test t0051 on travis, however this test consistently passes on my machine.  It has been tough to see exactly what is going on because output is not carrying over on restart for some reason.  One possibility that comes to mind is that t0051 accesses  a new data folder through relative paths, maybe travis doesn't allow this?

ipfs-cluster-service/migrate.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Reset with SnapshotSave
+	err = raft.SnapshotSave(consensusCfg, newState, clusterCfg.ID)
+	if err != nil {
+		return err
+	}
+	return nil
```
hsanjuan: yeah this workflow looks very nice and straightforward now :)

hsanjuan:
@ZenGround0 this option needs to be `true` very possibly to avoid snapshot hanging when a peer is removed:

https://github.com/ipfs/ipfs-cluster/blob/master/consensus/raft/config.go#L233

consensus/raft/raft.go
```diff
@@ -427,9 +434,106 @@ func (rw *raftWrapper) Peers() ([]string, error) {
 	return ids, nil
 }
 
-// only call when Raft is shutdown
-func (rw *raftWrapper) Clean() error {
-	dbh := newDataBackupHelper(rw.dataFolder)
+// latestSnapshot looks for the most recent raft snapshot stored at the
+// provided basedir.  It returns a boolean indicating if any snapshot is
+// readable, the snapshot's metadata, and a reader to the snapshot's bytes
+func latestSnapshot(raftDataFolder string) (bool, *hraft.SnapshotMeta, io.ReadCloser, error) {
```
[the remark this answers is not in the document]
ZenGround0: Good catch, looks like we can use meta!=nil as the boolean flag.

ZenGround0:
@hsanjuan, got it.  Should I work on changing the default in this PR?

ZenGround0:
@hsanjuan I think this is finally ready.  Let me know if there is more to do here before I can merge.

hsanjuan:
Thanks @ZenGround0 ! Great job here, this was a very complex change and I'm very glad we found a nice approach to it.

PR #232, opened 2017-11-13
Add the packaging metadata to build the ipfs-cluster snap

elopio:
This package will let you publish the latest ipfs-cluster in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and Linux distributions. You just have to go follow https://tutorials.ubuntu.com/tutorial/continuous-snap-delivery-from-travis-ci and enable the automated continuous delivery.

### DIFF
--- a/snap/patches/snap-home.patch
+++ b/snap/patches/snap-home.patch
@@ -1,0 +1,26 @@
+diff --git a/ipfs-cluster-service/main.go b/ipfs-cluster-service/main.go
+index 80561f6..7144a35 100644
+--- a/ipfs-cluster-service/main.go
++++ b/ipfs-cluster-service/main.go
+@@ -6,7 +6,6 @@ import (
+ 	"fmt"
+ 	"os"
+ 	"os/signal"
+-	"os/user"
+ 	"path/filepath"
+ 	"syscall"
+ 
+@@ -116,12 +115,8 @@ func init() {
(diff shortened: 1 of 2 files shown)


elopio:
So, there are two reasons why I didn't propose this before and instead started maintain it myself. First, we have a bug in the builders of https://build.snapcraft.io that doesn't let us download the gx dependencies:

https://bugs.launchpad.net/launchpad-buildd/+bug/1688624

The workaround for that is to use a different system to build. I configured travis to do it. I can't do it for you on this repo because it needs credentials to add a secret to the travis settings. I could keep building and releasing from my travis account, or you can follow this: https://tutorials.ubuntu.com/tutorial/continuous-snap-delivery-from-travis-ci

The other problem is a hole in the way we isolate applications. It's not safe to let all the applications read and write to user home, specially to hidden directories like .gnupg that will contain private keys. So, the snap sandbox redirects $HOME to $HOME/snap/ipfs-cluster. The problem is that go goes directly to /etc/passwd, ignoring $HOME:

https://forum.snapcraft.io/t/go-homedir-reads-etc-passwd-not-home/2727

We need to fix both issues. But now that we have an external project that is affected, it's easier to raise their priority.

I'm available on #ipfs-cluster IRC channel in case you have questions.

pura vida

elopio:
\o/ thanks!

PR #233, opened 2017-11-13
Raft: do not ever remove state, rename it and leave it around

hsanjuan:
This commit changes the way that consensus.Clean() works. Before
it deleted the whole data folder. Now it renames it as <name>.old.0
and leaves it. When Clean() is called again, it renames <name>.old.0
as <name>.old.1, and the actual data becomes <name>.old.0. Higher number
means older. The number of backups is fixed to 5. When 5 backups exists
and a new one comes up again, the last one is discarded.

License: MIT

### DIFF
--- a/consensus/raft/data_helper_test.go
+++ b/consensus/raft/data_helper_test.go
@@ -24,10 +24,9 @@ func TestDataBackupHelper(t *testing.T) {
 			t.Fatal(err)
 		}
 		backups := helper.listBackups()
-		if i < RaftDataBackupKeep && len(backups) != i+1 {
-			t.Fatal("not saving enough backups")
-		} else if i >= RaftDataBackupKeep && len(backups) != RaftDataBackupKeep {
-			t.Fatal("saving too many backups")
+		if (i < RaftDataBackupKeep && len(backups) != i+1) ||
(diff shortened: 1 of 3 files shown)


consensus/raft/data_helper_test.go
```diff
@@ -0,0 +1,34 @@
...
+	os.MkdirAll("data_helper_testing", 0700)
+	helper := newDataBackupHelper("data_helper_testing")
+	for i := 0; i < 2*RaftDataBackupKeep; i++ {
+		err := helper.makeBackup()
+		if err != nil {
+			t.Fatal(err)
+		}
+		backups := helper.listBackups()
+		if i < RaftDataBackupKeep && len(backups) != i+1 {
+			t.Fatal("not saving enough backups")
```
ZenGround0: If something really weird happened with the helper and too many backups were saved before iteration `RaftDataBackupKeep` or alternatively too few backups were saved after iteration `RaftDataBackupKeep`, then the error messages will be misleading.  As the check stands, a better message would be "incorrect number of backups saved"

hsanjuan: ok fixed

ZenGround0:
LGTM

hsanjuan:
Thanks!

PR #234, opened 2017-11-14
Enable travis for snapcraft builds

hsanjuan:
@elopio this should make it. Can you transfer the snap to "hsanjuan" on the snap store?


hsanjuan:
@elopio ok done!

hsanjuan:
@elopio I have upstreamed your patch (see last commit)

elopio:
<3 thanks for patching the code!
I transferred the name. Everything seems correct here, but let me know in case you didn't get it.

And now I'm wondering how the enable-ci command worked for you if you didn't own the name. Hum, I need to give that a test, maybe the credential

hsanjuan:
I'm gonna merge this and see if it works, otherwise I'll re-run the process. I thought the generated key would just be associated to the account and that it only fails on push from ci if you don't own the snap. But now I do so it might just work.

(don't worry about the test failures, they are getting fixed on a different PR)

hsanjuan:
ah no, I actually broke the code :) let me fix that

elopio:
Hum, yeah, that's probably a bug, we should fail early if the snap is not yours. Thanks a lot for helping us test this. Any suggestions you have will be very useful.

hsanjuan:
@elopio is this normal?:

```
[14:09:48] ~ $ snap find ipfs-cluster
The search "ipfs-cluster" returned 0 snaps
```

are there any extra steps to include the snap in the search? It seems that they are correctly published...

elopio:
Yes, it is normal. Take a look at:

```
$ snap info ipfs-cluster
name:      ipfs-cluster
summary:   Collective pinning and composition for IPFS
publisher: hsanjuan
description: |
  ipfs-cluster allows to replicate content (by pinning) in multiple IPFS nodes.
snap-id:     LRO8NhaZ4hXMqMQ3LzCLilNYYxTWKBO5
channels:               
  stable:    –              
  candidate: –              
  beta:      –              
  edge:      master (3) 9MB -
```

So, you have four channels. The idea is to organize your manual and automated testing in edge, beta and candidate, and get your early adopters to accept different levels of risk depending on how much time they want to dedicate to your project, how willing are they to experience downtimes and how eager are they to try new features. Once you are confident that a revision is good, you move it to the stable channel and it will be discoverable by everybody.

For example, look at ipfs today:
```
$ snap info ipfs
[...]
channels:                  
  stable:    v0.4.11 (275) 10MB -
  candidate: v0.4.12 (313) 10MB -
  beta:      v0.4.12 (313) 10MB -
  edge:      master  (315) 10MB -
$ snap find ipfs
Name                Version  Developer  Notes  Summary
ipfs                v0.4.11  elopio     -      global, versioned, peer-to-peer filesystem
```

There is one thing that will make your live easier with respect to versioning. If you use `version: git` on the snapcraft.yaml, then when you publish a tag, the version of the snap will be the name of that tag, instead of the static `master` that we currently have here. This requires the use of [annotated tags](https://stackoverflow.com/questions/4971746/why-should-i-care-about-lightweight-vs-annotated-tags), which is the right thing to do anyway (IMO).

hsanjuan:
Thanks, well I'll do an annotated-signed tag for the next release so we can try it

elopio:
Good deal! I made #239 to get ready for that tag.

PR #236, opened 2017-11-14
Fix/minor fixes

hsanjuan:
A couple of really small things that came up while I was looking at the code while documenting...


hsanjuan:
See commit messages for extended description

hsanjuan:
I think this fixes many random test failures. @ZenGround0 you may want to rebase #220 on master after I merge.

PR #238, opened 2017-11-15
cluster: safeguard consensus not set when calling ID

hsanjuan:
SwarmConnect on the ipfs connector calls rpc Peers() which
requests IDs for every peer member. If that peer member
is booting, it might get the request after RPC is setup
but before consensus is initialized. In which case
a panic happens. Probability that this happens is small, but still.

Also increase the connect swarms delay to 30 seconds, which
should be a bit longer than the default wait_for_leader timeout,
otherwise we might connect swarms while there's not even a leader.

License: MIT

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -586,7 +586,12 @@ func (c *Cluster) ID() api.ID {
 		addrs = append(addrs, multiaddrJoin(addr, c.id))
 	}
 
-	peers, _ := c.consensus.Peers()
+	peers := []peer.ID{}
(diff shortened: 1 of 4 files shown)


hsanjuan:
sorry @ZenGround0 , just added something to fix the test that failed (which is not related, and randomly fails, but still)... so you need to re-approve

hsanjuan:
@ZenGround0 look at all that greenness :) thanks again!

PR #242, opened 2017-11-16
Move the snap deploy script to a separate file

elopio:
I tested this in my branch:
 
 * amd64: https://travis-ci.org/elopio/ipfs-cluster/builds/302877222#L2028
 * armhf: https://travis-ci.org/elopio/ipfs-cluster/builds/302877222#L2890
 * arm64 https://travis-ci.org/elopio/ipfs-cluster/builds/302877222#L3757

### DIFF
--- a/snap/snap-in-docker.sh
+++ b/snap/snap-in-docker.sh
@@ -1,7 +1,0 @@
-#!/bin/bash
-#
-# Build the ipfs-cluster snaps and push them to the store using a docker container.
-
-set -ev
-
-docker run -v $(pwd):$(pwd) -t snapcore/snapcraft sh -c "apt update -qq && cd $(pwd) && ./snap/snap-multiarch.sh"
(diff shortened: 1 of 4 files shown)


elopio:
While testing this in arm I realized that the variables where not being passed correctly, so I added another file to do that in a clearer way.

For now I removed i386, because it was failing, I don't have any hardware at hand to test it, and it's late.

snap/snap-in-docker.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+#
+# Build the ipfs-cluster snaps and push them to the store using a docker container.
+
+set -ev
+
+docker run -v $(pwd):$(pwd) -t snapcore/snapcraft sh -c "apt update -qq && cd $(pwd) && ./snap/snap-multiarch.sh"
```
hsanjuan: I think you can just call this line directly in `.travis.yml`

elopio: You are right, thanks for catching that.

elopio:
crossing fingers...
Little did I know what was I getting into when I started looking at https://github.com/elopio/ipfs-snap/issues/2 this week :)

hsanjuan:
:) yes I was doing my own testing and saw you've probably hit every nitty picky error to come to the current formulation explicitally setting the compilers etc...

PR #252, opened 2017-11-28
Feat/add pin name

### LINKED ISSUE #249: Provide Names for Pins to Cluster
There doesn't seem to be a good way to name pins on ipfs-cluster. I find the context of a name for pins to provide useful information about what I have pinned to my cluster.

Perhaps this can be added to the stored consensus or handled via the underlying pinned ipfs object.

te0d:
This adds a Name attribute to Pins. It exposes a URL parameter to set the name and returns the name via the allocations API call. A command line argument added for ipfs-cluster-ctl. I've also updated the map_state version because I believe this change would require a migration.

This PR fixes #249 by adding storing the name in the consensus layer.

These changes do not touch the output of the status command.

### DIFF
--- a/ipfs-cluster-ctl/main.go
+++ b/ipfs-cluster-ctl/main.go
@@ -249,12 +249,17 @@ peers should pin this content.
 							Value: 0,
 							Usage: "Sets a custom replication factor for this pin",
 						},
+						cli.StringFlag{
+							Name:  "name, n",
+							Value: "",
+							Usage: "Sets a name for this pin",
+						},
 					},
 					Action: func(c *cli.Context) error {
 						cidStr := c.Args().First()
 						_, err := cid.Decode(cidStr)
 						checkErr("parsing cid", err)
-						query := fmt.Sprintf("?replication_factor=%d", c.Int("replication"))
+						query := fmt.Sprintf("?replication_factor=%d&name=%s", c.Int("replication"), c.String("name"))
 						resp := request("POST", "/pins/"+cidStr+query, nil)
 						formatResponse(c, resp)
 						if resp.StatusCode < 300 {
--- a/ipfs-cluster-ctl/main.go
+++ b/ipfs-cluster-ctl/main.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


te0d:
Also note that the names are not unique identifiers and are optional.

state/mapstate/map_state.go
```diff
@@ -15,7 +15,7 @@ import (
 
 // Version is the map state Version. States with old versions should
 // perform an upgrade before.
-const Version = 2
+const Version = 3
```
hsanjuan: I think if you rebase on master, this probably causes an error message when launching cluster and ask the user to upgrade. You'd need to add a migration here:

https://github.com/ipfs/ipfs-cluster/blob/master/state/mapstate/migrate.go#L17

What the migration is going to do is simply pretty much the same as Unmarshal (https://github.com/ipfs/ipfs-cluster/blob/master/state/mapstate/map_state.go#L159). But it needs to do it there in the migrate code (and it cannot re-use unmarshal because if it changed in the future, the migration from version 2 of the state would fail).

We just added migration handling, so I just think this is an opportunity to give it a test. Would you be up for it? Otherwise, just leave Version = 2, which will work anyways (I think).

ipfs-cluster-ctl/main.go
```diff
@@ -249,12 +249,17 @@ peers should pin this content.
...
+							Value: "",
+							Usage: "Sets a name for this pin",
+						},
 					},
 					Action: func(c *cli.Context) error {
 						cidStr := c.Args().First()
 						_, err := cid.Decode(cidStr)
 						checkErr("parsing cid", err)
-						query := fmt.Sprintf("?replication_factor=%d", c.Int("replication"))
+						query := fmt.Sprintf("?replication_factor=%d&name=%s", c.Int("replication"), c.String("name"))
```
hsanjuan: Need to do url-escaping stuff with name, otherwise some names are gonna go wrong...

state/mapstate/map_state.go
ZenGround0: Mostly in agreement.  I think this may cause issues when starting with existing v2 formatted raft state even if you keep `Version = 2`.  I had some issues getting the serialized v1 state to decode into a v2 state, and I suspect the change to pinSerial here will cause a similar problem between v2 and v2'.  But as the current v2 format has been out for under a day and not bundled into a release, I'm doubting such a break is a real problem.

te0d: Yes, I'll look into upping to `Version = 3` and adding a migration.

te0d: So I've started a clean single-node cluster on master and pinned an item. I then shutdown the cluster, switched to feat/add-pin-name with `Version = 2` and did not observe an issue on startup or listing pins.

Doing the same (without a migration) except keeping the `Version = 3` change appears to cause an issue where I am unable to list items I previously had pinned.

I'll see what happens in the second scenario after I add a migration, but would it be best to leave `Version = 2` if that doesn't cause issues?

te0d:
Didn't get the chance to play around with the migration much, but I think that Version = 2 is working appropriately. Even if it isn't, it should be ok to leave the version the same since it has not his a stable release yet?

te0d:
@hsanjuan Sorry, think I just dismissed your review because I noticed your review for my other PR which suggested removing single quotes from the ipfs-cluster-ctl formatter. Just snuck that in here since I did the same thing in this PR.

PR #257, opened 2017-11-29
Added Hostname Property to Configuration

### LINKED ISSUE #250: Provide Identifier for Cluster Peers
Currently peers only seem to be identified by their hash and addresses. I suggest adding a human-readable identifier which can make it easier for users to recognize peers.

Perhaps this can be stored in consensus or inferred by some means.

te0d:
I added a "hostname" property to a node's configuration file. Its
value defaults to the hostname provided by the OS, but can be
modified to anything beside an empty string in the config file.

The "hostname" was added to the output of the "id" call. Thus, peer
hostnames are available when listing peers.

Fixes #250


te0d:
Note that this does not preserve a peer's hostname. As it is only present in the ID request reponse from the peer, the identifier is not known if the peer cannot connect.

api/types.go
```diff
@@ -239,6 +239,7 @@ type ID struct {
 	RPCProtocolVersion    protocol.ID
 	Error                 string
 	IPFS                  IPFSID
+	Hostname              string
```
hsanjuan: Let's not name this `Hostname` because it is definitely not a hostname in the DNS-sense of it (it does not matter if it's resolvable and the users can set it to whatever arbitrary value they want). Let's name it `Alias` or simple `Name` (I'm undecided, you, @ZenGround0 ?)

cluster_config.go
```diff
@@ -226,6 +234,10 @@ func (cfg *Config) LoadJSON(raw []byte) error {
 	}
 	cfg.ID = id
 
+	if jcfg.Hostname != "" {
```
hsanjuan: you can use `config.SetIfNotDefault(jcfg.Hostname, &cfg.Hostname)`

cluster_config.go
```diff
@@ -194,6 +199,9 @@ func (cfg *Config) Validate() error {
 
 // this just sets non-generated defaults
 func (cfg *Config) setDefaults() {
+	hostname, _ := os.Hostname()
```
hsanjuan: For politeness and explicitness, read the error, and if it's not nil, set to empty string.

ipfs-cluster-ctl/formatters.go
```diff
@@ -81,7 +81,7 @@ func textFormatPrintIDSerial(obj *api.IDSerial) {
 		return
 	}
 
-	fmt.Printf("%s | Sees %d other peers\n", obj.ID, len(obj.ClusterPeers)-1)
+	fmt.Printf("%s | '%s' | Sees %d other peers\n", obj.ID, obj.Hostname, len(obj.ClusterPeers)-1)
```
hsanjuan: better not single-quote the alias. I'm thinking when running a silly grep + cut and and piping things it's causes an extra step to remove quotes

api/types.go
ZenGround0: Either SGTM

te0d:
Sounds good, I'll look into making these changes and add a test.

api/types.go
hsanjuan: `Peername` ?

te0d: Peername sounds good to me so I've changed it to that. Easy enough to change it if we want it to be something else.

Issue #254, opened 2017-11-29
state upgrade should not happen when a cluster-service is running


hsanjuan:
`ipfs-cluster-service state upgrade` (or any operations) should not be attempted when cluster is running.

This will need that we keep a PID file in the fashion of https://github.com/moby/moby/blob/master/pkg/pidfile/pidfile.go or a lock file (should check how ipfs does it).

Should allow to easily add `ipfs-cluster-service shutdown` too.

PR #262, opened 2017-12-05
ipfs-cluster-service lock

ZenGround0:
This addresses issue #254.  It uses the same locking mechanism as go-ipfs 

A few things I can see right now that could use some examination:
- the lock file is pretty modular but still in the main package, maybe it belongs in its own package?
- my way of handling any operation with nonexistant basedir is to not lock (I did this to get init ops to work properly upon creation of a new config folder)
- unlock is a noop if there was no lock, I liked this solution but as it was tricky maybe I am missing an issue with it


ipfs-cluster-service/lock.go
```diff
@@ -0,0 +1,69 @@
...
+// Global file, close to relinquish lock access
+var lockCloser io.Closer
+
+func lock(lockPath string) error {
+	// set the lock file within this function
+	logger.Debug("Checking lock")
+	lk, err := filelock.Lock(path.Join(lockPath, lockFileName))
+	if err != nil {
+		lockCloser = nil
+		if err == syscall.EAGAIN {
```
hsanjuan: I think we are fine just logging the error message whatever it is with logger.Debug() and returning it, without needing to do error parsing.

Right now, the message is shows to the user is:

```
cannot acquire lock: Lock FcntlFlock of /home/hector/go/src/github.com/ipfs/ipfs-cluster/ipfs-cluster-service/d1/cluster.lock failed: resource temporarily unavailable
```

I think regardless of the reason of the failure we can let the user know something like `could not obtain execution lock. If no other cluster process is running, remove ~/path/to/lock, or make sure that folder is writable for the user running cluster. Run with -d for more information about the error`. Perhaps written nicely with line-breaks etc.

ipfs-cluster-service/lock.go
```diff
@@ -0,0 +1,69 @@
...
+
+func unlock() error {
+	// Noop in the uninitialized case
+	if lockCloser == nil {
+		logger.Debug("locking not initialized, unlock is noop")
+		return nil
+	}
+	err := lockCloser.Close()
+	if err != nil {
+		logger.Debug("Error unlocking ipfs-cluster-service lock on shutdown")
```
hsanjuan: might not be `on shutdown` technically, you can remove that part

ipfs-cluster-service/main.go
```diff
@@ -276,7 +277,16 @@ removal, upgrade state using this command, and restart every peer.
 		if c.Bool("debug") {
 			setupDebug()
 		}
-		return nil
+
+		if _, err := os.Stat(absPath); err != nil {
```
hsanjuan: I think abspath is always going go be set here so you'll have a place to lock.

I would add a hidden --lockfile flag which pretty much gets the same treatment as the `config` flag (default initialization etc) but eventually allows to customize the lock location.

ipfs-cluster-service/main.go
```diff
@@ -276,7 +277,16 @@ removal, upgrade state using this command, and restart every peer.
...
+			// lockfile dir not init? don't lock
+			return nil
+		}
+		return lock(absPath)
```
hsanjuan: I think we don't need to lock every time. Things like `cluster-service version` don't need locking. A future `cluster-service shutdown` will need to run while locked anyway. This means you'd lock at the start of the commands that need locking (`deamon`, `migrate`) and you can `defer unlock()`

ipfs-cluster-service/lock.go
```diff
@@ -0,0 +1,69 @@
...
+	"syscall"
+
+	filelock "go4.org/lock"
+)
+
+// The name of the file used for locking
+const lockFileName = "cluster.lock"
+
+// Global file, close to relinquish lock access
+var lockCloser io.Closer
```
hsanjuan: We should avoid global variables. Let's wrap the locking implementation in a `LockHelper` or `Locker` object which can lock/unlock.

ipfs-cluster-service/main.go
hsanjuan: Also, calling checkErr() on a locking error will ensure only the error message is printed, right now, it prints all the help when locking fails, which makes it difficult to realize the error is printed at the very top.

sharness/t0020-service-basic-commands.sh
```diff
@@ -27,4 +27,8 @@ test_expect_success "cluster-service --version succeeds and matches ctl" '
     [ "$SERV_VERSION" = "$CTL_VERSION" ]
 '
 
+test_expect_success "starting a second cluster-service process fails" '
```
hsanjuan: 👍

hsanjuan:
> the lock file is pretty modular but still in the main package, maybe it belongs in its own package?

We just need a very lightweight wrapper around the already packaged `lock` module.

>    my way of handling any operation with nonexistant basedir is to not lock (I did this to get init ops to work properly upon creation of a new config folder)

As pointed above, should only lock in operations that need locking, and those already have an existing basedir (or should, you can always lock after config was sucessfully read)

>    unlock is a noop if there was no lock, I liked this solution but as it was tricky maybe I am missing an issue with it

should probably warn, since there should not be a code path that leads you to unlocking without having locked first.

ipfs-cluster-service/lock.go
ZenGround0: Sounds good, are you suggesting modifying the error message for greater readability as well, or only adding more logging?

hsanjuan: I'm suggesting we give an error message that tells clearly the problem and solutions, and we logger.Debug() the original one as it comes from the library.

ZenGround0:
Hey no worries, I'm on it!

ZenGround0:
A question about style: now that I am going back to calling unlock from a global locker in `checkErr`, the quickest rewrite would simply call unlock each time an error causes an exit.  This will work as unlock is a nop without a preceding lock, but it runs counter to your earlier comment that unlock should not run in a code path that has not yet called lock and should generate a warning in this case.  Would it help if I renamed unlock to something like `tryUnlock`?

hsanjuan:
True, so just run a noop silently there (or just with debug logging).

ZenGround0:
@hsanjuan how is this looking?

hsanjuan:
I'm curious that the build log shows that sharness is run before regular testing, and that sharness prints outputs even though they are redirected to /dev/null and we haven't touched anything :S

hsanjuan:
i totally missed that in the diff :P ok all good

GitCop:

 * Commit: 1747e05f8485f3e0a272e82a77b4a926f0e9a530
    **License: MIT  
Signed-off-by:**

---

GitCop:

 * Commit: de8ffb4a5b78c15a1700155f5c5c1d8404a3b3e2
    **License: MIT  
Signed-off-by:**

---

GitCop:

 * Commit: de8ffb4a5b78c15a1700155f5c5c1d8404a3b3e2
    **License: MIT  
Signed-off-by:**

---

PR #265, opened 2017-12-07
Fix: client requests and multiaddresses parsing

hsanjuan:
I'm seeing some errors in real cluster with ipfs-cluster-ctl and I think they might be because of this...

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,19 @@
+ratings:
+  paths: 
+  - "**/*.go"
+
+engines:
+  fixme:
+    enabled: true
+    config:
+      strings:
+      - FIXME
+      - HACK
+      - XXX
+      - BUG
+  golint:
+    enabled: true
+  govet:
+    enabled: true
+  gofmt:
+    enabled: true
--- a/api/rest/client/request.go
+++ b/api/rest/client/request.go
@@ -1,7 +1,6 @@
 package client
 
 import (
-	"context"
 	"encoding/json"
 	"io"
 	"io/ioutil"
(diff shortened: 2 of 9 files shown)


GitCop:

 * Commit: 033e029e18fb36a5dd544f4813f1f8a15f8136fc
    **License: MIT  
Signed-off-by:**

---

ZenGround0:
Is there a downside to doing the http request without a context?  Is it just that now we can't cancel on our own from `doRequest` (which we weren't doing in a useful way anyway, just on function exit)? 

Also I'm curious, what was the error you saw?

hsanjuan:
I was getting:

```
Code: 200
Messsage: context cancelled
```
and sometimes EOF errors. It seems that, since the request context was cancelled at the end of that function and the body is read later, if would error when attempting to read the body (somehow it doesn't happen locally but it happens in real deployment, not sure why). It seems that cancelling the request context prematurely closes the body. Using a context there was redundant anyway, as the client now has a Timeout parameter set.

I have another commit coming up here

hsanjuan:
Ready for review

hsanjuan:
ok, I made code climate happy after some suffering. leaving this as it is now and hoping tests pass.

.codeclimate.yml
```diff
@@ -0,0 +1,19 @@
+ratings:
+  paths: 
+  - "**/*.go"
+
+engines:
+  fixme:
+    enabled: true
+    config:
+      strings:
```
ZenGround0: What about TODO?

hsanjuan: I included and it gave me a bunch of errors for every TODO in the code, which are not things that can be fixed so easily. So I took it out, at least until we can go over the list. We should do the TODOs, but also shouldn't feel discouraged to mark things with TODO comments in the code for things which are not urgent just because codeclimate is going to complain.

PR #266, opened 2017-12-08
Feat/client features

hsanjuan:
* Auto-resolve dns multiaddresses
* Support Host/Port as strings in the config
* Return regular errors from methods

### DIFF
--- a/api/rest/client/client.go
+++ b/api/rest/client/client.go
@@ -2,11 +2,13 @@ package client
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"time"
 
 	logging "github.com/ipfs/go-log"
 	ma "github.com/multiformats/go-multiaddr"
+	madns "github.com/multiformats/go-multiaddr-dns"
 	manet "github.com/multiformats/go-multiaddr-net"
 )
 
@@ -32,8 +34,14 @@ type Config struct {
 	Username string
 	Password string
 
-	// The ipfs-cluster REST API endpoint
+	// The ipfs-cluster REST API endpoint in multiaddress form
+	// (takes precedence over host:port)
 	APIAddr ma.Multiaddr
+
+	// REST API endpoint host and port. Only valid without
+	// APIAddr
+	Host string
+	Port string
 
 	// Define timeout for network operations
 	Timeout time.Duration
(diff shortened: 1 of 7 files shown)


api/rest/client/client.go
```diff
@@ -70,13 +79,32 @@ func NewClient(cfg *Config) (*Client, error) {
...
+
+	var host string
+	// APIAddr takes preference. If it exists, it's resolved and dial args
+	// extracted. Otherwise, host port is used.
+	if cfg.APIAddr != nil {
+		// Resolve multiaddress just in case and extract host:port
+		resolveCtx, cancel := context.WithTimeout(ctx, cfg.Timeout)
+		defer cancel()
+		resolved, err := madns.Resolve(resolveCtx, cfg.APIAddr)
+		cfg.APIAddr = resolved[0]
```
ZenGround0: Just double checking that the first element of resolved is what we want here.  Looking at madns I found that the first element in the case of an ipv4 address is the ip4maddr from: `ip4maddr, err := ma.NewMultiaddr("/ip4/" + ip4.String())`.  I had trouble tracking down what `ip4` exactly is from the code but was suspicious this might just be an ip and not include a transport.  Maybe that is fine too?  Just bringing up because that wouldn't match with the default multiaddr (ip + tcp addrs).

api/rest/client/client_test.go
```diff
@@ -63,249 +61,54 @@ func TestMultiaddressTakesPreference(t *testing.T) {
...
+	c, err := NewClient(cfg)
 	if err != nil {
 		t.Fatal(err)
 	}
 	if c.urlPrefix != "http://127.0.0.1:1234" {
 		t.Error("bad resolved address")
 	}
 }
-
-func TestVersion(t *testing.T) {
```
ZenGround0: Why did these tests make codeclimate unhappy?  Are we planning to replace them?

hsanjuan:
Code climate complained about function length, complexity, fake duplicates... I'm adjusting the settings, somethings are better fixed, others are not so realistic.

api/rest/client/client.go
hsanjuan: I guess you saw:

```
		ip4maddr, err := ma.NewMultiaddr("/ip4/" + ip4.String())
		if err != nil {
			return result, err
		}
               parts := append([]ma.Multiaddr{ip4maddr}, encap...)
               result = append(result, ma.Join(parts...))
```

See it does ma.Join on the result, so the resulting multiaddress has all the parts after the "/ip4/ip", including the /tcp/port etc..

api/rest/client/client_test.go
hsanjuan: oh, i forgot to add the file, i moved them

hsanjuan:
@ZenGround0 added the missing file

api/rest/client/client.go
ZenGround0: Ah I was thrown off thinking result result already has something else in it, thanks this makes sense

PR #268, opened 2017-12-11
ipfs-cluster dag/file sharding RFC WIP

ZenGround0:
@hsanjuan, this is the beginning of the rough draft RFC I have been working on.  Consider it under heavy development this week.  The unfilled sections are coming soon, plus there is more I would like to add to the RAID 0 section.  

I decided to go with a document + PR review process rather than an issue so that I could add and version and receive comments with better UX.  I'm not necessarily trying to add this document to the repo.

Forgive the somewhat scattered nature of a few paragraphs, particularly the TODOs, I am at a point where I have a lot of questions and ideas and don't want to lose track of them.  Let me know if you don't know what I mean by something and I'll add clarity.

By the end of this I'm hoping for a reasonably clear path forward to implement these new features to cluster, so I will be grateful if you point out places where I am assuming something incorrect,  optimizing for unmotivated use cases, overlooking simpler solutions, ignoring dangerous inefficiencies or doing something else that will get in the way of that.

### DIFF
--- a/docs/dag-sharding-rfc.md
+++ b/docs/dag-sharding-rfc.md
@@ -1,0 +1,182 @@
+# Introduction
+
+This document is a Request For Comments outlining a proposal for supporting sharded DAGs in ipfs-cluster, it outlines motivations for sharding DAGs across the nodes of an ipfs-cluster, and proposes a development path to achieve the motivated goals.
+
+# Motivation
+
+There are two primary motivations for adding data sharding to ipfs-cluster.  See the WIP use cases documents in PR #215, issue #212, and posts on the discuss.ipfs forum (ex: https://discuss.ipfs.io/t/ubuntu-archive-on-top-of-ipfs/1579 and https://discuss.ipfs.io/t/segmentation-file-in-ipfs-cluster/1465/2 and https://discuss.ipfs.io/t/when-file-upload-to-node-that-can-ipfs-segmentation-file-to-many-nodes/1451) for some more context.
+
+## Motivation 1: store files too big for a single node
+
+This is one of the most requested features for ipfs-cluster.  ipfs-pack (https://github.com/ipfs/ipfs-pack) exists to fill a similar need by allowing data stored in a POSIX filesystem to be referenced from an ipfs-node's datastore without acutally copying all of the data into the node's repo.  However certain use cases require functionality beyond ipfs-pack. For example, what if a cluster of nodes needs to download a large file bigger than any of its nodes' individual machines' disk size?  In this case using ipfs-pack is not enough, such massive files would not fit on a single node's storage device and coordination between nodes to store the file becomes essential.
+
+Storing large files is a feature that would serve many use cases and the more general ability to store large ipld DAGs of arbitrary structure has the potential to support others.  As an example consider splitting a massive cryptocurrency ledger or git tree among the nodes of an ipfs cluster.
+
+## Motivation 2: store files with different replication strategies for fault tolerance
+
+This feature is also frequently requested, including in the original thread leading to the creation of ipfs-cluster (See issue #1), and in @raptortech-js 's thorough discussion in issue #9.  The idea is to use sharding to incorporate space-efficient replication techniques to provide better fault tolerance of storage.  This would open up ipfs-cluster to new usage patterns, especially storing archives with infrequent updates and long storage periods.
+
+Again, while files storage will benefit from efficient, fault tolerant encodings, these properties are potentially also quite useful for storing arbitrary merkle DAGs. 
+
+# Background
+
+To add a file ipfs must "import" it.  ipfs chunks the file's raw bytes and builds a merkle DAG out of these chunks to organize the data for storage in the ipfs repo. The format of the merkle DAG nodes, the way the chunks are produced and the layout of the DAG depend on configurable strategies. Usually ipfs represents the data added as a tree which mimics a unix-style hierarchy so that the data can be directly translated into a unix-like filesystem representation.
+
+Regardless of the chunking or layout strategy, importing a file can be viewed abstractly as a process that takes in a stream of data and outputs a stream of DAG nodes, or more precisely "blocks", which is the term for the representation of DAG nodes stored on disk.  As blocks represent DAG nodes they contain links to other blocks.  Together these blocks and their links determine the structure of the DAG built by the importing process.  Furthermore these blocks are content-addressed by ipfs, which resolves blocks by these addresses upon user request.  Although the current imorting process adds all blocks corresponding to a file to a single ipfs node, this is not important for preserving the DAG structure.  ipfs nodes advertise the address of every block added to their storage repo.  If a DAG's blocks exist across multiple ipfs peers the individual blocks can readily be discovered by any peer and the information the blocks carry can be put back together.  This location flexibility makes partitioning a file's data at the block level an attractive mechanism for implementing sharding in ipfs-cluster.
+
+Therefore we propose that sharding in ipfs-cluster amounts to allocating a stream of blocks among different ipfs nodes.  Cluster should aim to do so:
+1. Efficiently: best utilizing the resources available in a cluster as provided by the underlying ipfs nodes
+2. Lightly: the cluster state should not carry more information than relevant to cluster, e.g. no more than is relevant for coordinating allocation of collections of blocks
+3. Cleverly: the allocation of blocks might benefit from the information provided by the DAG layout or ipld-format of the DAG nodes.  For example cluster should eventually support grouping together logical subgraphs of the DAG.
+
+On a first approach we aim for a simple layout and format-agnostic strategy which simply groups blocks output from the importing process into a size-based unit, from here on a "shard", that fits within a single ipfs node's storage repo.
+
+## Implementation
+### Overview
+We propose that cluster organizes sharding by pinning a management "Cluster DAG" defines links to groups of sharded data.  We also propose a sharding pipeline for ingesting data, constructing a DAG, and sharding across cluster nodes.  To allow for ingestion of datasets too large for any one ipfs node's repo we propose a pipeline that acts on a stream of data, forwarding blocks across the cluster while reading from the input stream to avoid overloading local resources.  First the data passes through the HTTP api of the adding cluster peer.  Here it is dispatched through an importing process to transform raw data into the blocks of a dag.  Finally these blocks are used to create the Cluster DAG and the data of both graphs is distributed among the nodes of the cluster.  Note that our goal of supporting large files implies that we cannot simply hand off data to the local ipfs daemon for importing, in general the data will not fit in the repo and AFAIK streaming blocks out of an ipfs daemon endpoint during importing is not something ipfs readily supports. The Cluster DAG's structure and the ingestiong pipeline steps are explained in further detail in the following sections.
+
+### Cluster DAG
+We propose to pin two dags in ipfs-cluster for every sharded file.  The first is the ipfs unixfs dag encoded in the blocks output by the importing process.  Additionally ipfs-cluster builds the Cluster DAG specifically built to track and pin the blocks of the file dag in a way that groups data into shards.  Each shard is pinned by (at least) one cluster peer.  The graph of shards is organized in a 3 level tree.  The first level is the cluster-dag root, the second represents the shards into which data is divided with at least one shard per cluster peer.  The third level lists all the cids of blocks of the file dag included in a shard:
+```
+The root would be like:
+{
+  "shards" : [
+     {"/": <shard1>},
+     {"/": <shard2>},
+     ...
+   ]
+}
+
+where each shard looks like:
+
+{
+  "blocks" : [
+     {"/": <block1>},
+     {"/": <block2>},
+     ...
+   ]
+}
+
+```
+Each cluster node recursively pins a shard-node of the cluster-dag, ensuring that the blocks of file data referenced underneath the shard are pinned by that node.  The index of shards, i.e. the root, can be (non-recursively) pinned on any one/all cluster nodes.
+
+With this implementation the cluster can provide the entire original ipfs file dag on request.  For example if an ipfs peer queries for the entire file they first resolve the root node of the dag which must be pinned on some shard.  If the root has child links then ipfs peers in the cluster are guaranteed to resolve them with content, as all blocks are pinned somewhere in the cluster as children of a shard-node in the cluster-dag.  Note that this implementation conforms well to the goal of coordinating "lightly" above because the cluster state need only keep track of 1 cid per shard.  This prevents the state size from growing too large or more complex than necessary for the purpose of allocation.
+
+The state format would change slightly to account for linking together the root cid of an ipfs file dag and the cluster-dag pinning its leaves
+
+```
+"<cid>": {
+  "name": <pinname>,
+  "cid": <cid>,
+  "clusterdag": <cidcluster>,
+  "allocations" : []
+}
+"<shard1-cid>": {
+  "name": "<pinname>-shard1",
+  "cid": <shard1-cid>,
+  "clusterdag": nil,
+  "allocations": [cluster peers]
+}
+....
+```
+
+Under this implementation replication of shards proceeds like replication of an ordinary pinned cid.  As a consequence shards can be pinned with a replication factor.
+
+### User APIs
+
+#### Importing files for sharding
+Data sharding must begin with a user facing endpoint.  Ideally cluster could stream files, or other data to be imported, across its HTTP API.  This way `ipfs-cluster-ctl` could launch the command for adding and sharding data in the usual way, and users also have the option of hitting the HTTP endpoint over the network.  Multiple files are typically packaged in an HTTP request as the parts of a `Content-Type: multi-part` request and HTTP achieves streaming with `Transfer-Type: chunked`.  Because golang http and mime/multipart [support chunked (a.k.a. streamed) transfer of multipart messages](https://gist.github.com/ZenGround0/49e4a1aa126736f966a1dfdcb84abdae), building the user endpoint into the HTTP API should be relatively straightforward.
+
+The golang multipart-reader exposes a reader to each part of the body in succession.  When the part's body is read completely, a reader to the next part's body can be generated by the multipart-reader.  The part meta-data (e.g. this part is a directory, that part is a file) and the data read from the part, must then be passed to the importer stage of the pipeline.  Our proposal is to call an external library, with a stream of data and metadata as input, from within the HTTP API to move to this next stage.
+
+As this endpoint will be used for streaming large amounts of data it has the potential to be somewhat difficult to use as system limits could be pushed and errors could occur after annoyingly long time investments.  Some questions that may come up as usability becomes a bigger concern:
+
+* Is it possible that remote users pushing data to the endpoint could overload cluster's resources by pushing at a rate faster than processing can occur in cluster? If this is an issue how can the endpoint maintain connections in a resource-aware non-degrading way?  One potential solution we are considering is to provide the HTTP API over libp2p, which is built for situations like this.
+* What features can cluster provide to make the process easier?  For example if cluster could report an estimate aggregate of all of the space in its nodes' repos then users could check that a huge file will fit on the cluster before importing it.
+* How should we handle the case where the sharding process fails part way through leaving behind pinned blocks in nodes across the cluster?  Could we have a built in mechanism for attempting cleanup after a failure is detected?  Could we provide a command for launching such a cleanup?
+* After implementing an HTTP API endpoint should we build out other methods of ingesting data such as pulling from a filesystem device or location address over the network?
+
+#### Sharding existing DAGs
+An endpoint should also exist for sharding an existing DAG among an ipfs-cluster.  A user will run something like `ipfs-cluster-ctl pin add --shard <cid>` and pin shards of the DAG rooted at `cid` across different nodes.  The importing step of the pipeline can be skipped in this case.  The naive implementation is to interact with the local daemon's DAG API to retrieve a stream of blocks as input to the sharding component.  While this implementation would still be quite useful for distributing DAGs shards from a node with a large amount of repo storage across a cluster, we currently suspect that this is not going to work well for large DAGs that cannot fit in the node's repo.  The "Future Work" subsection: "sharding large existing DAGs" discusses some possibilities for scaling this to large DAGs on nodes of arbitrary repo size.
+
+### Importer module
+As they are currently implemented in ipfs, importers run in three main phases.  The importer first begins by traversing a collection of data and metadata that can be interpreted as files and directories.  For example the tar importer reads data from a tar archive and the file importer used by `ipfs add` iterates over a collection of file objects.  Files are then broken into chunks with a chunker.  These chunks are then used as the data to make the dag-nodes constructed by the dag layout building process.  The dag-nodes are serialized as blocks and saved to the repo.
+
+Importing is currently a stream friendly process.  The objects traversed in the first stage are typically all wrappers around a reader of some kind and so the data and metadata (AFAIK) can be streamed in.  Existing chunkers read from their reader as needed when `NextBytes()` is called to get the next chunk.  The layout process creates in-memory DAG nodes derived from the leaf nodes, themselves thin wrappers around chunk data.  Nodes are flushed to the dagservice's `Batch()` process as soon as their parent links are established, and care is taken to avoid leaving references in memory so the golang garbage collector can reclaim the memory.  The balanced layout builder makes a balanced tree and so the memory high water mark is logN for a DAG of N nodes (I will have to investigate trickle's potential for streaming but my guess right now is it's good too).  Furthermore the dagservice batching collects a group of nodes only up to a threshold before forwarding them to their destination in the dagservice's blockstore.  Our least intrusive option is to create a "streaming blockstore" for the dagservice used in the importing process's dagbuilder helper for forwarding the imported blocks to the final stage of the pipeline.
+
+As it stands it is not clear that the existing unix files version of the first stage (go-ipfs/core/coreunix/add.go) is particularly streaming friendly as at a glance its traversal structure implies that all of the files of a directory exist while processing the directory.  Additionally there are a fair amount of non-importer specific ipfs dependencies used in the current implementation (most strikingly mfs) and we should evaluate how much of this we can live without.  We are currently expecting to move this stage of the pipeline to an importers github repo, allowing the modified importing process to be called as a library function from the HTTP API.  Perhaps this will be a step forward for the "dex" project to boot.  The tar format is a good example moving forward of how we might rewrite the files importer, as it's input is a logical stream of readers bearing many similarities to the multipart reader interface that will serve as input to the cluster importer.
+
+The format we use to encode file metadata in HTTP multipart requests can be mimicked from ipfs and should be standardized as much as possible for intuitive use of the API endpoint.
+
+Designing a streaming blockstore for outputting block nodes begs the question of how the blocks will be transferred to the cluster sharding component.  Multiple methods exist for streaming blocks out and we will need to investigate our desired assumptions and performance goals further to make a decision here.  Note that in addition to transferring data over rpc or directly over a libp2p stream muxer we have another option: pass data to the sharding cluster component by instantiating the sharding functionality as a component of the blockstore used in the importer.
+
+### Sharding cluster component
+
+At a high level the sharding component takes in a stream of dag-node blocks, maintains the necessary data to build a shard node of the Cluster DAG, and forwards the stream of blocks to a peer for pinning.  When blocks arrive the sharding process records the cids which are needed for constructing links between the shard node and the blocks of the DAG being sharded.  This is all the sharding component must do before forwarding blocks to other peers.  The peer receiving the shard is calculated once by querying the allocator.  The sharding component must be aware of how much space peers have for storing shards and will maintain state for determining when a shard is complete and the receiving peer needs to be recalculated.  The actual forwarding can be accomplished by periodically calling an RPC that runs DAG API `put` commands with processed blocks on the target peer.  After all the blocks of a shard are transmitted the relevant sub-graph of the Cluster DAG is constructed and sent as a final transmission to the allocated peer.  Upon receiving this final transmission the allocated peer recursively pins the node of the Cluster DAG that it receives, ensuring that all blocks of the shard stay in its ipfs repo.  When all shards are constructed and allocated the final root node pointing to all shard node is pinned (non-recursively) throughout the cluster.
+
+One remaining challenge to work out is preventing garbage collection in the receiving node from removing the dag blocks streamed and added between the first and final transmission.  We will either need to find a way to disable gc over the ipfs API (AFAIK this is not currently supported), or work with the potentially expensive workaround of pinning every block as it arrives, recursively pinning the shard node at the end of transmission, and then unpinning the individual blocks.
+
+# Future work
+
+## Fault tolerance
+Disclaimer: this needs more work and that work will go into its own RFC. This section provides a basis upon which we can build.  It is included to demonstrate that the current sharding model works well for implementing this important extension.  We will bring this effort back into focus once the prerequisite basic sharding discussed above is implemented.
+
+### Background reading
+This [report on RS coding for fault tolerance in RAID-like Systems by Plank](https://web.eecs.utk.edu/~plank/plank/papers/CS-96-332.pdf) is a very straightforward and practical guide.  It has been helpful in planning out how to approach fault tolerance in cluster and will be very helpful when we actually implement.  It has an excellent description of how to implement the algorithm that calculates the code from data, and how to implement recovery.  Furthermore one of his example use case studies include algorithms for initialization of the checksum values that will be useful when replicating very large files.
+
+### Proposed implementation
+#### Overview
+The general idea of fault tolerant storage is to store your data in such a way that if several machines go down all of your data can be reconstructed.  Of course you could simply store all of your data on every machine in your cluster, but many clever approaches use data sharding and techniques like erasure codes to achieve the same result with fewer bits stored.  A standard approach is to store shards of data across multiple data devices and then store some kind of checksum information in separate, checksum devices.  It is a simple matter to extend the basic sharding implementation above to work well in this paradigm.  When storing a file in a fault tolerant configuration ipfs-cluster, as in basic sharding, will store the ipfs files DAG without its leaves and an cluster-DAG.  However now the cluster-DAG has additional shards not referencing the leaves of the ipfs files DAG, but rather to checksum data taken over all the file's data.  For an m out of n encoding:
+
+```
+The root would be like:
+{
+  "shards" : [
+     {"/": <shard1>},
+     {"/": <shard2>},
+     ...
+     {"/": <shardN>},
+   ]
+   "checksums" : [
+     {"/": <chksum1>},
+     {"/": <chksum2>},
+     ...
+     {"/": <chksumM>},
+}
+```
+
+While there are several RAID modes using different configurations of erasure codes and data to checksum device ratios, my opinion is that we probably can ignore most of these as using m,n RS coding is superior in terms of space efficiency for fault tolerance gained.  However different RAID modes have different time efficiency properties in their original setting anyway.  It is unclear if implementing something (maybe) more time efficient but less space efficient and fault tolerant than RS has much value in ipfs-cluster.  I lean towards no but I should investigate further. (TODO -- answer these questions, any counter example use cases?  any big gains for using other RAID modes that help these use cases?) On another note in Feb 2019 the tornado code patent is set to expire (\o/) and we could check back in then and look into the feasibility of using (perhaps implementing if no OSS exists yet?!?) tornado codes (which are faster).  There are others we'll want to check the legal/implementation situation for (biff codes) so pluggability is important.
+
+Overall this is pretty cool for users because the original DAG (recall how basic sharding works) and the original data exist within the cluster.  This way users can query any cid from the original DAG and the cluster ipfs nodes will seamlessly provide it, all while silently and with very efficient overhead they are protecting the data from a potentially large number of peer faults.
+
+We have some options for allowing users to specify this mode.  It could be a cluster specific flag to the "add" endpoint or a config option setting a cluster wide default.
+
+#### Importing with checksums
+If memory/repo size constraints are not a limiting factor it should be straightforward for the cluster-DAG importer running on the adding node to keep a running tally of the checksum values and then allocate them to nodes after getting every data shard pinned.  Note this claim is specific to RS as the coding calculations are simple linear combinations of operations and everything commutes, while I wouldn't be surprised if potential future codes also had this property it is something we'll need to check up on once we get serious about pluggability.
+
+If we are in a situation where shards approach the size of ipfs repo or working memory then we can gather inspiration from the report by Plank, specifically the section "Implementation and Performance Details: Checkpointing Systems".  In this section Plank outlines two algorithms for setting checksum values after the data shards are already stored by sending messages over the network.  From my first read-through the broadcast algorithm looks the most promising.  This algorithm would allow cluster to send shards one at a time to the peer holding a zeroed out checksum shard and then perform successive updates to calculate the checksums, rather than requiring that the cluster-DAG importer hold the one shard being filled up for pinning alongside the m checksum shards being calculated.
+
+## Clever sharding
+Future work on sharding should aim to keep parts of the DAG that are typically fetched together within the same shard.  This feature should be able to take advantage of particular chunking and layout algorithms, for example grouping together subDAGs representing packages when importing a file as a linux package archive.  It would also be nice to have some techniques, possibly pluggable, available for intelligently grouping blocks of an arbitrary DAG based on the DAG structure.
+
+## Sharding large existing DAGs
+AFAIK this would require new features in go-ipfs.  Two approaches are apparent.
+
+If ipfs-cluster is to stream DAG nodes from its local daemon to create a Cluster DAG, the ipfs api would need to include an endpoint that provides a stream of a DAG's blocks without over-committing the daemon's resources (i.e. not downloading the whole DAG first if the repo is too small).  One relevant effort is the [CAR (Certified ARchives) project](https://github.com/ipld/specs/pull/51) which aims to define a format for storing DAGs.  One of its goals as stated in a recent proposal is to explicitly allow users to "Traverse a large DAG streamed over HTTP without downloading the entire thing."  Cluster could shard DAG nodes extracted from a streamed CAR file, either kept on disk or provided at a location on the network.  CAR integration with go-ipfs could potentially allow resource aware streaming of DAG nodes over the ipfs daemon's api so that the sharding cluster peer need only know the DAG's root hash.
+
+Another relevant effort is that of ipld selectors.  ipld selectors are interesting because they are an orthogonal approach to the Cluster DAG for succinctly organizing large pinsets.  We did not end up including them in the main proposal because adding large datasets necessitates importing data to a DAG anyway.  In this case, from our perspective, building a management DAG for pinning and directly transferring blocks is simpler than adding a portion of the DAG to the local ipfs node, executing a remote selector pin, blocking on ipfs to do the transfer, and then clearing resources from the local node.  However in the case that the dag is already on ipfs, pinning selectors for shards is a potential implementation strategy.  The selectors would essentially act as the shard-nodes of the Cluster DAG, selecting shards of the DAG that could fit on individual nodes and getting pinned by the assigned nodes.  At the very least this seems to require that ipld selectors can select based on sub-tree size.  It is currently unclear how well matched ipld selectors would be for this use case.
+
+# Work plan
+
+There is a fair amount that needs to be built, updated and serviced to make all of this work.  Here I am focusing on basic sharding, allowing support for files that wouldn't fit in an ipfs repo.  Support for fault tolerance, more clever sharding strategies and sharding large existing DAGs will come after this is nailed down.
+
+* Build up a version of the `importers` (aka `dagger` aka `dex`) project that suits cluster's needs (if possible keep it general enough for general use).  In short we want a library that imports data to ipfs DAGs from arbitrary formats (tar, deb, unix files, etc) with arbitrary chunking and layout algorithms.  For our use case it is important that the interfaces support taking in a stream of data and returning a stream of blocks or DAG nodes.
+* Implement a refactored importers library (e.g. proto `dex`) to be imported and called by the HTTP API.  It will need to provide an entry call that supports streaming and an output mechanism that can connect to the sharding component.
+* Create a cluster add endpoint on the REST API that reads in file data on HTTP multipart requests and calls the cluster importer library entrypoint function.  Listen on an output stream and put the output blocks into the local ipfs daemon and pin the root with cluster.  This will require nailing down the APIs of these components and development of a garbage collection management strategy.
+* Build out the sharding component, implementing creation of the Cluster DAG and pin blocks by way of pinning shards.  This is still all confined to one cluster peer.  This stage may require defining a cluster ipld-format (if we go that route).
+* Update the state format so that pinned cids reference Cluster DAGs and so that the state differentiates between cids and shards.
+* Build support for allocation of different Cluster DAG shard-nodes to different peers.  This includes implementing RPC transmission of all chunks of a given shard to the allocated peer by the sharding component.
+* Work to make the state format scale to large numbers of cids
+* Add in usability features: don't make sharding default on add endpoint but trigger with --shard, maybe make configs that can set sharding as default or importer defaults, allow add to specify different chunking and importing algorithms like ipfs, ipfs add proxy endpoint can also take in --shard and then calls same functionality as add endpoint, add in tools useful for managing big downloads like reporting the total storage capacity of cluster or the expected download time.
+* Testing should happen throughout and we should have a plan with regards to which tests we run in place early (maybe before we start).  Eventually we will want a huge cluster in the cloud with a few TB of storage for ingesting huge files
+* Move on beyond basic sharding and start design process again to support clever techniques for certain DAGs/usecases, sharding large existing DAGs, and fault tolerance


hsanjuan:
I leave this here as relevant: https://github.com/ipfs/notes/issues/144

hsanjuan:
And this https://github.com/ipfs/go-ipfs/blob/master/importer/chunk/splitting.go#L14

hsanjuan:
Hey, I am giving some thought to this. I've come to realize some things:

### Something stored on IPFS has two very distinct parts

  * The DAG/Merkle Tree - represents the layout of the data
  * The data nodes (leaves) - plain data

Therefore, cluster should figure out how to divide those two things among peers. These are separate problems. For simplicity, we probably want to assume all DAG/Merkle trees fit in a single IPFS node for the moment, and **worry only about the data nodes**.

### IPFS is very efficient when tracking DAGs

We should aim to minimize cluster-state size and track as little elements as possible, let IPFS nodes deal with as much of the DAG-data as we can.

### Pinning the DAG

The DAG should be relatively small compared to the data it references in most cases (at least the ones we want to tackle). If we assume it fits in a single ipfs node, storing the dag would be easy with something like `ipfs pin <dagroot> --no-leaves` and tracking only the dagroot. We would have to see if we implement such thing in ipfs or cluster, and how.

### Pinning the data

When data is ingested, it's turned to blocks which can be referenced from the DAGs. Since we don't worry about the DAG, we just worry about potentially distributing a large list of blockhashes. We do not worry at all about the layout of those blocks (i.e. a blocks from a single file could end up in several cluster peers), yet.

Cluster should not track all this large list of CIDs when IPFS is perfectly suited to it. Instead, we need a `cluster-dag-node` (which maybe is a simple unixfs), or `cluster-ipld-node`. This dag is independent from the pinned-object-dag and represents how the data is splitted in cluster. I'm thinking of two levels: 

  * first level represents the cluster-parts in which the data is divided and can be allocated to a cluster peer. There are not too many entries in this first level. These parts are supposed to fully fit in an ipfs-node and are tracked by cluster.
  * second level lists all the ipfs data leaves contained in the first-level group. Cluster does not track each of these leaves, that is just data stored in ipfs.

This way the cluster dag is stored on ipfs and cluster tracks the root ID and first-level of that cluster-DAG only.

Therefore, let's imagine a cluster-dag:

```
The first level would be like: 
{
  "parts" : [
     {"/": <part1>},
     {"/": <part2>},
     ...
   ]
}

where each part looks like (this might well be a unixfs node with multiple sublevels because directory sharding is needed):

{
  "blocks" : [
     {"/": <block1>},
     {"/": <block2>},
     ...
   ]
}

```

Then, the cluster state entry for a pin and it's parts could look like:

```
"<cid>": {
  "name": <pinname>,
  "cid": <cid>,
  "cluserdag": <cidcluster>,
  "allocations" : []
}
"<part1>": {
  "name": "<pinname>-part1",
  "cid": <part1>,
  "clusterdag": nil,
  "allocations": [cluster peers]
}
....
```

When adding data to cluster:

* It would use the ipfs importer of choice to split it and generate the dag..
* As it goes it would add dag nodes to IPFS.
* It would use the cluster splitter of choice to group data blocks and reference them with a cluster-dag- part (probably based on size on the first approach)
* Once enough blocks can form a part, it would find a cluster peer to allocate them, block/put them there, and pin the cluster-dag-part in cluster normally (which will trigger replication strategies).
* Once all parts are processed, a cluster-dag is created referencing the different parts. This is pinned in cluster.
* Once the cluster dag is ready, it would pin the ipfs-dag object (without the leaves) normally in cluster, including the reference to the resulting cluster-dag.

### Open questions

* What's the right format to build a cluster DAG. unixfs? custom-ipld? Both (depending on level)? It should scale.
* Cluster should be able to receive a stream of data and, as it comes:
  * Decide when it's enough to make a part that can be allocated to a single peer
  * Create it and allocate it
  * Move on and keep splitting and building the ipfs dag and the cluster dag
  * We have to be very clear that data needs to be placed in it's final destination as it arrives, and we cannot wait for a full upload and do it somehow in steps.
* No matter how we put it, it is possible to foresee that the cluster shared state will grow heavily and will be hard to store and work with it in-memory as we do now. We probably need to start thinking of incorporating leveldb, boltdb or something else to track things (re-implement the state component so it scales).
* We need an easy way to ask ipfs to pin only DAG-nodes and not data nodes.
* The approach above is dummy. We may want to make more intelligent "parts", given the context provided by the DAG (or parts of it). This opens a few questions on how this would work, and I mostly see that the result is many more parts to track than dummy-size-based-splitting: The solution to store intelligently an apt repository might be to make 1-part-per-package. This means cluster will end up tracking many parts in its state.
* I have no idea how big the state map can grow right now. I think we are mostly limited by RAM. We should figure this out to get a feel of how many parts a cluster could handle.
* I have realized RAID0 is a bad goal because it adds some complexities that we should not have to worry about. If a file is split in parts, and the number of parts corresponds to the number of peers, and the replication factor is 1, and we have an allocator which somehow can allocate parts to cluster peers where no other parts of that file are stored, then you have raid0. Since we don't have such allocator, two or more of the parts could end up in the same cluster peer because that's the one that has space. So I don't want to support RAID0 (it's not even a good way to keep data safe). I want to support things with replication_factor>1 because then it becomes easy to evenly distribute the different replicas among peers (just how it works now) and you can actually take some failures. In short, for a start: **it should not matter how data parts are distributed, as long as they reach a certain replication factor among cluster peers**. Probably one of the other RAID* levels fits better here. Perhaps we should not talk about any standard for the moment, just make sure we can do our thing for a start and that it works.
* One day we might have to worry about DAGs too large to fit in a single ipfs node. If our system treat them as plain data, could it distribute them among cluster peers? Would that even work or be useful in such case?
* What new modules would we need in our code to approach an implementation like this? How easy is to re-use ipfs importers?
* When receiving content, and a part is ready, it needs to be allocated. Before pinning, content needs to be added to the allocated peers (at least to one). The allocated peer is probably different than the one from which the content is being added so, how do we forward this data? Is our RPC suited for such large chunks?
* What about the role of IPLD? do we forget about the `Pin(<ipld_path>)` support for the moment? Does it make sense to allow the user provide a cluster-dag directly? As a way to ingest data to a cluster (assuming the data is already in ipfs somewhere else in the world)?
* Should we prototype a tool that does importing?:
  * Receives a stream of data
  * Outputs ipfs dag nodes
  * outputs cluster-dag nodes
  * for each cluster-dag-part outputs data blocks
* Is there a better approach than a cluster-dag described above?

ZenGround0:
> Therefore, cluster should figure out how to divide those two things among peers. These are separate problems. For simplicity, we probably want to assume all DAG/Merkle trees fit in a single IPFS node for the moment, and worry only about the data nodes.

I enthusiastically agree with this assumption for now.   When we start thinking about storing general merkle dags we may need to revisit the assumption that leaf nodes are the only nodes with significant size.

> We should aim to minimize cluster-state size and track as little elements as possible, let IPFS nodes deal with as much of the DAG-data as we can.

I appreciated you pointing this out yesterday, I am onboard with this.

> storing the dag would be easy with something like ipfs pin <dagroot> --no-leaves and tracking only the dagroot. We would have to see if we implement such thing in ipfs or cluster, and how.

Big +1.  I am planning to dive into ipfs pinning sometime soon and will keep an eye out for a place this functionality could be added if it doesn't exist yet.  It seems generally useful as part of go-ipfs.

> Then, the cluster state entry for a pin and it's parts could look like:

I'm not exactly sure what the proposed clusterdag field is in the state, but my impression is that we only need cids, one for the root non-recursively pinned and one for the part recursively pinned.  I guess I am suggesting a recursive-pinned field in the state which might be expressively equivalent to the clusterdag field?



>When adding data to cluster:
>* It would use the ipfs importer of choice to split it and generate the dag..
>* As it goes it would add dag nodes to IPFS.

+1 

>* It would use the cluster splitter of choice to group data blocks and reference them with a cluster-dag- part (probably based on size on the first approach)

One thing that is not 100% clear from this bullet is whether the data is sharded first by cluster and then these blocks are imported into ipfs (this is how I am reading it), or if the reference cluster splitter is operating on something more than raw data (the phrase "group data blocks" makes me think maybe the cluster splitter is aware of ipfs block boundaries).  

> * Once enough blocks can form a part, it would find a cluster peer to allocate them, block/put them there, and pin the cluster-dag-part in cluster normally (which will trigger replication strategies).

Hmm I am a bit confused about some of the things being proposed here.  Why not just recursively pin the root of the cluster-dag-part in the storing peer? The dag should reference all of the ipfs imported dag down to the leaves. Is the block/put there for performance reasons?  Also is all of this importing going on in one ipfs node?  If that's the case then it seems like the dag generation can happen in one step instead of two, we would just use a custom layout (instead of `ipfs add --trickle`,  `ipfs add --cluster-layout <n>` where n is the number of parts we want.  I'm now seeing that maybe this is what you meant by "cluster splitter" whereas I originally thought of this as the splitter turning a datastream into chunks to send to different cluster peers.

> * Once all parts are processed, a cluster-dag is created referencing the different parts. This is pinned in cluster.

+1, constructing some finalizing index is unavoidable.  Collecting together hashes under one root node should be functionality in the `dagger`/`dex` tool. 

> * Once the cluster dag is ready, it would pin the ipfs-dag object (without the leaves) normally in cluster, including the reference to the resulting cluster-dag.

+1

Onto open questions:

> * What's the right format to build a cluster DAG. unixfs? custom-ipld? Both (depending on level)? It should scale.

This is a good question.  In the case of simply adding files, which is where we are starting, I think the constraint is going to be that many of the use cases requiring sharding are dealing with a "too big for a single repo" directory, and in that case they will want to be able to use paths through the dag with directory semantics, e.g. `<root-hash>/apt-main-dir/cool-packages/subfolder/wonderbar-package`, and would rather not have to keep track of which part hash certain packages are on.  I'd like to look through unixfs sharded directories (e.g. how wikipedia mirror worked) a little more closely and see if there is someway to make this addressing possible with cluster.  We could potentially do some address translation on query also (related discussion of cluster interface below).

I think this question is even more interesting when we think about arbitrary formatted dags.  If we shard arbitrary dags across a cluster do we want the cluster dag to be of the same ipld-format/splitting alg/dag layout as the original dag?  Do we want cluster dags to be free to group unrelated dags (e.g. allow cluster dags with one part being a unixfs dag and another part being a git dag)?  

FYI this discussion is also relevant when we move from basic sharding to different replication formats.  For example, should cluster's representation of a RAID-4 replicated arbitrary dag include the RAID-4 parity metadata as a part of the format original format (e.g. an additional file in unixfs), or should the parity metadata be treated as part of a more abstract piece of a *meta-format*.

>* Cluster should be able to receive a stream of data and, as it comes:
>  * Decide when it's enough to make a part that can be allocated to a single peer

+ 1 for mentioning this.  Knowing the total size makes this easy.  If we don't know the total size we could start making guesses (one third the current repo's size), and if we underestimate later combine blocks that are sent to the same repo.  Gets a little trickier at first glance with lots of variance in repo size.

Another thing on my mind, we should probably try to split along boundaries that the ipfs chunker uses on import so we get the same leaf hashes (depending on usecases/ get workflow/ cluster interface we may want this, more discussion below).  This should be simple for the size splitter (multiples of chunk size) maybe more involved for the rabin splitter.  Functionality probably belongs in `dagger`/`dex`

>  * We have to be very clear that data needs to be placed in it's final destination as it arrives, and we cannot wait for a full upload and do it somehow in steps.

One thing I am leaning towards that I am unsure you have the same opinion about: the raw data should be sent out over libp2p (rpc?) and then imported locally on each cluster node instead of being streamed to one node, imported and then transferred to another node through ipfs via a pin.


> * No matter how we put it, it is possible to foresee that the cluster shared state will grow heavily and will be hard to store and work with it in-memory as we do now. We probably need to start thinking of incorporating leveldb, boltdb or something else to track things (re-implement the state component so it scales).

A quick note I haven't put much thought into, but maybe we could eventually store state, or parts of state as ipfs hashes that could be resolved lazily to keep the working memory needed for state manipulations down?

> * We need an easy way to ask ipfs to pin only DAG-nodes and not data nodes.

++1!

> * The approach above is dummy. We may want to make more intelligent "parts", given the context provided by the DAG (or parts of it). This opens a few questions on how this would work, and I mostly see that the result is many more parts to track than dummy-size-based-splitting: The solution to store intelligently an apt repository might be to make 1-part-per-package. This means cluster will end up tracking many parts in its state.

Specifically on the apt repo point, elopio mentioned a lot of directory structure in the archive.  A sensible apt-archive importer would generate a unixfs directory tree probably with a lot of top level directories.  I don't see why we couldn't keep parts down to one per cluster and just put a bunch of directories under a single part, or in other words add another layer to the cluster-dag.  The utility of parts as proposed comes from one fitting per cluster and minimizing tracked cids in cluster state.  Since cluster controls the creation of the cluster dag it seems reasonable to construct it to fit this invariant regardless of the underlying dag structure.  

I think the room for improvement on the proposed cluster dag structure is more general.  There are two issues I have thought of so far that can probably be ignored while we start with a simple prototype, but eventually will need to be addressed.

In general for using a custom dag layout for cluster sharding:
Lack of redundancy: now clusters of size 15 and size 16 pin different hashes for the same data.  Data has a different hash from what someone adding to a regular ipfs node would have.  — whether or not this is actually a problem depends a lot on use case, and the decisions we make about an ipfs-cluster’s interface (more discussion below).  Also I can already think of ways that might get around this so maybe it won’t be so bad even if it’s undesirable.

Specific to the proposed custom dag layout format:
It doesn’t seem to match well with the idea of dynamic clusters where peers come and go.  Say you start with a 10 peer cluster and lose 1 peer, now you have a balanced graph for 10 peers and try to allocate it across 9.  Maybe you could take part 10 and give it to peer 9, but what if peer 9 doesn’t have quite enough space, but together peers 7,8 and 9 have enough space?  Under the even parts assumption it sounds like you might want to rebuild the dag which gets complicated but might work.  You could also keep the dag the same and start pinning different subtrees of the graph (maybe using ipld selectors), but because the dag is 10 even subtrees, it doesn’t lend itself well to reorganizing into 9 subgraphs.  Supporting sharded dags in dynamic clusters is an interesting problem, you could imagine constructing graphs that can be rebalanced fairly and keep it so that only a small number of cids need to be pinned.  Admittedly this is probably a lot further down the road.



> * I have no idea how big the state map can grow right now. I think we are mostly limited by RAM. We should figure this out to get a feel of how many parts a cluster could handle.

It sounds like we have our first sharding inspired metric!  Perhaps we could sweep number of cids (for now using current state) over a few cluster sizes. Maybe we should sweep cluster sizes first to figure out a good range.

> * I have realized RAID0 is a bad goal because it adds some complexities that we should not have to worry about. If a file is split in parts, and the number of parts corresponds to the number of peers, and the replication factor is 1, and we have an allocator which somehow can allocate parts to cluster peers where no other parts of that file are stored, then you have raid0. Since we don't have such allocator, two or more of the parts could end up in the same cluster peer because that's the one that has space. So I don't want to support RAID0 (it's not even a good way to keep data safe). I want to support things with replication_factor>1 because then it becomes easy to evenly distribute the different replicas among peers (just how it works now) and you can actually take some failures. In short, for a start: **it should not matter how data parts are distributed, as long as they reach a certain replication factor among cluster peers**. Probably one of the other RAID* levels fits better here. Perhaps we should not talk about any standard for the moment, just make sure we can do our thing for a start and that it works.

Mostly agree here.  RAID 0 was already starting to feel a little forced, I mostly was trying to avoid reinventing the wheel/naming but yeah it is a bit mismatched from baseline cluster sharding so I agree doing our own thing for while makes more sense.  +1 for RAID 0 being pretty useless in cluster, it was never intended as more than a stepping stone.  However, I do think considering the implications of sharding design on implementing better replication strategies is worthwhile right now, as it is one of the two big user benefits that I see sharding leading up to.  From this view having support for RAID 0 in cluster gives us some benefits that perhaps you are overlooking.  For example the allocator you described sounds like it will be important down the road for getting the best fault tolerance out of a cluster, as eventually we *will* want replication factor == 1 in the interest of space efficiency as we can push improved fault tolerance to encodings. 

> * One day we might have to worry about DAGs too large to fit in a single ipfs node. If our system treat them as plain data, could it distribute them among cluster peers? Would that even work or be useful in such case?

I take it you mean DAGs as in the cluster metadata necessary to organize subgraphs of a larger file?  My take is that if we are importing a file into cluster we assume we have enough total storage for all of its raw data and the dag structure around it.  We could always push some of the cluster organizational dag to specific subdags assigned to single node in an effort to minimize the total shared metadata so that it can fit on every node.  For example take the apt-package archive structure from above.  Perhaps at the top level the archive has many directories so that the organizational dag necessary to subdivide the dag into n (number of cluster peers) parts is too big to fit on some nodes.  In that case some of the "cluster-dag" structure specific to some archives could be pinned to the ipfs node with those archives, so that now the part root contains both ipfs and cluster dag.  Now the organizational ipfs-cluster dag can be pinned by everyone.

>* What new modules would we need in our code to approach an implementation like this? How easy is to re-use ipfs importers?

I would be pretty into taking on the `dagger`/`dex` extraction project if there is broader consensus from PL that this should happen.  Personally I think it should be freed from go-ipfs!  Depending on how people feel about it I could look into making an external copy of these libraries for use by cluster with the eventual goal of total extraction so that we can more easily add our own layouts and splitters.  It might not be worth it if the go-ipfs project doesn't prioritize this because then I might find myself maintaining a fork.  I could potentially start a new repo with only our implementations that could also serve as a starting point for extraction, and then we could import go-ipfs importer/ to cluster (not sure how you feel about go-ipfs subpackages as dependencies?).

> * When receiving content, and a part is ready, it needs to be allocated. Before pinning, content needs to be added to the allocated peers (at least to one). The allocated peer is probably different than the one from which the content is being added so, how do we forward this data? Is our RPC suited for such large chunks?

+ 1 for questioning how suited RPC is, this sounds like potentially our second sharding inspired metric!  I think I have a slightly different idea than you, especially in the case of data that is too big to fit on a single node, in that I am leaning towards sending raw data to the allocating node, adding content to its ipfs node (ipfs-dag construction) and then pinning on the allocating node.  Assuming an even spread of allocations this gets us parallelization.  I haven't thought through this all the way so I appreciate you filling me in if I'm missing something.


>* What about the role of IPLD? do we forget about the `Pin(<ipld_path>)` support for the moment? Does it make sense to allow the user provide a cluster-dag directly? As a way to ingest data to a cluster (assuming the data is already in ipfs somewhere else in the world)?

++1!  For ipld selectors I think once ipfs's merkledag situation is figured out a little bit better (https://github.com/ipfs/notes/issues/255) this might fall right out of the pinning interface.  One thing we might eventually want is a record of the namespaces of cids in our state so that we can handle both "ipfs/" and "ipld/" when this starts being part of ipfs resolution.

 As for ingesting already-built dags I think that while right now this is a bit out of scope (go-ipfs doesn't have much support for non-ipfs files dags, AFAIK), this has the potential to serve really important cluster use cases.  Imagine for example ingesting immense git repos or blockchains to a cluster that could provide built-in fault tolerant storage.  

Depending on how we do data division there are some close parallels between the importing process and a process using ipld transforms.  Say there is an existing ipfs file-dag pinned at a node that you want to shard.  If you had room on the node you could fetch a transform of the graph that rearranges its dag to bunch groups of its file data into the leaves of n trees joined together by a cluster dag at the root.  It seems plausible such a transform exists for any dag.  I have been thinking along similar lines when it comes time to tackle more complex replication strategies: perhaps we can have a general transform to make a dag transform date to RAID-4 or ReedSolomon encoded replication in the same way we can have a transform for breaking a dag into n cluster parts.

Of course like support for non-ipfs dags ipfs doesn't have ipld transforms on the near horizon of its roadmap.  In the meantime I think reorganizing a dag for storage is pretty manual making ingestion of already-built-dags complex and time consuming, and another downside (in the short term anyway) of using a custom dag layout for ipfs-cluster dags.

>* Should we prototype a tool that does importing?:
>  * Receives a stream of data
>  * Outputs ipfs dag nodes
>  * outputs cluster-dag nodes
>  * for each cluster-dag-part outputs data blocks

I want to think about this more and pin down precisely what we want it to do and perhaps add more, but I think suggesting a prototype along these lines will now be part of the RFC.

* Is there a better approach than a cluster-dag described above?
I can imagine situations where we would not want to change the dag layout, for example if we wanted cluster to pin the same hash as a specific layout, or if we want to pin already added content.  If this path is properly motivated (I'm not convinced of that yet, particularly for our first implementation) we could investigate a solution based on ipld selectors that allows cluster peers to collaboratively pin a dag while bounding the total number of cids that each peer needs to track. 

 To make this more concrete consider a balanced layout ipfs dag with 4 children from the root being collaboratively pinned by 4 cluster peers.  In this case the allocation would be trivial: each peer would take one of these first level children and then non-recursively pin the root.  I can imagine, though haven't yet dug into this, that for a lot of dags things may not be this balanced, but they have a structure such that you can search for a small set of nodes in the dag that the peers can pin so that you get minimal overlap between peers (approximate sharding), approximately equal sharding among n peers, and relatively small number of cids for each peer to track. 

 Sounds like an interesting problem, and maybe we could encode its solution into an ipld selector (i.e. select all nodes that satisfy the property of being the solution to the above problem) to retrieve the nodes in a simple way.  However, as ipld selectors are a ways off and it is unclear if leaving the dag untouched is very valuable right now, I am in favor of starting with modifying the layout through the cluster-dag proposed.

Ok almost done, here's some final thoughts I wrote up some point during my first read through:

We’re talking a lot about getting data into cluster, a sensible place to start.  We should keep in mind that this is only a part of the cluster workflow for users, they will need to fetch or search over data stored in cluster at some point too.  In a sense these ideas fit into the current workflow (cluster daemon pins, ipfs daemon provides) because ipfs blocks can be fetched by their canonical hashes as the “ipfs-dag” maintains the expected structure.  But when we transform the upper layer of the dag to make it easy for our state to track data, we are of course modifying the total hash of all the data.  Will users want to be aware of cluster’s clever dag reorganizations?  Alternatively will we want to keep a mapping between the canonical data’s hashes/paths and the final cluster dag’s hashes/paths?  I realize a lot of this may stay unclear for a while until we start adapting to specific use cases, but if we can get a working idea of the interfaces we want to focus on this will help us build the right thing.

My central point of confusion is how committed we are to the idea of a “vnode” interface, i.e. interacting with an ipfs-cluster as a virtual ipfs node.  Do you imagine cluster supporting an opaque ipfs interface that hides details of its organization of data (i.e. the cluster-dag layout + different hash) alongside the interface of today that is more or less transparent?  Do you think this is something ipfs-cluster could benefit from/should focus on?

Alright thank you for reading!  Looking forward to any responses, I will get started incorporating these ideas into an updated draft of the RFC.

hsanjuan:
I think perhaps it wasn't clear: in my proposal, we are NOT touching the original ipfs DAG. We are producing that DAG and chunking data exactly how ipfs does it. The difference is that we have an additional cluster-dag, which in the end references the same leaves, but grouping them in subtrees.

Thus, when the user does `ipfs get <root-hash-pinned-in-cluster-of-something-really-big>/path/to/myfile`, any ipfs node will resolve that and fetch the necessary blocks. Those blocks may be in different cluster-dag-parts, in different ipfs daemons, but they can be fetched and put together.

If the user needs to re-create the cluster-dag (because a peer has come or gone), they should be able to do a Pin operation which recalculates the cluster-dag, and re-allocates content accordingly.

I think we need to do chunking and dag building ourselves as the data is ingested because we don't know how much data there is or if the local daemon can take it. My idea is that we chunk it in cluster like ipfs does. And as leaves are produced, we group them, and once grouped we find them an allocation and forward the chunk there, where it is injected to ipfs (with block/put because it's already chunked) and pinned. Maybe I'm wrong though.

> I'm not exactly sure what the proposed clusterdag field is in the state, but my impression is that we only need cids, one for the root non-recursively pinned and one for the part recursively pinned. I guess I am suggesting a recursive-pinned field in the state which might be expressively equivalent to the clusterdag field?

the clusterdag is a CID. The field links the original ipfs-dag CID to the Cluster DAG. It is how we know that for a given CID, cluster is pinning a bunch of  cluster-dag-subtrees rather than pinning directly the content like it works now.

hsanjuan:
I need to have a look to dagger/dax. Perhaps we can make a call afterwards and share our thoughts

ZenGround0:
>I think perhaps it wasn't clear: in my proposal, we are NOT touching the original ipfs DAG. We are producing that DAG and chunking data exactly how ipfs does it. The difference is that we have an additional cluster-dag, which in the end references the same leaves, but grouping them in subtrees.

Oooh thanks for the clarification.  I read through that thinking that the cluster dag referenced *per node ipfs dags*.  I'll have to go back through the proposal in light of this.

> Those blocks may be in different cluster-dag-parts, in different ipfs daemons, but they can be fetched and put together.

I should spend some time understanding ipfs name resolution better because I think that's where my confusion comes from.  I was under the impression that if you are referencing a file through a dag then you need to explicitly denote each dagnode in the path.  Because the cluster dag takes care of sharding data across parts it necessarily adds in some nodes unrelated to directory structure so under this assumption path names would need to change.

> I think we need to do chunking and dag building ourselves as the data is ingested because we don't know how much data there is or if the local daemon can take it.

This makes sense.  My perspective above was that you were proposing sending manageable chunks off to each cluster peer so that they would be sure their ipfs daemons could take it.  This could still be something we do especially in the case of adding data that can't fit on any repo.

> My idea is that we chunk it in cluster like ipfs does. And as leaves are produced, we group them, and once grouped we find them an allocation and forward the chunk there, where it is injected to ipfs...

I think I'm still missing something.  At the top you said that we are still producing the ipfs dag, but now it seems like you are saying that we are not, only chunking the data the same way as ipfs and building the cluster dag.  Otherwise how is the ipfs dag getting built?  Is it being imported in parallel with the cluster-dag block aggregation and distribution among peers?  If so it seems like it violates the above assumption that the data may not fit in a repo.

> where it is injected to ipfs (with block/put because it's already chunked)

Got it, using block put makes sense now

> the clusterdag is a CID. The field links the original ipfs-dag CID to the Cluster DAG. It is how we know that for a given CID, cluster is pinning a bunch of cluster-dag-subtrees rather than pinning directly the content like it works now.

Hmm ok got it.  Now it definitely seems like you are proposing building the ipfs-dag (to get the hash).  I was thinking of something along these lines so that cluster need not expose its internal mechanism to users and can just respond to queries on the ipfs-dag root hash.

Definitely interested in a call at some point, maybe tomorrow or Friday?

ZenGround0:
> * One day we might have to worry about DAGs too large to fit in a single ipfs node. If our system treat them as plain data, could it distribute them among cluster peers? Would that even work or be useful in such case?

This is closer to what I thought we were talking about in my original response.  Handling this would be necessary not only for very large ipfs dags, but also for sharding any dags that don't carry the bulk of their data in their leaves.  This will probably come up in the future but for now I am comfortable putting this aside.

hsanjuan:
>  Because the cluster dag takes care of sharding data across parts it necessarily adds in some nodes unrelated to directory structure so under this assumption path names would need to change.

I think this is the root of the misunderstanding (either yours or mine).

I have drawn it:

image

The user can still reference all the data according to the regular ipfs DAG, and the things have the paths they would have in ipfs etc.

However cluster creates an **additional** DAG which instead of grouping the data blocks by file and then by folder, does it by "parts". And so each "part" is cluster-pinned and allocated to a cluster peer.

The users would still reference their data with the original CID and paths from it. I.e. in the drawing, if the user does `ipfs cat <root_sum/file1.txt>`, ipfs will resolve that to the file1_sum, will fetch that DAG node and will find the list of blockhashes it needs to form the file1.txt. If it does not have them locally, it will check the DHT to see who has them and fetch them. In this case, 3 blocks would come from cluster peers pinning part1 and the last part from cluster peers pinning part2.

Is this what you understood? In that case, then it's me missing something.

ZenGround0:
@hsanjuan I *just* figured this out and posted an edit in my second to last response above:

>--edit--
Oh now maybe I see, if we keep the ipfs-dag minus the leaves then cluster peers can resolve any path through the ipfs-dag and then the leaves are pinned at the bottom of the cluster dag so they can be resolved from the node pinning them. Is this the right idea?
--edit--

ZenGround0:
@hsanjuan let me know if I'm missing anything else.  At this point I feel like I'm on the same page.

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,176 @@
...
+The root would be like:
+{
+  "parts" : [
+     {"/": <part1>},
+     {"/": <part2>},
+     ...
+     {"/": <partN>},
+   ]
+   "checksums" : [
+     {"/": <chksum1>},
```
hsanjuan: I am very new to this topic.

Each `checksum` is supposed to be formed from strips of data stored in every disk. It's straight forward on a location-addressed disk setting because you take the different words at position X and calculate the checksum(s).

But what are the words here that would make the first checksum (for example)? Is it the first block addressed by every cluster-dag "part"? What if every part has different block counts in them?

Can the checksum part be abstract to the state format and be stored just like another "part"?

Can the RS-encoding be something that can happen as a transform of a cluster dag ? (as a separate operation, after ingesting all data etc. you would say 'enableRSEncoding(cid)' and it would calculate checksums and rebuild the cluster dag to include them). Provided there's an allocator which ensures each part lives in a different cluster peer.

If erasure encoding/fault tolerance is a problem that can be dealt with separately from the importing/splitting data, we should probably move these sections to their own RFC. This looks like adding significant complexity.

It's great that you gathered some useful and usable papers! This makes a very good starting point to think on how to implement this stuff.

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,176 @@
...
+
+# Complications
+
+## Files larger than local repos
+
+### Improve cluster UX for large files
+"How much room do I have in my cluster?" is a very predictable question that users importing huge files will want to know.  We should include cluster cli tools (something like ipfs metrics, or the still-unmade connectivity-graph) that inform users how much space they have available, and the largest recommended file the cluster can ingest.  Perhaps later on we can get even fancier and include predictions of insertion time, and a visualization of how many shards of what size will be added.
+
+Although the ipfs add proxy endpoint is the logical place to launch the adding process we could also provide a user friendly command that takes in the option of a network address to add from.  This way cluster could properly manage the connection with a remote server while buffering and potentially blocking for long periods of time and prevent the user from having to do clever unix things to download large files.  Note I don't have much experience with this kind of thing, so maybe wget etc have really good UX for things like this and this would be overkill.
```
hsanjuan: the ipfs add proxy endpoint is not the logical point. Cluster may intercept that endpoint and do its thing, but cluster will provide it's own endpoints/methods.

> we could also provide a user friendly command that takes in the option of a network address to add from.  This way cluster could properly manage the connection with a remote server while buffering and potentially blocking for long periods of time and prevent the user from having to do clever unix things to download large files.

You mean like fetching a URL ? Yeah maybe, but let's not worry about this in the first round. It should be an easy thing to do if the rest is working.

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Challenges adding files
+When files become larger than local repos this becomes challenging for two reasons.  First, the importing node now needs to be able to block the importing process periodically to transfer data over to the pinning peer so that its local resources (ipfs repo and maybe memory) don't get overloaded.  Many related complications also seem likely.  For example if one shard takes up 90% of the node's ipfs repo and if the importing node makes use of its repo for temporary pinning and transfer then the importing node cannot pin its own shard until all other nodes have been served.  Second, I am unclear if either of the two proposed data transfer methods will provide acceptable performance when transferring large amounts of data (e.g. on the order of 10 GB shards).
```
hsanjuan: When cluster ingests data, it should chunk and forward each chunk to the final destination. This should not take local disk space and should be memory-conservative. Perhaps some intelligent chunking methods require to read a lot of data first (in a way that you have a ram memory bound), but we'll have to live with that. I'm thinking under the assumption that data can be chunked as it comes in in most cases (normally format-indexes are in headers so reading those headers helps you chunk the rest as it comes in).

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,176 @@
...
+
+The first challenge will require rethinking the go-ipfs dag building functionality and on the bright side may contribute to building a versatile foundation for the potential dagger/dex project.  The second challenge is best addressed by evaluating the performance characteristics of both data transfer methods.  Perhaps the ongoing ipfs metrics work will already answer many questions about the first method.  If the second transfer method (libp2p rpc) is a serious contender we should plan on evaluating its performance as part of implementation.  (TODO -- create an issue defining some metrics that we will want to evaluate) (TODO -- see if any existing ipfs metrics are relavent to first method, e.g. measuring remote pin time of dags with many large leaves)
```
hsanjuan: We are going to be sending chunks. current RPC will have a little overhead but not much. I think it's a perfectly valid way to transfer/forward the data (considering the advantages offered by underlying libp2p) once it's chunked.

What I don't think will work is to be able to send and process a stream (prior to chunking). For that we need a different way of forwarding that data among components (if we need to forward it).

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Collaborative importing
+We could potentially use alternatives to importing the stream on one master node that then sends out shards to the assigned pinning nodes.  One other possibility is "collaborative importing", where the cluster node doing the "add" operation streams the raw data out to the pinning nodes that then do importing and pinning in place.
```
hsanjuan: This is an interesting idea. I don't know how much improvement there would be between sending 'chunks' to other peers and 'sending a stream to chunk', and how difficult it would be and how much performance would be added. It seems simpler to chunk everything in one place.

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,176 @@
...
+
+One upside is that the interaction between the ipfs files dag and cluster-dag importing process becomes much more simple.  The original importer could run to completion uninterrupted, and then the node could build the cluster-dag and unpin the leaves from the ipfs files dag.  Althought this wouldn't directly address the issues brought up above it is possible collaborative importing would result in a simpler download process as now downloads and imports are not interleaved.  It also seems likely that, if done right, collaborative importing could be faster overall, as importing would be parallelized across nodes.
+
+
+## Arbitrary dags
+
+TODO -- I have some thoughts on these (many voiced in my looong comment in response to @hsanjuan's proposal), I need to take some time to organize and record them here
+
+### Dags too big to fit in ipfs repo
```
hsanjuan: should not be a problem by considering dags block just like any data block.

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Dags with data not in the leaves
```
hsanjuan: same as above.

The problem here is that it will make fetching slower because the ipfs daemon will need to fetch not only the data leaves but the intermediary nodes when resolving a path.

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,176 @@
...
+* metric sweeps
+  * design metrics (rpc and/or ipfs remote pin unpin, and state size) and sweeps
+  * if we want better performance then we can optimize (rpc) or build new transfer tools out of libp2p
+* improving state component to scale (@hsanjuan from your proposal open questions)
+* simple updates to state format to allow for cluster-dag support
+* probably should work on improving migration framework as significant state changes are in the works
+* adding functionality to ipfs-cluster add ipfs proxy endpoint
+* changing configs and user commands to include `--shard` on add, in general figure out UX to determine when we shard
+* potentially significant modifications to cluster peers signalling that a cid should be pinned (now they may be transferring data via rpc before hand) and so potentially significant modifications to how peers interact with state.  (@hsanjuan, this is something I haven't thought through enough to convince myself it won't be too different from now so I might be overstating the difficulty here)
+
```
hsanjuan: I think of a work plan little by little. For example, at first the importer does not need to do cluster-sharding, it just needs to work (like ipfs does). So:

* Need to start the `importers` project. Figure out what ipfs and ipfs-pack do to chunk data and create dags and create a library that does that. Paying attention that as data is streamed in, dags nodes are streamed out.
* Particularly important to figure out how unixFS file/folder data can be streamed (via reading from disk directly, vs. via http POST multipart upload...)
* cluster-Importer Component: figure out the interface it provides and implement it without the cluster-sharding.
* Using the importers project, the cluster-importer component should be able to split data, object/put it in local ipfs, and pin it in cluster. Figure out what API cluster provides to use the importer component. Including REST API part.
* Once that works, a second round on the importer allows it to build a cluster dag, object/put it in local ipfs and pin it in cluster (need to design cluster-dag format).
* Once that works, the state format is upgraded to include the cluster dag.
* Once that works, the importer allocates different cluster-dag-parts to different cluster-peers and ipfs-nodes.
* Once that works: scaling the state.
* Usability features etc...

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,176 @@
...
+Storing large files is a feature that would serve many use cases, the more general ability to store large ipld dags of arbitrary structure has the potential to support others.  As an example consider importing a cryptocurrency ledger to an ipfs cluster when no one node could hold the entire dag.
+
+## Motivation 2: store files with different replication strategies for fault tolerance
+This feature is also frequently requested, including in the original thread leading to the creation of ipfs-cluster (See issue #1), and in @raptortech-js 's thorough discussion in issue #9.  The idea is to bring RAID modes of storage, including modes that incorporate space-efficient FEC techniques, to provide better fault tolerance of storage.  This would open up ipfs-cluster to new usage patterns, especially storing archives with infrequent updates and long storage periods.
```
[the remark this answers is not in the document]
ZenGround0: I understand your perspective.  My reason for including RAID specifically stems from the following passage in the original cluster design proposal linked in issue #1 (https://github.com/ipfs/notes/issues/58):
> **We want**
>...
>*(Mostly Certain) Use RAID configurations for "pin-replication" cluster modes. 
>  * RAID modes are well understood and well documented. so they're the obvious default choice.
>  * but we should consider other well understood data replication configuration systems.

It would be helpful at some point to go through this proposal and issue #1 together so I can know for sure what is still relevant.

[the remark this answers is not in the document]
hsanjuan: Yeah I know where this comes from, but I think talking about RAID modes does not make understanding easier. You could say cluster now supports RAID 0 (replication level 1) and RAID 1 (replication level = 2). Is a larger replication factor a different RAID level?

It does not make sense to speak of byte-level stripping in cluster, so would are we talking about ipfs-block-level stripping instead? Or merkle-tree stripping? Would ipfs-block-level stripping make sense? For performance? Is it better that ipfs-blocks from one file are in the same daemon or potentially is as many daemons as possible? Most RAID modes are not really used (according to wikipedia), do we aim to support them?. How does the RAID 6 standard deal with disks being added or removed? 

Sorry but I find it super confusing to speak in terms of RAID vs:

* Your data is replicated X times in different nodes
* Or, your tree is distributed in different nodes and includes a checksum for recovery in case of disaster
* Or, each branch of your tree is stored in a different node

I don't even know how much of the RAID motivations and error correction codes are part of spinning magnetic disk scenarios (which have certain performance issues, and certain types of errors). So yep, I don't feel comfortable expressing our capabilities in terms of RAID levels.

hsanjuan:
We may also consider a single cluster-DAG rather than one per pinned-item.

this cluster-DAG could have a part for every current cluster peer which points to all things stored in that peer. Asigning an ipfs-block to a peer would be modifying that part (which points to all the ipfs-blocks tracked in that peer). ipfs provides object/patch methods which would allow to do this (adding items and removing from branches and obtaining the new root cid).

It would seem this requires touching more parts but it might make things easier in others, like block-stripping. It might help making the cluster state smaller (no need to track parts for every pinned item), but also create bottlenecks: patch operation might be slow and needs to happen sequentially. 

It also creates some problems if content is only pinned (not added directly from cluster). In that case, it might not be available, or ipfs may fail to fetch it. Maybe some cluster dag operations hang in that case or take too long because enumerating the subtree becomes a very expensive operation. On the other side, instead of looping tracked-cids for syncing, you could just re-pin the cluster-dag-part and that should ensure all items are pinned.

docs/dag-sharding-rfc.md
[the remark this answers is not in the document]
ZenGround0: @hsanjuan solid points all around.  Thank you for elaborating.

docs/dag-sharding-rfc.md
ZenGround0: Agree on all points especially on the difficulty implementing.  I'll take this out for now.  We can reconsider approaches like this if importing becomes a bottleneck on data streaming to an extent that specific use cases are suffering.

docs/dag-sharding-rfc.md
ZenGround0: FYI adding this to another place in the document and explicitly stating that it is a frozen WIP that will have its own RFC one day and that we will only pick it up after crushing the basics.

Good point, yes the same number of words is needed across parts when taking calculating codes.  We could always pad with zeros to the largest part in the case of uneven blocks, or better yet assume that if a word doesn't exist on a part then the checksum calculation adds in zero words in its calculation (coding and decoding).  This way the padding only needs to be stored in the checksum parts.

The words could be any sequence of bits (provided its length is above a certain threshold I think), so probably independent of blocks as block size will in general be different.  I was thinking that all the blocks would be concatenated together to get the sequence of bits needed for the calculation.

I am fairly certain the checksum can be stored like another part.  As it would be a single fragment and generally quite large (to match parts) we might run into some ipfs datastore blocksize limits and need to chunk it.  In that case we could rely on the same approach parts use to store their data in chunks.  This would require calling on the importer component when storing checksums.

Yes, we already went over this a little bit in our call since you wrote this, but check out the fanout and broadcast algorithms in Plank's paper if you are still curious.  I briefly address this in the Importing With Checksums section too.

docs/dag-sharding-rfc.md
ZenGround0: Thanks totally addresses concerns here

docs/dag-sharding-rfc.md
ZenGround0: And addresses concerns here too 🎉  🌮  🎉

docs/dag-sharding-rfc.md
ZenGround0: Got it.  On this topic could you teach me how to do this (e.g. get a buffered, lazily read stream of data from an http endpoint serving, for example, the ubuntu archive into the stdin of ipfs-cluster) with just unix tools and a shell?  It would make the process seem more friendly if I knew how to do it myself.

ZenGround0:
@hsanjuan Take a look at this.  It is significantly cleaned up and to the best of my knowledge fully addresses your latest comments.  Especially after going through your updated workplan bullets more carefully this is now looking very actionable.  Altogether pretty cool.

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,157 @@
...
+
+There will likely be several endpoints entering into the functionality for adding sharded data.  For example, there will be a go endpoint `Add(r io.Reader, opts...)` which just adds arbitrary data. There will be an HTTP endpoint to which users can post multipart data just like it is now posted to ipfs and which creates the unixfs merkle dag.  If dags can be streamed node by node through ipfs safely without possibly overloading the process and its repo, cluster will likely support an AddDag(chan <-*dagNode) endpoint which allocates already-imported-dags.
```
hsanjuan: I would remove the last sentence. After the example I would add that how the API endpoints look will be defined in a later phase. But maybe that's a bit vague for a proposal. The idea is that:

* We'll be able to receive arbitrary data, chunk it and distribute it in the cluster
* We'll be able to receive filesystem data, chunk it and distribute it in the cluster
* We'll be able to receive a CID, and distribute the referenced tree in the cluster

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,157 @@
...
+
+We will likely abstract the importing functionality into an importer component of cluster.  Following current conventions other cluster modules would communicate with this component over RPC.  However, the RPC module we use does not handle transmitting streams of data so this component and the components containing the endpoints described above will need to expose a stream friendly abstraction, potentially an HTTP endpoint or socket.
```
hsanjuan: From here til the end everything is in cursive... intended?

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,157 @@
...
+
+# Workplan
+
+There is a fair amount that needs to be built, updated and serviced to make all of this work.  Here I am focusing on basic sharding and allowing support for files that wouldn't fit in an ipfs repo.  Fault tolerance, clever sharding and support for sharding existing dags will come after this is nailed down.  
+
+* Build up a version of the `importers` (aka `dagger` aka `dex`) project that suits cluster's needs (if possible keep it general enough for general use).  In short we want a library that imports data to ipfs dags with arbitrary chunking and layout algorithms.  For our use case it is important that the interfaces support taking in a stream of data and returning a stream of blocks or dag nodes.
+* Figure out how unixFS file/directory data can be streamed (via reading from disk directly or via http POST multipart upload)
+* Implement a cluster importer component.  It will need to provide endpoints that supporting streams as input and output.
```
hsanjuan: I named it "sharding component" above to separate it from the importer/dagger/dex module

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,157 @@
...
+* Create a cluster add endpoint that uses the cluster importer component to split data and create dags.  Put the output blocks into the local ipfs daemon and pin each block with cluster.  This will require figuring out the importer component's api, including the REST API part.
```
hsanjuan: and pins the root cid, not "each block" (i mean, not directly).

docs/dag-sharding-rfc.md
ZenGround0: No, but if you view as markdown the effects go away.  I'll see if I can fix the raw text so github doesn't do this.

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,157 @@
...
+
+
+### Adding files
+Currently the proposed implementation of basic sharding relies on ipfs-cluster having an incoming stream of data that can be read lazily as cluster sends off blocks to remote peers.  This allows sharding files that are too big to fit in a single repo or whose dags are too big to read into memory all at once, as resources can be cleared while streaming occurs.  Only the cids of blocks in the shard need to be maintained to construct the shard node of the cluster-dag, not the entire block or in-memory node structure.  Therefore in our basic sharding implementation we assume that sharded data must be added to ipfs-cluster directly.  This is because, as far as we know, there is not currently a way to access such a stream of blocks belonging to an already-imported-dag without potentially overloading the local ipfs daemon's resources.
```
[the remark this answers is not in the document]
ZenGround0: @hsanjuan Are you suggesting we try to incorporate this in our first pass of implementing sharding?  I totally agree this should exist eventually but my understanding was that this is going to be experimental for a while for the reasons listed in this paragraph (please let me know if these reasons are overstated and you think this belongs in our first implementation).  

I could add your suggestion here but would prefer to put it in the later section "Sharding existing dags" which already exists to address this case.  My current view is that this part of the document describes the "basics" that we will work on first and later on I address how we will extend the basics in "Future Work"

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,157 @@
...
...o an importer to create an ipfs files dag chunked (size-splitter or rabin-splitter) and layed out (trickle or balanced) per user instruction.  Before importing blocks of each shard the cluster peer calls the `allocate()` function to find the pinning peer for the upcoming shard.  Concurrent with ipfs files dag creation and transmission of the dag's blocks to the allocated cluster peer, the importing ipfs-cluster node will oversee construction of a cluster-dag shard node using the cids of the blocks of the ipfs files dag.  Once a shard is filled with cid references it will also be transmitted and pinned to the allocated ipfs-cluster node.  This process continues until both dags are fully constructed and pinned as expected.  The importing node transmits blocks by calling `ipfs block put` and `ipfs pin` on the pinning node's ipfs daemon via an RPC to the pinning cluster peer.   The pinning node should check the state to ensure it was actually assigned the pin before running the rpc to completion.  Note that this will be one of the more involved parts of implementing basic sharding and will involve extracting, refactoring and adding features to the dag building functionality of go-ipfs.
```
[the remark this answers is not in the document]
ZenGround0: I agree that would be ideal.  With this in mind I was pessimistic enough about this working out that I put this here.  I can walk back the language a bit until I am 100% that we can't reuse everything.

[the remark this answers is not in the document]
ZenGround0: One thing to keep in mind along the lines of reusing code is that the dagbuilder helper *does* output dagnodes as it creates the layout.  IIRC it outputs them to a blockstore which may or may not be the right abstraction.  Maybe we could hack together a "stream" implementation of a blockstore and use that to our advantage.

docs/dag-sharding-rfc.md
ZenGround0: SGTM (also fixes github's italics formatting issue!)

docs/dag-sharding-rfc.md
ZenGround0: Ok, noted.  I will separate the two logically for now in the RFC.  FYI in my opinion the cluster dag builder's code will eventually belong abstracted somewhere in the importer/dagger/dex (This thing really needs a standard name :) ), but we'll cross that bridge when we come to it.

docs/dag-sharding-rfc.md
ZenGround0: 100% with you.  I'll make it clearer (pin root recursively)

docs/dag-sharding-rfc.md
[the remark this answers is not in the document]
hsanjuan: It's actually easier to do because you don't need the importer part (or your importer is just something that fetches from ipfs). I am wondering if we should do this first. But I also think the importers are really important. In any case, I don't think this is so complex that cannot be done along with the first round of things or shortly after.

docs/dag-sharding-rfc.md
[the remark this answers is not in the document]
hsanjuan: Yeah, I really need to spend some time looking at how things work in ipfs and pack, because right now I can't say much.

docs/dag-sharding-rfc.md
hsanjuan: it comes from chan <-*dagNode. I guess just quoting that will help.

docs/dag-sharding-rfc.md
hsanjuan: I think so too.

ZenGround0:
@hsanjuan I've been spending some time thinking about the open questions you've listed to provide some first attempts at solutions before finishing the latest revision.  Let me know what you think of using HTTP chunked transfer-encoding like this: https://gist.github.com/ZenGround0/af448f56882c16aaf10f39db86b4991e to solve the "upload method" question.  I think it allows us to sufficiently manage resources and fits well with the client or restapi components.

hsanjuan:
Regarding "How to add data to ipfs-cluster"?

* Adding via HTTP API (push) (thus enabling something like `ipfs-cluster-ctl add`):
  * POST data to endpoint
    * Single file/content would allow to read/stream the content. This would allow potentially unlimited upload size.
    * Adding multiple files/folders in a single would require a multipart request. Usually each part of the multipart is assembled separately. This would allow unlimited number of files, but the size of each would be constrained by the local-peer memory.
    * Can multipart be treated as a stream? Not from the looks of standard tooling (each part needs to be done before it can be read), might have to write our own if we need that...
    * Could also create a custom multipart if that's too difficult
* Adding locally via `ipfs-cluster-service add <filesystempath>`
  * Would mean reading from local disk/device.
  * Very easy to stream things as they are read (no need to handle connections)
* Importing from somewhere (pull) via `ipfs-cluster-service import <url>`
  * Ability to grab things directly from S3, or any other place able to hold data
  * Can [possibly] discover unixfs layouts before downloading the stuff, so good for streaming.

So all these methods mildly overlap. In the end, we usually want:

* A stream of data
* Unixfs metadata: is it a file, is it a folder wrapping that file? (no metadata would mean everything is a single file).

The "importer" component (for the chunking) will need to receive these two things (either via a go reader object, a channel, a filesystem `pipe`...). Ideally, internally, both the REST api component and the `ipfs-cluster-service` commands submit the data to the importer in the same way. Probably once we know how the "importer" looks like, it will become more clear what the best way to use it from the different places in the code.

ZenGround0:
@hsanjuan I've written a proof of concept demonstrating http streaming with a multipart content-type: https://gist.github.com/ZenGround0/49e4a1aa126736f966a1dfdcb84abdae.  Only standard libraries, no custom parser needed.  I think this should make our life a little easier.

hsanjuan:
We'll that's really nice @ZenGround0 . Guess we can use multipart upload in the rest API. We just need to decide how to send the stuff from there to the chunker then.

GitCop:

 * Commit: ce9e7621e3e79940bbf106da43ec4d1fab18bfcc
    **License: MIT  
Signed-off-by:**

---

ZenGround0:
@hsanjuan when you a chance I'm looking forward to your feedback.  Thanks!

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,182 @@
...
+
+The golang multipart-reader exposes a reader to each part of the body in succession.  When the part's body is read completely, a reader to the next part's body can be generated by the multipart-reader.  The part meta-data (e.g. this part is a directory, that part is a file) and the data read from the part, must then be passed into the importer component to begin the next phase of the pipeline.  We can explore different methods of this transfer in pursuit of efficiency and simplicity.  One approach would be polling the part readers in the HTTP API and transferring data as it arrives over RPC, while another would be to assume the importer is on the same machine as the HTTP API and send the multi-part reader, or the readers to individual parts, over a channel.
```
hsanjuan: I've been thinking a bit, we'll just have an importer that we can instantiate inside the http api. It would be an external library, so the http api component will do the chunking for api uploads. Once a dag block is produced, we send that via RPC to the sharding component.

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,182 @@
...
+
+As this endpoint will be used for streaming large amounts of data it has the potential to be somewhat difficult to use as system limits could be pushed and errors could occur after annoyingly long time investments.  Some questions that may come up as useability becomes a bigger concern:
+
+* Is it possible that remote users pushing data to the endpoint could overload cluster's resources by pushing at a rate faster than processing can occur in cluster? If this is an issue how can the endpoint maintain connections in a resource-aware non-degrading way?  Will it be worthwhile to open up an endpoint running a different protocol (i.e. libp2p stream-muxer)?
```
hsanjuan: We might choose to provide the HTTP API via libp2p which might be a good thing for this too.

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,182 @@
...
+* What features can cluster provide to make the process easier?  For example if cluster could report an estimate aggregate of all of the space in its nodes' repos then users could check that a huge file will fit on the cluster before importing it.
```
hsanjuan: And what to do when importing fails half through and you've pinned a bunch of things already

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,182 @@
...
+* What is the distribution of user preferences regarding pushing or pulling data to or into the endpoint?  Is it worth including location addressed link resolution (i.e. `ipfs-cluster-ctl add --shard http://www.awesomefilez.warez`), or file system device reading support?
```
hsanjuan: I guess it's safe to say we will implement first adding via the http api, and then we might provide additional methods (like pulling). Adding via http api should cover adding from filesystem device.

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,182 @@
...
+* Any others?
+
+#### Sharding existing DAGs
+An endpoint should also exist for sharding an existing DAG among an ipfs-cluster.  A user will run something like `ipfs-cluster-ctl pin --shard <cid>` and pin shards of the DAG rooted at `cid` across different nodes.  The importing step of the pipeline can be skipped in this case.  The naive implementation is to interact with the local daemon's DAG API to retrieve a stream of blocks as input to the sharding component.  We currently suspect that this is not going to work well for large DAGs that cannot fit in the node's repo.  The "Future Work" subsection: "sharding large existing DAGs" discusses some possibilities for scaling this to large DAGs.
```
hsanjuan: Good point, still pretty useful to take something from a huge node and distribute it among several

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Importer module
+
+As they are currently implemented in ipfs, importers run in three main phases.  The importer first begins by traversing a collection of data and metadata that can be interpreted as files and directories.  For example the tar importer reads data from a tar archive and the file importer used by `ipfs add` iterates over a collection of file objects.  Files are then broken into chunks with a chunker.  These chunks are then used as the data to make the dag-nodes constructed by the dag layout building process.  The dag-nodes are serialized as blocks and saved to the repo.
```
hsanjuan: I wish such clear concise explanations of how things work would be written way more often in our projects 👍

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,182 @@
...
...tream of blocks to a peer for pinning.  When blocks arrive the sharding process records the cids which are needed for constructing links between the shard node and the blocks of the DAG being sharded.  This is all the sharding component must do before forwarding blocks to other peers.  The peer receiving the shard is calculated once by querying the allocator.  The sharding component must be aware of how much space peers have for storing shards and will maintain state for determining when a shard is complete and the receiving peer needs to be recalculated.  The actual forwarding can be accomplished by periodically calling an RPC that runs DAG API `put` commands with processed blocks on the target peer.  After all the blocks of a shard are transmitted the relevant subgraph of the Cluster DAG is constructed and sent as a final transmission to the allocated peer.  Upon receiving this final transmission the allocated peer recursively pins the node of the Cluster DAG that it receives, ensuring that all blocks of the shard stay in its ipfs repo.  When all shards are constructed and allocated the final root node pointing to all shard node is pinned (non-recursively) throughout the cluster.
```
hsanjuan: We do need to make sure that ipfs GC does not run between the moment when we add the first block of a shard and the moment we pin the shard. This is workaroundable by pinning every block individually (and later unpinning leaving only the recursive), but probably at a high price.

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,182 @@
...
+
+# Workplan
+
+There is a fair amount that needs to be built, updated and serviced to make all of this work.  Here I am focusing on basic sharding, allowing support for files that wouldn't fit in an ipfs repo.  Support for fault tolerance, more clever sharding strategies and sharding large existing DAGs will come after this is nailed down.  
+
+* Build up a version of the `importers` (aka `dagger` aka `dex`) project that suits cluster's needs (if possible keep it general enough for general use).  In short we want a library that imports data to ipfs DAGs from arbitrary formats (tar, deb, unix files, etc) with arbitrary chunking and layout algorithms.  For our use case it is important that the interfaces support taking in a stream of data and returning a stream of blocks or DAG nodes.
+* Implement a cluster importer component.  It will need to provide endpoints that supporting streams as input and output.
```
hsanjuan: probably this can be removed as `dex` will do the job. and the rest api will use it directly.

docs/dag-sharding-rfc.md
```diff
@@ -0,0 +1,182 @@
...
+* Create a cluster add endpoint on the REST API that reads in file data on HTTP multipart requests.  Implement forwarding of data to the cluster importer component.  Put the output blocks into the local ipfs daemon and pin the root with cluster.  This will require figuring out the connection strategy between the importer component and the REST API and nailing down the APIs of these components.
```
hsanjuan: Would need rewrite:  chunking happens directly in the component using `dex`. Might need a simple sharding component already, even if it doesn't shard and just allocates blocks.

docs/dag-sharding-rfc.md
ZenGround0: good point

GitCop:

 * Commit: ce9e7621e3e79940bbf106da43ec4d1fab18bfcc
    **License: MIT  
Signed-off-by:**

---

docs/dag-sharding-rfc.md
```diff
@@ -87,35 +87,35 @@ Under this implementation replication of shards proceeds like replication of an
...
+An endpoint should also exist for sharding an existing DAG among an ipfs-cluster.  A user will run something like `ipfs-cluster-ctl pin --shard <cid>` and pin shards of the DAG rooted at `cid` across different nodes.  The importing step of the pipeline can be skipped in this case.  The naive implementation is to interact with the local daemon's DAG API to retrieve a stream of blocks as input to the sharding component.  While this implementation would still be quite useful for distributing DAGs shards from a node with a large amount of repo storage across a cluster, we currently suspect that this is not going to work well for large DAGs that cannot fit in the node's repo.  The "Future Work" subsection: "sharding large existing DAGs" discusses some possibilities for scaling this to large DAGs on nodes of arbitrary repo size.
```
hsanjuan: Typo, missing "add": `ipfs-cluster-ctl pin add --shard <cid>`

ZenGround0:
@hsanjuan thank you for all the review.  I'll close this now and post to ipfs/notes.  And word, I'll happily point people to the RFC at the next all hands.

hsanjuan:
@ZenGround0 maybe merging would be good, so it can be ammended via PR... (regardless of posting to notes or not)

GitCop:

 * Commit: f7896477b31bfb0f8c6c7d29f93ab63b1f5f6afd
    **License: MIT  
Signed-off-by:**

 * Commit: 7c61915ef1a648499e988927e1a6990df00474ac
    **License: MIT  
Signed-off-by:**

 * Commit: f572df01adc973b392d18a23eb088c32963ed69c
    **License: MIT  
Signed-off-by:**

 * Commit: ac49ade5530bce71a6f51493c487fe0a7bb8161f
    **License: MIT  
Signed-off-by:**

 * Commit: 51c6bd95f98c28ffd19fb740df1eb3a4c9c678f1
    **License: MIT  
Signed-off-by:**

 * Commit: f602cc2eae4d088333e776611e63d3ab7247b704
    **License: MIT  
Signed-off-by:**

 * Commit: 08893851e59351db303c484d3920239fc16a247e
    **License: MIT  
Signed-off-by:**

 * Commit: 2b4b919e6ff9ffe57048400d4d00cb11934fa35b
    **License: MIT  
Signed-off-by:**

---

PR #274, opened 2017-12-12
cluster-service: add version subcommand and change some startup logging

### LINKED ISSUE #201: ipfs-cluster-service should have a version subcommand (rather than a flag)
--version is not a modifier of the behaviour, it has it's own behaviour. Should be a subcommand. (keep supporting `--version` via a hidden flag).

hsanjuan:
The --version flag is default from our cli library so I left that. The
version subcommand prints only the version number + the short commit
so it's a bit more easy to parse.

I have additionally reduced the amount of output on start up by converting
some messages to debug. I wish there was a level between INFO and DEBUG
though.

Fixes #201 

License: MIT

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -86,7 +86,7 @@ func NewCluster(
 		return nil, err
 	}
 
-	logger.Infof("IPFS Cluster v%s listening on:", Version)
+	logger.Infof("IPFS Cluster v%s-%s listening on:", Version, Commit[0:8])
 	for _, addr := range host.Addrs() {
 		logger.Infof("        %s/ipfs/%s", addr, host.ID().Pretty())
 	}
@@ -450,7 +450,7 @@ func (c *Cluster) ready() {
 	}
 	close(c.readyCh)
 	c.readyB = true
-	logger.Info("IPFS Cluster is ready")
+	logger.Info("** IPFS Cluster is READY **")
 }
 
 func (c *Cluster) bootstrap() bool {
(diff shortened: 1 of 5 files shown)


consensus/raft/raft.go
```diff
@@ -108,15 +108,15 @@ func newRaftWrapper(peers []peer.ID, host host.Host, cfg *Config, fsm hraft.FSM)
...
+		logger.Info("initializing raft cluster")
 		err := hraft.BootstrapCluster(cfg.RaftConfig,
 			log, stable, snap, transport, srvCfg)
 		if err != nil {
 			logger.Error("bootstrapping cluster: ", err)
 			return nil, err
 		}
 	} else {
-		logger.Info("raft cluster is already bootstrapped")
+		logger.Debug("raft cluster is already bootstrapped")
```
ZenGround0: As you changed bootstrapping to initializing above it seems like you should do it here too for consistency.

cluster.go
```diff
@@ -86,7 +86,7 @@ func NewCluster(
 		return nil, err
 	}
 
-	logger.Infof("IPFS Cluster v%s listening on:", Version)
+	logger.Infof("IPFS Cluster v%s-%s listening on:", Version, Commit[0:8])
```
ZenGround0: As TestClusterStateSync doesn't hit the cluster service version command this slice dereference seems to be the cause of the slice out of bounds error that Travis catches for these changes.

hsanjuan: Thanks

Issue #253, opened 2017-11-29
"ipfs-cluster-service state <import/export/cleanup>" commands


hsanjuan:
This is a follow up to #220 . We should have more commands to interact with the cluster state and make it user-readable.

ZenGround0:
PR #276

Issue #230, opened 2017-11-10
Mapstate Migration framework doesn't scale for development 


ZenGround0:
This is something I noticed working on PR #220.  The obvious way to allow upgrades when introducing a new state format is to add transformation logic from every old format to the new one.  This will start getting cumbersome after a while.  One better way to get from version x to version y would be to call "(version x+1).TransformFrom(versionx)" in a loop until arriving at verison y.  This way each new state format only needs to implement a transformation from its immediate predecessor.  (This solution would still require keeping around old state format type definitions.  If this starts to take up too much space we could also move towards doing something like ipfs and downloading dependencies at runtime.)

ZenGround0:
PR #283 attempted to address this but probably could use a little more polish.  The only way we're really going to get a sense for how to do this is after updating the state format a few times.  Let's keep this open until we've gone through that process and are satisfied.

ZenGround0:
So far the process has been pretty painless and the primary concern voiced here of needing to write a migration from every old state for each new state is taken care of.  We can revisit further improvements to the migration framework in the future in other issues.

PR #276, opened 2017-12-19
Feat #253 state import/export/cleanup

ZenGround0:


ZenGround0:
sharness tests on their way

cluster_config.go
```diff
@@ -409,32 +411,33 @@ func clusterSecretToKey(secret []byte) (string, error) {
...
+		return errors.New("skipping export")
+	}
+	fname := time.Now().UTC().Format("20060102_15:04:05")
+	f, err := os.Create(filepath.Join(folder, fname))
+	if err != nil {
+		logger.Error(err)
+		return err
+	}
+	defer f.Close()
+	err = state.Snapshot(f)
```
hsanjuan: I got very confused with the `Snapshot` name function. It's not the same format as Raft snapshots.

We may want to just use the result of `mapstate.List()` (which each item in the list converted to PinSerial and then everything to JSON). And with that drop the Snapshot method. The Restore method should probably just be used for Migrations (perhaps renamed to Migrate() and without doing json-unmarshaling).

The import would be then a for-loop of `Add()` operations (which can be implemented as `MultiAdd([]api.Pin)`).

I think this would allow to export a state and re-import it to a different `state` implementation which is something nice to have. JSON marshaling and unmarshaling can be ipfs-cluster-service task.

ipfs-cluster-service/main.go
```diff
@@ -261,7 +261,82 @@ removal, upgrade state using this command, and restart every peer.
...
 						checkErr("upgrading state", err)
-						return err
+						return nil
+					},
+				},
+				{
+					Name:  "export",
+					Usage: "save the IPFS Cluster state to a human readable json file",
+					Description: `
+
```
hsanjuan: We need to unify the command description spacing. The ones you added have an 2 line-breaks before and after. `init` has not.

ipfs-cluster-service/main.go
```diff
@@ -261,7 +261,82 @@ removal, upgrade state using this command, and restart every peer.
 					Action: func(c *cli.Context) error {
 						err := upgrade()
 						checkErr("upgrading state", err)
-						return err
+						return nil
+					},
+				},
+				{
+					Name:  "export",
+					Usage: "save the IPFS Cluster state to a human readable json file",
```
hsanjuan: is it human readable?

ipfs-cluster-service/main.go
```diff
@@ -261,7 +261,82 @@ removal, upgrade state using this command, and restart every peer.
...
+					Description: `
+
+This command reads the current cluster state and saves it as json for 
+human readability and editing.  In general any version of cluster will only
```
hsanjuan: Let's not say "in general", either things are going to be migrated or not (edit `FromJson` does not seem to allow for any migrations).

If  my recommendations above are taken, we can get rid of Restore (and FromJSON). As long as `api` types don't change in an incompatible way, we should be able parse to re-import states via calls to Add(). And otherwise it's always easy to parse some JSON and transform it with a helper tool. 

It's not a crazy requirement to say that exported states should only be imported in the same cluster version, and then use the regular `upgrade` path to upgrade.

 We can revisit upgrading paths as part of #230 anyways.

ipfs-cluster-service/main.go
```diff
@@ -261,7 +261,82 @@ removal, upgrade state using this command, and restart every peer.
...
+						checkErr("exporting state", err)
+						return nil
+					},
+				},
+				{
+					Name:  "import",
+					Usage: "load an IPFS Cluster state from an exported human readable json file",
+					Description: `
+
+This command reads in a json encoding of the state specified in the argument
```
hsanjuan: "reads an exported state file"

ipfs-cluster-service/main.go
```diff
@@ -261,7 +261,82 @@ removal, upgrade state using this command, and restart every peer.
...
+storing the state as a persistent snapshot to be loaded as the cluster state
+when the cluster peer is restarted.  Paths are assumed to be relative to the
```
hsanjuan: hmm just open the given path directly. `ipfs-cluster-service import state.json` should read state.json from the local folder, regardless of the cluster base directory.

ipfs-cluster-service/main.go
```diff
@@ -261,7 +261,82 @@ removal, upgrade state using this command, and restart every peer.
...
+cluster base directory.  In the case no matches are found the path is then
+assumed to be absolute.
+`,
+					Action: func(c *cli.Context) error {
+						// Get the importing file path
+						importFile := c.Args().First()
```
hsanjuan: as an idea, `ipfs-cluster-service import` without arguments could support reading from stdin..

ipfs-cluster-service/state.go
```diff
@@ -44,11 +46,62 @@ func upgrade() error {
...
+
+	stateToExport := mapstate.NewMapState()
+	err = stateToExport.Restore(r)
+	if err != nil {
+		return err
+	}
+	return clusterCfg.ExportState(stateToExport)
+}
+
+func stateImport(r io.Reader) error {
```
hsanjuan: I think before calling this we need to explain the user that the peer's state is going to be fully replaced.

Also:

* Does import need to run in all peers?
* Probably import is not meant to run in a running cluster, so we need to lock (same for export).

ipfs-cluster-service/main.go
```diff
@@ -261,7 +261,82 @@ removal, upgrade state using this command, and restart every peer.
...
+					Usage: "save the IPFS Cluster state to a human readable json file",
+					Description: `
+
+This command reads the current cluster state and saves it as json for 
+human readability and editing.  In general any version of cluster will only
+work with the latest state format, so out of date snapshots will be migrated
+before the export.
+`,
+					Action: func(c *cli.Context) error {
+						err := export()
```
hsanjuan: `ipfs-cluster-service state export -f mypath/state.json` should write to mypath/state.json. `ipfs-cluster-service state export` should write to stdout.

hsanjuan: Not sure this can happen while cluster is running. maybe since it reads the last snapshot. We might refuse to run while cluster runs anyway.

ipfs-cluster-service/main.go
```diff
@@ -261,7 +261,82 @@ removal, upgrade state using this command, and restart every peer.
...
+					Name:  "cleanup",
+					Usage: "cleanup persistent consensus state so cluster can start afresh",
+					Description: `
+This command removes the persistent state that is loaded on startup to determine this peer's view of the
+cluster state.  While it removes the existing state from the load path, one invocation does not permanently remove
+this state from disk.  This command renames cluster's data folder to <data-folder-name>.old.0, and rotates other
+deprecated data folders to <data-folder-name>.old.<n+1>, etc for some rotation factor before permanatly deleting 
+the mth data folder (currently defaults to 5)
+`,
+					Action: func(c *cli.Context) error {
```
hsanjuan: Ask for confirmation, override with global `-f`

ipfs-cluster-service/main.go
```diff
@@ -261,7 +261,82 @@ removal, upgrade state using this command, and restart every peer.
...
+						_, _, _, _, consensusCfg, _, _, _, _ := makeConfigs()
+
+						// Execution lock
+						err := locker.lock()
+						checkErr("acquiring execution lock", err)
+						defer locker.tryUnlock()
+
+						dataFolder, err := raft.MakeDataFolder(consensusCfg.BaseDir, consensusCfg.DataFolder)
```
hsanjuan: I think it's not necessary to export/use this. CleanupRaft ends up calling internally to mkdir.

ipfs-cluster-service/main.go
ZenGround0: Is "mypath" the basdir of the cluster config (i.e. same directory as service.json) or cwd?

ipfs-cluster-service/main.go
hsanjuan: yes, nvm

ipfs-cluster-service/main.go
ZenGround0: I am on board.  One implication of getting rid of Restore (Migrate) in this case is that `export` must be called with a version of `ipfs-cluster-service` with the same state representation as the snapshot being exported.  Unmarshal won't do anything besides version parsing on other state formats.  I am fine with this assumption, just making it explicit.

My wording here was confusing but your suggested assumption that exported states should only be imported in the same cluster version was exactly what I was trying to communicate.

ipfs-cluster-service/state.go
ZenGround0: Are you thinking an interactive "Do you want to override state [y/n]" prompt would be good here?  

In most cases I think import would need to be run in all peers.  The implementation calls raft.SnapshotSave which if you recall takes the term and index values from the latest snapshot.  If all peers saved at the same place along the log then you would see conflicts if their states differed.  However if a peer was already ahead it could import ok because other peers would need to catch up anyway.  For UX purposes its probably best to say that it should be done on all peers.  

I'll put the warning that all peers need to import the same thing alongside the [y/n] prompt (assuming this is what you are suggesting).

ZenGround0: Also in case it wasn't clear the way this is set up right now the state, while difficult for a user to reset, is not totally wiped from disk, just rotated to an old folder by SnapshotSave's internal cleanup operation.  Regardless I'm in favor of the prompt.

ipfs-cluster-service/main.go
ZenGround0: Well I guess it depends on the human, definitely more so than msgpack :) if you prefer I can remove this wording

ZenGround0:
@hsanjuan I am happy with this.  Would you mind taking another look?

ipfs-cluster-service/main.go
hsanjuan: > One implication of getting rid of Restore (Migrate) in this case is that export must be called with a version of ipfs-cluster-service with the same state representation as the snapshot being exported

ok, we can do the version check just like when cluster is started and warn the user if it doesn't match right?

> Unmarshal won't do anything besides version parsing on other state formats

I'm not sure what you mean here

ipfs-cluster-service/state.go
```diff
@@ -44,11 +42,62 @@ func upgrade() error {
...
+	if !snapExists {
+		return errNoSnapshot
+	}
+
+	stateToExport := mapstate.NewMapState()
+	bs, err := ioutil.ReadAll(r)
+	if err != nil {
+		return err
+	}
+	err = stateToExport.Unmarshal(bs)
```
hsanjuan: Calling Restore() here would do the same with the advantage that we'd upgrade any old state before exporting right ? We may wanna do that, unless I'm missing some downside

ipfs-cluster-service/state.go
```diff
@@ -44,11 +42,62 @@ func upgrade() error {
 	}
 
 	// Reset with SnapshotSave
-	err = raft.SnapshotSave(consensusCfg, newState, clusterCfg.ID)
+	return raft.SnapshotSave(consensusCfg, newState, clusterCfg.ID)
+}
+
+func export(output string) error {
```
hsanjuan: most of this function is very similar to upgrade() isn't it?

should be abstracted into `restoreStateFromDisk` (which runs state.Restore() (or Migrate())?

and then upgrade() and export() can re-use that, one of them saving a new snapshot and the other writing to disk the state.

ipfs-cluster-service/main.go
```diff
@@ -251,17 +252,114 @@ configuration.
...
+the mth data folder (m currently defaults to 5)
+`,
+					Action: func(c *cli.Context) error {
+						err := locker.lock()
+						checkErr("acquiring execution lock", err)
+						defer locker.tryUnlock()
+
+						if !c.GlobalBool("force") {
+							logger.Info("Cleaning up cluster state.  All other peers must also cleanup to avoid raft errors")
+							if !yesNoPrompt("The peer's state will be removed from the load path.  Existing pins may be lost.  Continue? [y/n]") {
```
hsanjuan: colon at the end `[y/n]:`

ipfs-cluster-service/main.go
```diff
@@ -251,17 +252,114 @@ configuration.
...
+deprecated data folders to <data-folder-name>.old.<n+1>, etc for some rotation factor before permanatly deleting 
+the mth data folder (m currently defaults to 5)
+`,
+					Action: func(c *cli.Context) error {
+						err := locker.lock()
+						checkErr("acquiring execution lock", err)
+						defer locker.tryUnlock()
+
+						if !c.GlobalBool("force") {
+							logger.Info("Cleaning up cluster state.  All other peers must also cleanup to avoid raft errors")
```
hsanjuan: Warning

hsanjuan: Actually, should only log this when the operation is going to happen (after this if block) or has happened.

Should not say that other peers must also be cleaned up. It may confuse the user. Maybe a single peer needs to be cleaned up.

Perhaps a `the %datafolder  folder has been rotated. Next start will use an empty state.` suffices at the end.

ipfs-cluster-service/main.go
hsanjuan: `This will rename the %datafolder folder and thus clean all consensus data from this cluster peer. Continue? [y/n]:`

ipfs-cluster-service/main.go
```diff
@@ -251,17 +252,114 @@ configuration.
...
+If an argument is provided, cluster will treat it as the path of the file to
+import.  If no argument is provided cluster will read json from stdin
+`,
+					Action: func(c *cli.Context) error {
+						err := locker.lock()
+						checkErr("acquiring execution lock", err)
+						defer locker.tryUnlock()
+
+						if !c.GlobalBool("force") {
+							logger.Info("Importing new state. All cluster peers must import the same state to avoid raft errors")
```
hsanjuan: log info after operation has happened. `the given state has been correctly imported to this peer. Make sure all peers have consistent states.`

ipfs-cluster-service/main.go
```diff
@@ -251,17 +252,114 @@ configuration.
...
+							if !yesNoPrompt("The peer's state will be replaced.  Existing pins may be lost.  Continue? [y/n]") {
```
hsanjuan: colon `[y/n]:`

hsanjuan: `Existing pins may be lost` sounds a bit tragic (like if things were to be unpinned). Perhaps it suffices with the first sentence, the state will be replaced. How about `This peer's state will be replaced. Run with -h for details. Continue? [y/n]:` ?

ipfs-cluster-service/main.go
```diff
@@ -251,17 +252,114 @@ configuration.
...
+								return nil
+							}
+						}
+
+						// Get the importing file path
+						importFile := c.Args().First()
+						var r io.ReadCloser
+						if importFile == "" {
+							r = os.Stdin
```
hsanjuan: If I'm correct, this should just look like the command is hanging when called without piping anything. Maybe it's worth doing a log.Info `Reading from stdin, Ctrl-D to finish` message there.

I don't know if there is an easy way to detect that stdin is interactively connected and only print that message in that case and not when something is piped, otherwise probably it won't hurt to print it in any case.

ipfs-cluster-service/state.go
hsanjuan: Sorry, just saw this after the new review. Maybe the new snapshot should have the last snapshot's index + 1. In this case, maybe all peers would receive the newest snapshot. I am not 100% certain. Let's leave it like it is.

ipfs-cluster-service/main.go
hsanjuan: I understand now. Yeah, maybe using Restore() directly is the best, as it will do unmarshaling or migration as needed, but will get something done (and as pointed out in the last round, code can be shared with upgrade()).

ipfs-cluster-service/main.go
hsanjuan: it's good, at first it wasn't clear to me that we weren't exporting msgpack directly because of the Snapshot() function.

ipfs-cluster-service/state.go
```diff
@@ -1,54 +1,94 @@
...
+	err, stateToExport := restoreStateFromDisk()
 	if err != nil {
 		return err
 	}
 
-	newState := mapstate.NewMapState()
+	return exportState(stateToExport, w)
+}
+
+func restoreStateFromDisk() (error, *mapstate.MapState) {
```
hsanjuan: It's customary to let the error be the last of return values

ZenGround0:
`golint` does not report any issues and the return statement issues are fixed.  The `make check` problem is fixed by giving golint the proper flag (https://github.com/golang/lint/issues/65).

Issue #200, opened 2017-10-26
Add a test for the dockerfile


hsanjuan:
`make docker`

Should:

* Make sure the dockerfile builds correctly
* Make sure cluster-service returns correct version
* Make sure cluster-ctl returns correct version
* At least...

It should prevent things like #196 from happening...

hsanjuan:
@ZenGround0 closeable?

PR #282, opened 2018-01-04
fix/basic-docker-tests

ZenGround0:
To resolve issue #200

License: MIT


hsanjuan:
LGTM. But seems to need rebase

hsanjuan:
Oh, it's because I enabled a config option so that everything can only be merged to latest master so that tests always run on top of latest master. But now that I think of it, travis PR build should cover that...

PR #283, opened 2018-01-05
Generalize migration process

ZenGround0:
This is meant to address issue #230.  It implements the ideas put forward there to help ease the development of state migration changes when format versions increase.  With so few supported versions at present the overhead may seem unnecessary but it should start obviously paying off when we get to version ~10.

@hsanjuan there is still a lot of updating required in the case of a version increase.  Suggestions to improve this are particularly welcome.

### DIFF
--- a/state/mapstate/migrate.go
+++ b/state/mapstate/migrate.go
@@ -1,5 +1,10 @@
 package mapstate
 
+// To add a new state format
+// - implement the previous format's "next" function to the new format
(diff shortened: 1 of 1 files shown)


hsanjuan:
> @hsanjuan there is still a lot of updating required in the case of a version increase. Suggestions to improve this are particularly welcome.

Will improve as we need it I guess...

Thanks!

PR #285, opened 2018-01-08
Fix #275: Attempt to wait for updates when shutting down

### LINKED ISSUE #275: Snapshots won't be taken until last index is applied
These messages are probably bad:

```
10:34:41.377 ERROR  consensus: could not snapshot raft: cannot take snapshot now, wait until the configuration entry at 12 has been applied (have applied 11). consensus.go:164
```

Might want to try to wait for the state to reach the last index before snapshotting.

hsanjuan:
Still thinking if worth to do retry-ing or something like that.

Did we ever try to Snapshot after shutdown?

There doesn't seem to be a way to tell raft to not take any more updates so this is a best effort...


ZenGround0:
Adding in a fixed number of retries sounds like a good idea.  I don't recall trying snapshot after shutdown.

hsanjuan:
Tests pass, might be worth a try

hsanjuan:
Will have to test on local cluster

hsanjuan:
Ok, I have been testing this. Snapshotting does not work after shutdown, but the approach here seems to work well enough (I have stopped a node while hammering another with pins so that it did need to wait for updates). There might be still changes it will not work, but they are considerably smaller than originally.

PR #288, opened 2018-01-09
Harden Sharness

### LINKED ISSUE #181: Harden sharness/lib/test-lib.sh
Rework the shell scripts in this file to have more meaninful errors, and better resiliance when a previous sharness test has failed to run and leaves the docker container running.

Also, configuration folder should probably make sure there is no data subfolder when initializing.

ZenGround0:
Recent sharness failures prompted more investigation.  In particular sharness `cleanup` does not run cleanup functions on failure (see here: https://github.com/chriscool/sharness/blob/master/sharness.sh#L763).  To ensure isolation among test failures we need reliable teardown of ipfs and cluster between each testfile so now the cleanup functions are run directly before test_done.  

Error messages are also slightly improved.

The current `test-lib.sh` already ensures that there is no data folder with `rm -rf ipfs-cluster-data` after the config folder is created.

Altogether this should close #181.

### DIFF
--- a/sharness/t0020-service-basic-commands.sh
+++ b/sharness/t0020-service-basic-commands.sh
@@ -4,9 +4,7 @@ test_description="Test service startup and init functionality"
 
 . lib/test-lib.sh
 test_ipfs_init
-cleanup test_clean_ipfs
 test_cluster_init
-cleanup test_clean_cluster
 
 test_expect_success "prerequisites" '
     test_have_prereq IPFS &&
@@ -30,5 +28,8 @@ test_expect_success "cluster-service --version succeeds and matches ctl" '
 test_expect_success "starting a second cluster-service process fails" '
     test_expect_code 1 ipfs-cluster-service --config "test-config"    
 '
+
+test_clean_ipfs
+test_clean_cluster
 
 test_done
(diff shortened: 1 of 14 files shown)


sharness/t0020-service-basic-commands.sh
```diff
@@ -31,4 +29,7 @@ test_expect_success "starting a second cluster-service process fails" '
     test_expect_code 1 ipfs-cluster-service --config "test-config"    
 '
 
+test_clean_ipfs
```
hsanjuan: Hmm it seems that `cleanup ...` does not clean when the tests have failed (even though documentation says it does? https://github.com/chriscool/sharness/blob/master/API.md#final_cleanup).

I understand this fixes that? The reason clean up does not run on that case is probably to allow the user to debug things. I'm fine with this though, since debugging implies running stuff locally and we can then potentially comment this.

hsanjuan:
merge if you want @ZenGround0

sharness/t0020-service-basic-commands.sh
ZenGround0: Yes.  It was tough to see this because as you say the docs are a bit misleading.  But it explained behavior we saw and I'm reasonably confident after looking at sharness source (see link above)

My thoughts exactly.  It makes sense if you want a single failure to stop everything so you can debug, but running this on CI we don't even have a chance to investigate the left over state.

ZenGround0:
@hsanjuan Will do.  I'll leave the harden sharness issue open as it sounds like you have more in mind.  If you are looking to offload that work make sure to write down your thoughts on the issue because right now I'm drawing a blank on what else needs to be done.

PR #290, opened 2018-01-11
Fix #240: Avoid duplicated list of facilities

### LINKED ISSUE #240: Logging facilities listed twice for twice the confusion
It took me too long to figure out why I was not seeing log output after adding a new facility in logging.go.  Turns out there is another list of facilities in ipfs-cluster-service/main.go that only applies in the case the debug flag is not passed in.  The two lists should be consolidated so that facilities need to be added only once.  This should be a pretty quick fix.

hsanjuan:
This puts some sanity in this. It's not super correct (name of facilities
depend of the component and the main cluster component should not hard
code them), but it's clear enough. Imho, better than over-engineering
a more elegant approach.

License: MIT


logging.go
```diff
@@ -4,18 +4,30 @@ import logging "github.com/ipfs/go-log"
...
+	"pintracker":  "INFO",
+	"ascendalloc": "INFO",
+	"diskinfo":    "INFO",
+	"apitypes":    "INFO",
+	"config":      "INFO",
+}
+
+// LoggingFacilitiesExtra provides logging identifiers
+// used in ipfs-cluster dependencies, which may be useful
+// for to display. Along with their default value.
```
ZenGround0: little typo: "for to display"

nodebug.go
```diff
@@ -2,14 +2,13 @@
 
 package ipfscluster
 
-// This is our default logs levels
+// These are our default log levels
```
ZenGround0: A general comment about nodebug, silent, debug, and logging: is it possible to put them somewhere a little more cohesive than simply in the main ipfscluster directory?  Until right now I hadn't put it together that all these files were part of the logging system and think it would aid people understanding the codebase to somehow group them better.  I can believe right now that they need to be part of ipfscluster and each use `init` functions so as it stands they can't be put into a single file or put into a subfolder.  Maybe it doesn't need to happen in this PR, but if you agree then maybe we could open an issue for this?

hsanjuan: Actually, I have realized that it is possible to pass flags to tests just like any executable. This nodebug etc files were just a hack to get such behaviour from long ago. I'll complete the PR getting rid of them.

hsanjuan: ok, how is looking now?

hsanjuan:
Hmm I broke something, travis loops over and over the same tests

hsanjuan:
@ZenGround0 made changes to `coverage.sh`. Will need another approval.

Issue #273, opened 2017-12-12
Code quality: address codeclimate issues


hsanjuan:
Codeclimate has surfaced a bunch of issues, like files too long, functions too complex or code duplicates. Some of them are valid, some of them are not really valid. The codeclimate interface allows to confirm/invalidate some of the stuff.

After that we can use this issue to reference all PRs which address codeclimate things.

kishansagathiya:
This is a little big refactor than I thought. Coding would be pretty straight forward once we agree on how to refactor this.
Let me know if this looks like a good plan to you

1. Move setting up the app object to a different file, say app.go
2. Move all api.Flags to a separate function and separate file say `flags() []cli.Flag`
3. Create a `commands` package. commands.go will contain `Commands() []cli.Command`. Each command would be in a separate file with a function like `Add() cli.Command` for add command.

kishansagathiya:
^ @lanzafame

kishansagathiya:
I think we can just close this or at least remove high priority from this.

PR #292, opened 2018-01-12
Fix #277: Introduce maximum and minimum replication factor

### LINKED ISSUE #277: [Feature] Sync only if min replicas reached
Could be interesting that ipfs-cluster only replicate a file if the minimum replicas are reached
Example:
max replication factor=7
min replication factor = 3

supose we have 10 nodes and 2 goes offline, nothing happens because we have 5 nodes with the replicated file.
but suppose 5 nodes that have same file goes offline so we have only the file replicated x 2, in this case is less than min replication factor so ipfs-cluster should automatically upload the file to a third node, if all the offline nodes go up again, the replication factor will be 8, so ipfs-cluster should remove the file from one of the nodes

hsanjuan:
This PR replaces ReplicationFactor with ReplicationFactorMax
and ReplicationFactor min.

This allows a CID to be pinned even though the desired
replication factor (max) is not reached, and prevents triggering
re-pinnings when the replication factor has not crossed the
lower threshold (min)

#277 

License: MIT


cluster_config_test.go
```diff
@@ -111,20 +112,51 @@ func TestLoadJSON(t *testing.T) {
...
+	}
+
+	j = &configJSON{}
+	json.Unmarshal(ccfgTestJSON, j)
+	j.ReplicationFactorMin = 0
+	j.ReplicationFactorMax = 0
+	tst, _ = json.Marshal(j)
+	err = cfg.LoadJSON(tst)
+	if err == nil {
+		t.Error("expected error when only rplMin/Max is not set")
```
ZenGround0: small issue: why use the word "only", it makes it sound like it is ok for rplMin/Max to be unset when something else is set.  I didn't see this from the code above, though perhaps I missed something.

state/mapstate/migrate.go
```diff
@@ -58,14 +65,40 @@ func (st *mapStateV2) unmarshal(bs []byte) error {
...
+	}
+	return &mst3
+}
+
+type mapStateV3 struct {
+	PinMap  map[string]api.PinSerial
+	Version int
+}
+
+func (st *mapStateV3) unmarshal(bs []byte) error {
```
ZenGround0: This is getting copied enough for us to consider using this as a default implementation for migrateable (if that's possible in golang).

state/mapstate/migrate.go
```diff
@@ -89,11 +123,11 @@ func (st *MapState) migrateFrom(version int, snap []byte) error {
 	for {
 		next = m.next()
 		if next == nil {
-			mst2, ok := m.(*mapStateV2)
+			mst3, ok := m.(*mapStateV3)
 			if !ok {
 				return errors.New("migration ended prematurely")
 			}
-			finalCopy(st, mst2)
+			finalCopy(st, mst3)
```
ZenGround0: I'm really interested to hear your thoughts about the current approach to updating the state.  Was it straightforward to figure out what had to be done for this specific instance? Were the reference directions at the top helpful?  Did you have any ideas for improvements to the framework while implementing this change?

hsanjuan:
ToDos:

* [ ] Decide how bad the codeclimate things are
* [ ] Modify rest api to work with the new replication factors
* [ ] Modify ipfs-cluster-ctl to allow providing the new factors
* [ ] Add tests specific for the new behaviours:
  * [ ] Should pin with Max when it can
  * [ ] Should pin with available when needed<=valid<Max
  * [ ] Should not pin at all when needed > valid
  * [ ] Should not repin when currentValid > needed

allocate.go
```diff
@@ -0,0 +1,180 @@
...
+//       may veto some depending on the allocation strategy.
+//     * The allocator returns a list of final candidate peers sorted by
+//       order of preference.
+//     * Take as many final candidates from the list as we can, until
+//       ReplicationFactorMax is reached. Error if there are less than
+//       ReplicationFactorMin.
+
+// allocate finds peers to allocate a hash using the informer and the monitor
+// it should only be used with valid replicationFactors (rplMin and rplMax
+// which are positive and rplMin < rplMax).
```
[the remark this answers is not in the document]
hsanjuan: yeah thanks

allocate.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	current := make(map[peer.ID]api.Metric)
+	candidates := make(map[peer.ID]api.Metric)
+
+	// Divide metrics between current and candidates.
+	for _, m := range metrics {
+		switch {
+		case m.Discard() || containsPeer(blacklist, m.Peer):
+			// blacklisted peers do not exist for us
```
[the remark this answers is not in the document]
hsanjuan: I changed it, it was more focused on the second part

allocate.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	return allocs
+}
+
+// getInformerMetrics returns the MonitorLastMetrics() for the
+// configured informer.
+func (c *Cluster) getInformerMetrics() ([]api.Metric, error) {
+	var metrics []api.Metric
+	metricName := c.informer.Name()
+	l, err := c.consensus.Leader()
```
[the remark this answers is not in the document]
hsanjuan: Yeah, that is correct. Monitoring is the place where Leader() leaks out of of the consensus layer.

The other easy alternative is to broadcast all metrics everywhere from everywhere. Way more resilient, at the expense of possibly bad scaling. (although for the cluster sizes we work with it might be doable).

We should come up with a better way to broadcast things like metrics though. Perhaps a ring-fashion where every peer forwards the metrics it knows about to the next one in the list. This will come up as a problem at some point I'm sure.

api/types.go
```diff
@@ -437,11 +438,12 @@ func PinCid(c *cid.Cid) Pin {
...
 type PinSerial struct {
-	Cid               string   `json:"cid"`
-	Name              string   `json:"name"`
-	Allocations       []string `json:"allocations"`
-	Everywhere        bool     `json:"everywhere,omitempty"` // legacy
-	ReplicationFactor int      `json:"replication_factor"`
+	Cid                  string   `json:"cid"`
+	Name                 string   `json:"name"`
+	Allocations          []string `json:"allocations"`
+	ReplicationFactor    int      `json:"replication_factor,omitempty"` //legacy
```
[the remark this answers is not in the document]
hsanjuan: To be honest, since we have state migration, it might just be ok to remove it. We don't have a policy, but communication between nodes running different versions can't happen because the libp2p streams are tagged with cluster version.

api/types.go
```diff
@@ -437,11 +438,12 @@ func PinCid(c *cid.Cid) Pin {
...
+	ReplicationFactorMin int      `json:"replication_factor_min"`
```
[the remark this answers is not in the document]
hsanjuan: Yes, see my comment in this pr from friday about what's missing

cluster.go
```diff
@@ -1274,104 +1281,6 @@ func (c *Cluster) getIDForPeer(pid peer.ID) (api.ID, error) {
 	return id, err
 }
 
-// allocate finds peers to allocate a hash using the informer and the monitor
```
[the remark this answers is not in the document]
hsanjuan: part of #273.. I try to do little by little anyway.

state/mapstate/migrate.go
hsanjuan: I don't think it's possible to have a default implementation for an interface. Probably not a horrible problem either to have copies of that.

state/mapstate/migrate.go
hsanjuan: It was straightfoward to copy paste the stuff and move on. I don't know if this can be done more elegantly, but works for me for the moment!

hsanjuan:
Most code is ready.

Still needs:

* To update the cluster guide explaining very well how this works
* Rebase on top of #293 and test migrations work as expected

ipfscluster_test.go
```diff
@@ -816,6 +816,275 @@ func TestClustersReplication(t *testing.T) {
...
+	}
+
+	p, err = safePeer.PinGet(h)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	secondAllocations := p.Allocations
+
+	if len(secondAllocations) != minInt(nClusters-2, 4) {
```
ZenGround0: We should have a more precise error message here and ideally remove the redundancy with the conditional on 1083 which will never get hit because of this conditional.  Maybe put this one after 1083 and report "insufficient reallocation, could have allocated to %d nodes but instead only allocated to %d nodes".

ipfscluster_test.go
```diff
@@ -816,6 +816,275 @@ func TestClustersReplication(t *testing.T) {
...
+
+	if fmt.Sprintf("%s", strings1) == fmt.Sprintf("%s", strings2) {
+		t.Error("allocations should have changed")
+	}
+
+	if len(secondAllocations) < 3 {
+		t.Error("allocations should be more than rplMin")
+	}
+}
+
```
ZenGround0: The last two tests hit the `needed <= 0` and `nCandidatesValid < needed` branches of obtainAllocations.  It doesn't look like we have a test for `wanted <=0` yet, i.e. repin with a smaller replication factor which would complete the set.

ipfscluster_test.go
```diff
@@ -816,6 +816,275 @@ func TestClustersReplication(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	p, err := clusters[0].PinGet(h)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(p.Allocations) != nClusters {
```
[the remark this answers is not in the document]
hsanjuan: You could always re-pin increasing the replication_min to max, forcing re-allocation on healthy nodes, and then re-pin reducing the replication_min again. Hmm, actually that might not work because it skips the pin and would not update the replication factor entries. I'll fix that..

Did I understand what you meant though?

ipfscluster_test.go
hsanjuan: done, but not that t.Error() does not abort the tests, it just marks as failure, so the second conditional would have happened too (unlike with t.Fatal).

ipfscluster_test.go
hsanjuan: Hmm yeah ok

ipfscluster_test.go
ZenGround0: ah thanks for the clarification

ipfscluster_test.go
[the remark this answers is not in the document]
ZenGround0: I believe so yes

hsanjuan:
@ZenGround0 ready for another look

allocate.go
```diff
@@ -40,11 +40,12 @@ import (
...
 // it should only be used with valid replicationFactors (rplMin and rplMax
 // which are positive and rplMin <= rplMax).
-// It only returns new allocations when needed. nil, nil means current
-// are ok.
+// It always returns allocations, but if no new allocations are needed,
+// it will return the current ones.
 func (c *Cluster) allocate(hash *cid.Cid, rplMin, rplMax int, blacklist []peer.ID) ([]peer.ID, error) {
 	// Figure out who is holding the CID
-	currentAllocs := c.getCurrentAllocations(hash)
+	currentPin, _ := c.getCurrentPin(hash)
```
ZenGround0: It looks like something subtle can happen that makes this function behave inconsistently with its specification.  Imagine pin C is allocated to all peers with a -1 -1 min/max factor.  Great.  From what I can see this creates an associated pin object in the shared state with an empty list of allocations as shorthand for "all peers allocate".  

The inconsistency comes in the edge case C is reallocated with a min factor lower than the nPeers but max is nPeers.  By the spec nothing should happen because no new allocations are needed and this number of allocations is acceptable (C is pinned at the max factor).  

However let's trace through and see what happens.  The code will read currentAllocs and hence set current metrics as an empty list.  It is a bit of a stretch but now imagine that one of the cluster peers was having connectivity troubles during the `getInformerMetrics` calls.  This peer is now invalid and so it is not considered in the candidate set.  `obtainAllocations` will now assign C to all but the peer marked invalid and so nPeers - 1 cluster peers now track C.  However by the spec nPeers cluster peers should track C as no changes to the allocation set were needed.

This is a pretty contrived example and it doesn't do much harm.  My bigger concern here is that interpreting the allocations field in two different ways in two different sections of the code is invalidating our assumptions and potentially a source of worse bugs that are tricky to find.

docs/ipfs-cluster-guide.md
```diff
@@ -217,7 +218,7 @@ Static clusters expect every member peer to be up and responding. Otherwise, the
...
 
-Dynamic clusters allow greater flexibility at the cost of stablity. Leave and, specially, join operations are tricky as they change the consensus membership. They are likely to fail in unhealthy clusters. All operations modifying the peerset require an elected and working leader. Also, bear in mind that removing a peer from the cluster will trigger a re-allocation of the pins that were associated to it. If the replication factor was 1, it is recommended to keep the ipfs daemon running so the content can actually be copied out to a daemon managed by a different peer.
+Dynamic clusters allow greater flexibility at the cost of stablity. Leave and, specially, join operations are tricky as they change the consensus membership. They are likely to fail in unhealthy clusters. All operations modifying the peerset require an elected and working leader. Note that peerset modifications may also trigger pin re-allocations if any of the pins from the departing cluster crosses the `replication_factor_min` threshold.
```
ZenGround0: Might be overkill but I think this goes from clear to exceptionally clear at the cost of a word if we say "if any of the pins from the departing cluster crosses below the `replication_factor_min` threshold."

sharness/t0052-service-state-export.sh
```diff
@@ -9,7 +9,8 @@ test_cluster_init
 
 
 test_expect_success IPFS,CLUSTER "state export fails without snapshots" '
-    cluster_kill && sleep 5 &&
+    cluster_kill
```
ZenGround0: Hmm why did getting rid of the &&s help?

docs/ipfs-cluster-guide.md
```diff
@@ -260,6 +261,12 @@ This involves:
 
 Errors in the first part of the process (before the entry is commited) will be returned to the user and the whole operation is aborted. Errors in the second part of the process will result in pins with an status of `PIN_ERROR`.
 
+Deciding where a CID will be pinned (which IPFS daemon will store it - receive the allocation) is a complex process. In order to decide, all available peers (those reporting valid/non-expired metrics) are sorted by the `allocator` component, in function of the value of their metrics (provided by the `informer` component). If a CID is already allocated to some peers (in the case of a re-pinning operation), those allocations are kept.
```
ZenGround0: I am unclear what you want to say by "in function of the value of their metrics".  Perhaps "as a function of the value" or "using a function of the value" or "in order of the value".

docs/ipfs-cluster-guide.md
```diff
@@ -260,6 +261,12 @@ This involves:
...
+
+New allocations are only provided when the allocation factor (healthy peers holding the CID) is below the `replication_factor_min` threshold. In those cases, the new allocations (along with the existing valid ones), will attempt to total as much as `replication_factor_max`. When the allocation factor of a CID is within the margins indicated by the replication factors, no action is taken. The value "-1" and `replication_factor_min` and `replication_factor_max` indicates a "replicate everywhere" mode, where every peer will pin the CID.
+
+Default replication factors are specified in the configuration, but every pinned as them associated to its entry in the *shared state*. Changing the replication factor of existing pins requires re-pinning them (it does not suffice to change the configuration). You can always check the details of a pin, including its replication factors, using `ipfs-cluster-ctl pin ls <cid>`. You can use `ipfs-cluster-ctl pin add <cid>` to re-pin at any time with different replication factors. But note that the new pin will only be commited if it differs from the existing one.
```
ZenGround0: I am unsure what you meant by "but every pinned as them associated to its entry"

ZenGround0: Perhaps you could end with "differs from the existing one in the way specified in the paragraph above" to be explicit that being different is not enough, you need to be outside of the new allowed range.  Though maybe this is unnecessary as you drive the point home in a few other places.

allocate.go
ZenGround0: FYI I'd be happy with a fix that simply made allocate aware of the special meaning of empty allocations and by checking for this case and setting all metrics to current.

ipfs-cluster-ctl/formatters.go
```diff
@@ -159,16 +159,16 @@ func textFormatPrintVersion(obj *api.Version) {
...
+	fmt.Printf("%s | %s | ", obj.Cid, obj.Name)
 
 	if obj.ReplicationFactorMin < 0 {
-		fmt.Printf("[everywhere]\n")
+		fmt.Printf("Repl. Factor: -1 | Allocations: [everywhere]\n")
 	} else {
 		var sortAlloc sort.StringSlice = obj.Allocations
 		sortAlloc.Sort()
-		fmt.Printf("%s\n", sortAlloc)
+		fmt.Printf("Repl. Factor: %d->%d | Allocations: %s\n",
```
ZenGround0: Maybe consider "%d--%d" without the arrow.  Im not used to seeing arrows in a range context.

ipfscluster_test.go
```diff
@@ -1095,6 +1143,12 @@ func TestClustersReplicationMinMaxRealloc(t *testing.T) {
 		t.Error("allocations should have changed")
 	}
 
+	lenSA := len(secondAllocations)
+	expected := minInt(nClusters-2, 4)
+	if lenSA != expected {
+		t.Errorf("Inssufficient reallocation, could have allocated to %d peers but instead only allocated to %d peers", expected, lenSA)
+	}
+
 	if len(secondAllocations) < 3 {
```
ZenGround0: It would make sense to reuse `lenSA` here

allocate.go
hsanjuan: I get lost here:

> However by the spec nPeers cluster peers should track C as no changes to the allocation set were needed.

* validAllocations is empty
* Wanted == nPeers
* Needed == something below nPeers but > 0
* currentValid == 0
* candidatesValid == nPeers - 1

Thus it will allocate normally and return a list of allocations of length (nPeers - 1)

In your example the pin mode is changing from everywhere to min-max, so it needs to fill-in the Allocations slice with whatever it can (which was empty). Maybe it's a spec problem? When it says "the allocations won't change if within margin" it means in the min/max realm. Switching from `everywhere` to min/max will fill-in or empty the allocations slice.

cluster.go
```diff
@@ -943,13 +945,8 @@ func (c *Cluster) Pins() []api.Pin {
...
 // returns an error if the given Cid is not part of the global state.
 func (c *Cluster) PinGet(h *cid.Cid) (api.Pin, error) {
-	cState, err := c.consensus.State()
-	if err != nil {
-		logger.Error(err)
-		return api.Pin{}, err
-	}
-	pin := cState.Get(h)
-	if pin.Cid == nil {
+	pin, ok := c.getCurrentPin(h)
```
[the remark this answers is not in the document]
hsanjuan: It was also the `	if curr, _ := c.getCurrentPin(pin.Cid); curr.Equals(pin) {` line. Before, if a re-pin just changed the allocation factors, but these did not trigger a change in the allocations, the new pin with new allocation factors (but same Allocations) was not commited. Now it commits the object if anything in it has changed.

docs/ipfs-cluster-guide.md
hsanjuan: `depending on the value of their metrics`

docs/ipfs-cluster-guide.md
hsanjuan: lol yeah took me a while to figure out what I meant

sharness/t0052-service-state-export.sh
hsanjuan: it didnt.. i think this is a leftover from trying stuff

cluster.go
```diff
@@ -943,13 +945,8 @@ func (c *Cluster) Pins() []api.Pin {
 // the item is successfully pinned. For that, use Status(). PinGet
```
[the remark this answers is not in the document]
hsanjuan: fixed

allocate.go
ZenGround0: I agree that changing the spec as suggested would also make this consistent

ZenGround0: I agree with everything you say until "Thus it will allocate normally", as from my perspective changing allocations when the min number is already met is abnormal.  I think the difference is I was taking "allocation" to mean assignment of a cid to a node and maybe you are interpreting it as "node's id listed in cid pin in shared state".

hsanjuan: In `pin everywhere` mode, a cid has no allocations. There is no min number concept there.

But I think I'm starting to get you. You mean that the peer that is down saw the pin as 'local' at the beginning,  but after the change, instead of keeping it when it comes up again, it will see it as remote and unpin it. And that in theory, it should still be pinning it because somehow it was part of the original set of allocations...

I think I'm happy living with that edge case, I'm not 100% sure if to call it bug as pinning everywhere or pinning with different replication factors are two separate modes (in my mind).

hsanjuan:
@ZenGround0 addessed your comment, rebased and regrouped commits, and fix a bug (last commit)

PR #294, opened 2018-01-16
Fix jenkins tests

hsanjuan:
I really want that jenkins works

License: MIT

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -1,1 +1,2 @@
-golang([test: "go test -loglevel CRITICAL -v ./..."])
+golang([test: "go test -v -timeout 20m ./..."])
+
--- a/cluster.go
+++ b/cluster.go
@@ -54,7 +54,7 @@ type Cluster struct {
 	readyB       bool
 	wg           sync.WaitGroup
 
-	// paMux sync.Mutex
+	paMux sync.Mutex
 }
 
 // NewCluster builds a new IPFS Cluster peer. It initializes a LibP2P host,
@@ -92,8 +92,8 @@ func NewCluster(
 	}
 
 	peerManager := newPeerManager(host)
-	peerManager.importAddresses(cfg.Peers)
-	peerManager.importAddresses(cfg.Bootstrap)
+	peerManager.importAddresses(cfg.Peers, false)
+	peerManager.importAddresses(cfg.Bootstrap, false)
 
 	c := &Cluster{
 		ctx:         ctx,
@@ -614,8 +614,8 @@ func (c *Cluster) PeerAdd(addr ma.Multiaddr) (api.ID, error) {
 	// starting 10 nodes on the same box for testing
 	// causes deadlock and a global lock here
 	// seems to help.
-	// c.paMux.Lock()
-	// defer c.paMux.Unlock()
+	c.paMux.Lock()
+	defer c.paMux.Unlock()
 	logger.Debugf("peerAdd called with %s", addr)
 	pid, decapAddr, err := multiaddrSplit(addr)
 	if err != nil {
@@ -761,7 +761,7 @@ func (c *Cluster) Join(addr ma.Multiaddr) error {
 	}
 
 	// Add peer to peerstore so we can talk to it
-	c.peerManager.addPeer(addr)
+	c.peerManager.addPeer(addr, true)
 
 	// Note that PeerAdd() on the remote peer will
 	// figure out what our real address is (obviously not
--- a/consensus/raft/consensus_test.go
+++ b/consensus/raft/consensus_test.go
@@ -40,7 +40,7 @@ func init() {
 	//logging.SetLogLevel("consensus", "DEBUG")
 }
 
-func makeTestingHost(t *testing.T, idn int) host.Host {
+func makeTestingHost(t *testing.T) host.Host {
 	priv, pub, _ := crypto.GenerateKeyPair(crypto.RSA, 2048)
 	pid, _ := peer.IDFromPublicKey(pub)
 
@@ -59,7 +59,7 @@ func makeTestingHost(t *testing.T, idn int) host.Host {
 }
 
 func testingConsensus(t *testing.T, idn int) *Consensus {
-	h := makeTestingHost(t, idn)
+	h := makeTestingHost(t)
 	st := mapstate.NewMapState()
 
 	cfg := &Config{}
--- a/consensus/raft/consensus_test.go
+++ b/consensus/raft/consensus_test.go
@@ -22,11 +22,17 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 )
 
-var p2pPort = 10000
-var p2pPortAlt = 11000
+func cleanRaft(idn int) {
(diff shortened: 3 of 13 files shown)


ci/Jenkinsfile
```diff
@@ -1 +1 @@
-golang([test: "go test -loglevel CRITICAL -v ./..."])
+golang([test: "go test -loglevel CRITICAL -v ."])
```
ZenGround0: Is it in scope to get all makefile commands running in this PR?  Would be dope to get `make docker` and `make test_sharness` in here with the suggested `node()` function

ipfscluster_test.go
```diff
@@ -163,26 +166,41 @@ func createClusters(t *testing.T) ([]*Cluster, []*test.IpfsMock) {
...
 	// }
+	// wg.Wait()
+
+	// ----------------------------------------------
 
+	// Alternative way of starting using bootstrap
+	// Start first node
+	clusters[0] = createCluster(t, cfgs[0], concfgs[0], apis[0], ipfss[0], states[0], trackers[0], mons[0], allocs[0], infs[0])
+	// Find out where it binded
+	bootstrapAddr, _ := ma.NewMultiaddr(fmt.Sprintf("%s/ipfs/%s", clusters[0].host.Addrs()[0], clusters[0].id.Pretty()))
```
ZenGround0: Out of curiosity what is going on with the bootstrap addr's form of <network addr>"/ipfs/"<cluster peer id>?  Asking because at a glance it seemed odd to have the ipfs string followed by an id that doesn't belong to an ipfs node.

ipfscluster_test.go
```diff
@@ -163,26 +166,41 @@ func createClusters(t *testing.T) ([]*Cluster, []*test.IpfsMock) {
...
+	// Use first node to bootstrap
+	for i := 1; i < nClusters; i++ {
```
ZenGround0: Is the bootstrap method required when moving to using port 0 for some reason?  Again this is just curiosity.

peer_manager.go
```diff
@@ -39,7 +43,10 @@ func (pm *peerManager) addPeer(addr ma.Multiaddr) error {
 			logger.Error(err)
 			return err
 		}
-		pm.importAddresses(resolvedAddrs)
+		pm.importAddresses(resolvedAddrs, connect)
+	}
+	if connect {
+		pm.host.Network().DialPeer(pm.ctx, pid)
```
ZenGround0: Stupid question: what is the purpose of Dialing the peer?

ci/Jenkinsfile
hsanjuan: just go tests. I am still unsure how the jenkins files would look like for that. Ideally, we can leave travis to run docker and sharness and jenkins to run go tests and be done with all fast.

ipfscluster_test.go
hsanjuan: it's a libp2p convention. `ipfs` means the libp2p protocol. There were plans to change it to `/p2p/<node_id>` but I'm not sure of the status...

ipfscluster_test.go
hsanjuan: yeah, cannot use the `Peers` array in the configuration because we have no idea where they are listening on before they are started. So need to start one peer and use it as bootstrap for the rest, in order to form the cluster. same as in kubernetes btw.

peer_manager.go
hsanjuan: Not so stupid, I'm not sure anymore.. I run into some problems with peers rejecting connection on certain ports when everyone bootstraps at the same time and played with the scenario of actively dialing peers first (setting up connection) rather than waiting to be contacted.

In the end I'm not sure it makes a difference. Most of my problems went away when put a lock around the PeerAdd so that not two peers can be joining in the same moment.

hsanjuan:
@ZenGround0 happy with this so far, I think we can merge (will use another PR to have travis and jenkins test different things when we can).

peer_manager_test.go
```diff
@@ -445,51 +454,54 @@ func TestClustersPeerJoinAllAtOnce(t *testing.T) {
...
-			t.Errorf("%s sees %d peers: %s", c.id, len(peers), peersIds)
-		}
-		pins := c.Pins()
-		if len(pins) != 1 || !pins[0].Cid.Equals(hash) {
-			t.Error("all peers should have pinned the cid")
-		}
-	}
-	runF(t, clusters, f2)
-}
+// This test fails a lot when re-use port is not available (MacOS, Windows)
```
ZenGround0: Do we have an issue to track debugging this?  Some things to specify and/or figure out.  Did this failure effect MacOS and Windows before moving to listening on port 0, or is it the combination of using these OSs with using port 0 that causes the failures.  What line of code throws the error and what is the expected message?  Is this an error primarily coming from the way the test is written or is it reflective of a problem with ipfs-cluster?

consensus/raft/consensus_test.go
```diff
@@ -22,22 +22,31 @@ import (
...
+	return cAddr
 }
 
 func init() {
 	_ = logging.LevelDebug
 	//logging.SetLogLevel("consensus", "DEBUG")
 }
 
-func makeTestingHost(t *testing.T, port int) host.Host {
+func makeTestingHost(t *testing.T, idn int) host.Host {
```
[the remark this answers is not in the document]
ZenGround0: Oh I guess it's used in the comments.  I'm not sure if the comment usage is correct though, idn shouldn't correspond to a generated port number right?

cluster.go
```diff
@@ -614,8 +614,8 @@ func (c *Cluster) PeerAdd(addr ma.Multiaddr) (api.ID, error) {
 	// starting 10 nodes on the same box for testing
```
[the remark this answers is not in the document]
hsanjuan: It's not a problem of thread safety. The issue here is that when adding a peer, you need to communicate to that peer the list of addresses of other peers in the cluster, so it knows how to reach them.

If you are adding 2 peers at the same time, they may end up not knowing about each-other. The lock essentially makes them queue and when they are processed one by one, you can be sure that they receive the correct list of addresses for everyone else.

consensus/raft/consensus_test.go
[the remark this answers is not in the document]
hsanjuan: I have removed the param, not needed. It was to pass in a fixed port to bind on.

ipfscluster_test.go
```diff
@@ -163,26 +170,41 @@ func createClusters(t *testing.T) ([]*Cluster, []*test.IpfsMock) {
 		clusterPeers[i] = addr
```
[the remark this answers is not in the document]
hsanjuan: I have commented these lines, in case we ever want to switch back.

peer_manager_test.go
hsanjuan: This is a problem with Windows and Mac not being able to use the listening port for outgoing connections, resulting in a new connection being created (with a new port) in order to bootstrap, but which cannot be used by other peers to reach that node. That, plus some operating system or Virtual Machine foo-barity. I'm happy to sacrify this test for the moment, but you're right, let's have an issue to keep in mind that it might be worth a look (and maybe enable it only in Linux): https://github.com/ipfs/ipfs-cluster/issues/306

PR #301, opened 2018-01-23
Fix #297: State unmarshal

### LINKED ISSUE #297: State migration problems
This issue is to document issues when performing state ugprades.

After deploying a new version of cluster in my test setup, it has correctly shown the error saying that the state needs an ugprade.

I have done `ipfs-cluster-state upgrade` correctly in all peers.

After starting the peers, then the following error has started to appear repeteadly:

```
14:20:54.765 ERROR libp2p-raf: error decoding snapshot: reflect: reflect.Value.SetString using unaddressable value fsm.go:126
14:20:54.765 ERROR       raft: Failed to restore snapshot: failed to restore snapshot 1000-442-1516371654704: reflect: reflect.Value.SetString using unaddressable value logging.go:86
```

I have the proceeded to stop the peers and run `state export` followed by `state import` in one node and `state clean` in every other node. This has resulted in a message about:

```
14:23:32.612 ERROR  consensus: Raft peers do not match cluster peers from the configuration. raft.go:159
14:23:32.612 ERROR  consensus: This likely indicates that this peer has left the cluster and/or raft.go:160
14:23:32.612 ERROR  consensus: has a dirty state. Clean the raft state for this peer raft.go:161
14:23:32.612 ERROR  consensus: (/home/hector/.ipfs-cluster/ipfs-cluster-data) raft.go:162
14:23:32.612 ERROR  consensus: bootstrap it to a working cluster. raft.go:163
```

(I don't know if it's intended that the peerset is reduced).

After that I have removed the peers from the configuration on that node and started it. Then pointed every other node to it (which had clean states) and bootstrapped them.

This has started the cluster, but now it seems only one node has a state, the others have not received the full state in the original node and thus are not tracking the pins (`ipfs-cluster-ctl status` shows only 1 node in the list for every pin and `ipfs-cluster-ctl pin ls` is empty in the nodes that had a clean state).

The snapshot metadata in peer with the imported state shows that the index is 1, which would be the same for a clean peer, so it may be that raft does not sent the snapshot to other peers because they think they are up to date (?)

hsanjuan:
It seems we have problems deserializing on top of an already
initialized state.

License: MIT

### DIFF
--- a/state/mapstate/map_state.go
+++ b/state/mapstate/map_state.go
@@ -12,6 +12,7 @@ import (
 
 	cid "github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log"
+
 	"github.com/ipfs/ipfs-cluster/api"
 )
 
@@ -150,10 +151,14 @@ func (st *MapState) Unmarshal(bs []byte) error {
 
 	// snapshot is up to date
 	buf := bytes.NewBuffer(bs[1:])
+	newState := MapState{}
 	dec := msgpack.Multicodec(msgpack.DefaultMsgpackHandle()).Decoder(buf)
-	err := dec.Decode(st)
+	err := dec.Decode(&newState)
 	if err != nil {
 		logger.Error(err)
 	}
+
+	st.PinMap = newState.PinMap
+	st.Version = newState.Version
 	return err
 }


hsanjuan:
This removes the errors in my tests for #297 on the migration part.

hsanjuan:
Btw @ZenGround0 , when you're testing about the state, use `ipfs-cluster-ctl pin ls` to check it corresponds to what you expect, not `ipfs-cluster-ctl status`. The output of status might be out of sync (it syncs to the shared state like every minute). I almost get bitten by this.

hsanjuan:
> Do you know why we get the reflector error on an existing state? It is the same one I ran into when we first made (un)marshaling the state's responsibility. As I recall it had to do with the interface not having a pointer value, but here st is a pointer.

I'm not sure, it probably has to do with replacing a map with another map without properly allocating it first because it existed already. It might even be an issue in msgpack code. All these coding/decoding is uses `reflect` and that is very tricky to use.



>    I'm having trouble understanding why this only occurs in the edge case you found. Doesn't unmarshalling into an existing state happen when ever a state is behind and needs to be caught up? If there is no upgrading/importing happening to the state first these errors don't happen, at least not in my checks today.

It's not very common that we send the full snapshot so that it needs to be unmarshaled on top of a existing state. When a node is behind by just a few entries, raft will send them separately rather than sending the full snapshot. In other cases (like starting up or catching up from scratch), the state is just empty, and it seems unmarshaling worked well in that case.

PR #303, opened 2018-01-24
Fix/297

ZenGround0:
This fixes the second problem discussed in issue #297, consensus peers are now directly copied from the config and imported snapshot replay is now forced on peers with empty state during bootstrap.

This has been tested out on a 3 peer cluster using the (export, _, _); (clean, clean, clean); (import, _,_); (start as single peer, bootstrap, bootstrap) order of operations.


hsanjuan:
Can you rebase on master?

Issue #300, opened 2018-01-23
ipfs-cluster-service [daemon] should have a flag to run migrations automatically


hsanjuan:
An `--upgrade` flag would automatically run any needed migrations on start. This makes it easy to perform an upgrade.

catilac:
should this be closed?

Issue #298, opened 2018-01-19
state version command


ZenGround0:
As distinct from cluster version this prints out the version of the state format used by a particular cluster version.  Good for users to assess whether a migration needs to occur and for sanity checking while debugging migration issues.

catilac:
should this be closed?

hsanjuan:
yep thanks

PR #307, opened 2018-01-26
Three small state features

### LINKED ISSUE #296: It should not be possible to update a current state (?)
Apparently we can run `state upgrade` even when the state is already up to date. Not sure, but it would seem more natural to say that 'the state doesn't need an upgrade'

ZenGround0:
Feat #300 -- `--upgrade` flag for automatically running upgrades when launching service
Fix #296   -- current state formats no longer cause upgrades.  No error but a warning
Feat #298 -- `ipfs-cluster-service state version` reports the mapstate version used by this cluster binary

### DIFF
--- a/ipfs-cluster-service/main.go
+++ b/ipfs-cluster-service/main.go
@@ -193,6 +193,10 @@ func main() {
 			Value: "disk-freespace",
 			Usage: "allocation strategy to use [disk-freespace,disk-reposize,numpin].",
 		},
+		cli.BoolFlag{
+			Name:  "upgrade, u",
+			Usage: "run necessary state migrations before starting cluster service",
+		},
 	}
 
 	app.Commands = []cli.Command{
@@ -424,6 +428,13 @@ func daemon(c *cli.Context) error {
 
 	// Load all the configurations
 	cfg, clusterCfg, apiCfg, ipfshttpCfg, consensusCfg, trackerCfg, monCfg, diskInfCfg, numpinInfCfg := makeConfigs()
+
+	// Run any migrations
+	if c.Bool("upgrade") {
+		err := upgrade()
+		checkErr("upgrading state", err)
+	}
+
 	// Execution lock
 	err := locker.lock()
 	checkErr("acquiring execution lock", err)
--- a/ipfs-cluster-service/main.go
+++ b/ipfs-cluster-service/main.go
@@ -193,10 +193,6 @@ func main() {
 			Value: "disk-freespace",
 			Usage: "allocation strategy to use [disk-freespace,disk-reposize,numpin].",
 		},
-		cli.BoolFlag{
-			Name:  "upgrade, u",
-			Usage: "run necessary state migrations before starting cluster service",
-		},
 	}
 
 	app.Commands = []cli.Command{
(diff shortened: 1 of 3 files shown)


ipfs-cluster-service/main.go
```diff
@@ -193,6 +193,10 @@ func main() {
 			Value: "disk-freespace",
 			Usage: "allocation strategy to use [disk-freespace,disk-reposize,numpin].",
 		},
+		cli.BoolFlag{
```
hsanjuan: I think I like this better as a flag specific for the `daemon` subcommand, rather than a global. `ipfs-cluster-service daemon --ugprade`

ipfs-cluster-service/state.go
```diff
@@ -32,38 +38,55 @@ func upgrade() error {
...
 	stateFromSnap := mapstate.NewMapState()
-	err = stateFromSnap.Migrate(r)
+	// duplicate reader to both check version and migrate
+	var buf bytes.Buffer
+	r2 := io.TeeReader(r, &buf)
+	raw, err := ioutil.ReadAll(r2)
+	if err != nil {
+		return nil, false, err
+	}
+	err = stateFromSnap.Unmarshal(raw)
```
hsanjuan: Something is not right here I think. If the state is completely different, `Unmarshal` might just fail because it doesn't know how to do it. That's why we did not unmarshal before and directly call `Migrate`. So it may make migrations fail.

Even if unmarshal does not fail, it might just not replace the `Version` field, which will be set to `mapstate.Version`, and give a false positive when doing `GetVersion() == mapstate.Version` below.

Looking around I see that we have this problem in the `mapstate.Migrate()` function too.

The straightfoward fix is that we make `first-byte-indicates-version` a requirement for all state marshaled-formats, and that we figure out the versions by just reading that first byte, and not by unmarshaling unknown formats on top of the current state types. A state helper (thinking `GetRawVersion(r io.Reader) int`) might help there. We have to remember to avoid loading full states into memory unless we need to (i.e. if we are going to fail or exist, because a version is not the expected, we shouldn't need to do `ReadAll` on the raw state, but just read the first byte.

ZenGround0: @hsanjuan
> The straightfoward fix is that we make first-byte-indicates-version a requirement for all state marshaled-formats, and that we figure out the versions by just reading that first byte, and not by unmarshaling unknown formats on top of the current state types.

My understanding is that this is what we are doing now.  I haven't been crystal clear about the requirement that first byte (and after a certain point varint as in multiformats) indicates version so thanks for bringing this up.  I should add this more clearly to docs.  As you can see from within Unmarshal however right now we are not unmarshaling unknown formats onto the state type without checking version, if the version doesn't match unmarshal won't decode. 

I agree with you that it is not ideal to read all the bytes before passing to Unmarshal because sometimes the first byte will suffice.  I am good with changing Unmarshal to take in a reader, consume the first byte to check for version, potentially quit if out of date, and continue reading otherwise.  Note that this is going to require some changes (easy I think) to libp2p-raft Restore, and DecodeSnapshot.

Am I understanding all of your concerns correctly?  Did I miss anything?  Are you ok with the suggested fix?  

Thanks!

hsanjuan: I totally missed that unmarshal checked the version byte before actually unmarshaling because it seemed we were doing that exact check afterwards. It's super weird that it doesn't return an error when it hasn't Unmarshaled anything because of a version mismatch. I think that's what confused me.

Let me have another look tomorrow with a fresh brain.

PR #310, opened 2018-01-29
Stream HTTP requests carrying multi-part files over the REST API for adding files via ipfs-cluster-ctl add

ZenGround0:
* parse files from fs to cmdkit.File objects
* stream over rest client to ipfs-cluster-service handler

### DIFF
--- a/api/rest/client/methods.go
+++ b/api/rest/client/methods.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 
 	cid "github.com/ipfs/go-cid"
+	"github.com/ipfs/go-ipfs-cmdkit/files"
 	peer "github.com/libp2p/go-libp2p-peer"
 	ma "github.com/multiformats/go-multiaddr"
 
@@ -171,4 +172,10 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
 	var graphS api.ConnectGraphSerial
 	err := c.do("GET", "/health/graph", nil, &graphS)
 	return graphS, err
+}
+
+// AddMultiFile adds new files to the ipfs cluster, importing and potentially
+// sharding underlying dags across the ipfs repos of multiple cluster peers
+func (c *Client) AddMultiFile(multiFileR *files.MultiFileReader) error {
+	return c.doStream("POST", "/files/add", multiFileR, nil)
 }
--- a/api/rest/client/methods.go
+++ b/api/rest/client/methods.go
@@ -177,5 +177,11 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
 // AddMultiFile adds new files to the ipfs cluster, importing and potentially
 // sharding underlying dags across the ipfs repos of multiple cluster peers
 func (c *Client) AddMultiFile(multiFileR *files.MultiFileReader) error {
-	return c.doStream("POST", "/files/add", multiFileR, nil)
+	headers := make(map[string]string)
+	headers["Content-Type"] = "multipart/form-data; boundary=" + multiFileR.Boundary()
+	err := c.doStream("POST", "/files/add", multiFileR, headers, nil)
+	if err != nil {
+		fmt.Printf("Error: %s\n", err.Error())
+	}
+	return err
 }
--- a/api/rest/client/methods.go
+++ b/api/rest/client/methods.go
@@ -179,9 +179,7 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
 func (c *Client) AddMultiFile(multiFileR *files.MultiFileReader) error {
 	headers := make(map[string]string)
 	headers["Content-Type"] = "multipart/form-data; boundary=" + multiFileR.Boundary()
-	err := c.doStream("POST", "/files/add", multiFileR, headers, nil)
-	if err != nil {
-		fmt.Printf("Error: %s\n", err.Error())
-	}
-	return err
+	return c.doStream("POST", "/allocations", multiFileR, headers, nil)
 }
+
+// Eventually an Add(io.Reader) method for adding raw readers as a multifile should be here.
--- a/api/rest/client/request.go
+++ b/api/rest/client/request.go
@@ -40,7 +40,7 @@ func (c *Client) doRequest(method, path string, body io.Reader) (*http.Response,
 
 // eventually we may want to trigger streaming with a boolean flag.
 // keeping functions seperate for now for development simplicity.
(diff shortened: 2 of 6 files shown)


ZenGround0:
This is about where I want this for a first round of code review.  `ipfs-cluster-ctl add` now streams files over the rest api and the cluster service reads the result and prints it to the terminal.  The add command takes a `-r` or `--recursive` flag like `ipfs add` which allows for adding directories recursively.

@hsanjuan clearly the final print out stage is just a placeholder for running the files through dex which is where i'll pick up in the next PR.  I duplicated a lot of code in api/rest/client/request.go to make streaming requests, and I copy pasted code from go-ipfs-cmds that couldn't be imported as far as I could tell.  I would love advice on how/if it's worthwhile to fix up these pieces.

Thanks!

ZenGround0:
@hsanjuan this is as far as I'll go in this PR (up do printing out the blocks output from importing the multipart streamed files).  I will remove dex dependency in a future PR.

api/rest/client/methods.go
```diff
@@ -172,3 +173,15 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
 	err := c.do("GET", "/health/graph", nil, &graphS)
 	return graphS, err
 }
+
+// AddMultiFile adds new files to the ipfs cluster, importing and potentially
+// sharding underlying dags across the ipfs repos of multiple cluster peers
+func (c *Client) AddMultiFile(multiFileR *files.MultiFileReader) error {
```
hsanjuan: It's good to have a method to send an already-prepared multifile thing, but it would expect also an `Add(io.Reader)` method that does a multifile upload. Don't know if now or later, but just to keep in mind.

api/rest/client/request.go
```diff
@@ -37,6 +38,44 @@ func (c *Client) doRequest(method, path string, body io.Reader) (*http.Response,
 	return c.client.Do(r)
 }
 
+// eventually we may want to trigger streaming with a boolean flag.
```
hsanjuan: are there any reasons that we'd want non-streaming file upload?

api/rest/client/request.go
```diff
@@ -37,6 +38,44 @@ func (c *Client) doRequest(method, path string, body io.Reader) (*http.Response,
...
+	if c.config.Username != "" {
+		r.SetBasicAuth(c.config.Username, c.config.Password)
+	}
+
+	for k, v := range headers {
+		r.Header.Set(k, v)
+	}
+
+	// Here are the streaming specific modifications
+	fmt.Printf("Here is the req before mods %v\n", r)
```
hsanjuan: remember to remove this or use logger.Debug

api/rest/restapi.go
```diff
@@ -273,6 +278,12 @@ func (api *API) routes() []route {
 			"/health/graph",
 			api.graphHandler,
 		},
+		{
+			"FilesAdd",
+			"POST",
+			"/files/add",
```
hsanjuan: call me crazy, but I think this might work `POST /allocations`. As it is effectively creating an Allocation (something to pin) and makes sense that it returns a PinInfo type (possibly)

api/rest/restapi.go
```diff
@@ -352,6 +363,56 @@ func (api *API) graphHandler(w http.ResponseWriter, r *http.Request) {
...
+		}
+
+		f = &files.MultipartFile{
+			Mediatype: mediatype,
+			Reader:    reader,
+		}
+
+	} else {
+		fmt.Printf("the mediatype: %s", mediatype)
+		sendAcceptedResponse(w, nil)
```
hsanjuan: mm is this a placeholder for non-streaming?

api/rest/restapi.go
```diff
@@ -352,6 +363,56 @@ func (api *API) graphHandler(w http.ResponseWriter, r *http.Request) {
...
+					fmt.Printf("\n")
+					break
+				}
+				fmt.Printf(string(buf[:n]))
+			}
+			fmt.Printf(string(buf[:n]))
+		}
+	*/
+
+	sendAcceptedResponse(w, err)
```
hsanjuan: eventually we should answer with the cid at least. just reminder

ipfs-cluster-ctl/main.go
```diff
@@ -151,6 +151,38 @@ requires authorization. implies --https, which you can disable with --force-http
...
+					Usage: "add directory paths recursively, default false",
+				},
+			},
+			Action: func(c *cli.Context) error {
+				paths := make([]string, c.NArg(), c.NArg())
+				for i, path := range c.Args() {
+					paths[i] = path
+				}
+				// Unclear if multiFileR is ready for streaming, but hypothesis is yes.
+				// Files are all opened but not read until they are sent.
```
hsanjuan: Haha I had the same question... it seems so.. since `SerialFile` opens as `ReaderFile`

api/rest/client/request.go
ZenGround0: My intention behind this comment was rather that we would have less code duplication if we only used `do` and `doRequest` functions (for file upload + all other requests) that take as an argument a streaming flag.

I am content leave the streaming specific functions for now though.

api/rest/restapi.go
ZenGround0: I like this idea too.  I'll do this (leaving off the pininfo for this PR)

api/rest/restapi.go
ZenGround0: No, thanks for pointing this out.  I should be returning an error here.  Right now the file handler requires that files are transmitted as multipart/form mediatype (we could probably be a little less restrictive in the future).

ipfs-cluster-ctl/main.go
ZenGround0: In case you were wondering I ended up confident that cmdkit.multiFileRs are compatible with streaming huge inputs.  On a similar note cmdkit.multiPartFiles are also compatible which isn't super intuitive when you first look at the module.  As far as I can tell multiPartFiles break expected interface functionality to stay streamable.  When you call `NextFile` on a streamed directory that was somewhere in the middle of a file it always returns nil, the only directory that gets out files from calls to `NextFile` is the root directory (the first part over the wire).  To stream we can keep a reference to the rootfile and keep calling `NextFile` on it to generate each file in succession.

api/rest/client/request.go
hsanjuan: ah ok i understand it now

ipfs-cluster-ctl/main.go
hsanjuan: yep

hsanjuan:
You can merge when you want

PR #316, opened 2018-02-09
Add files to calling cluster peer's ipfs daemon

ZenGround0:
Minimal code for adding files to ipfs via ipfs-cluster-ctl add.  Restapi now uses importToChannel and calls IPFS block/put on outgoing dag nodes' raw blocks' bytes.  Error checking and context cancellation could use some work.  Garbage collection is not considered.  As we are using block put the added data won't survive an ipfs gc.  

No golang tests yet, but I have tested by hand with complex directories (some paths up to 5 layers deep).  And things generally look good.  More extensive and precise tests to follow.

License: MIT

### DIFF
--- a/api/rest/client/request.go
+++ b/api/rest/client/request.go
@@ -59,8 +59,9 @@ func (c *Client) doStreamRequest(method, path string, body io.Reader, headers ma
 	}
 
 	// Here are the streaming specific modifications
-	r.ProtoMajor = 1
-	r.ProtoMinor = 1
+	// Using HTTP 2.0 to enable parallel reading req and writing resp
+	r.ProtoMajor = 2
+	r.ProtoMinor = 0
 	r.ContentLength = -1
 
 	return c.client.Do(r)
--- a/api/rest/restapi.go
+++ b/api/rest/restapi.go
@@ -386,7 +386,7 @@ func (api *API) addFileHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	outChan := make(chan *ipld.Node)
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(api.ctx)
 	go func() {
 		for nodePtr := range outChan {
 			node := *nodePtr
--- a/api/rest/restapi.go
+++ b/api/rest/restapi.go
@@ -386,6 +386,7 @@ func (api *API) addFileHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	outChan := make(chan *ipld.Node)
+	ctx, cancel := context.WithCancel(context.Background())
 	go func() {
 		for nodePtr := range outChan {
 			node := *nodePtr
@@ -398,15 +399,17 @@ func (api *API) addFileHandler(w http.ResponseWriter, r *http.Request) {
 				&pinS)
 
 			/* Verify that block put cid matches*/
-			if err != nil {
-				logger.Warning(err)
-			}
 			if node.String() != pinS.Cid { // node string is just cid string
 				logger.Warningf("mismatch. node cid: %s\nrpc cid: %s", node.String(), pinS.Cid)
 			}
+			if err != nil { // error will be triggered in importer
+				logger.Error(err)
+				cancel()
+				return
+			}
 		}
 	}()
-	err := dex.ImportToChannel(f, outChan, context.Background())
+	err := dex.ImportToChannel(f, outChan, ctx)
 	/*	buf := make([]byte, 256)
 		for {
 			file, err := f.NextFile()
--- a/api/rest/restapi.go
+++ b/api/rest/restapi.go
@@ -21,13 +21,10 @@ import (
 	rpc "github.com/hsanjuan/go-libp2p-gorpc"
 	cid "github.com/ipfs/go-cid"
 	"github.com/ipfs/go-ipfs-cmdkit/files"
-	ipld "github.com/ipfs/go-ipld-format"
 	logging "github.com/ipfs/go-log"
 	peer "github.com/libp2p/go-libp2p-peer"
 	ma "github.com/multiformats/go-multiaddr"
 	manet "github.com/multiformats/go-multiaddr-net"
-	//	"io"
-	//	"fmt"
 )
 
 var logger = logging.Logger("restapi")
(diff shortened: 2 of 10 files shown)


api/rest/restapi.go
```diff
@@ -384,8 +385,31 @@ func (api *API) addFileHandler(w http.ResponseWriter, r *http.Request) {
 		sendAcceptedResponse(w, errors.New("unsupported media type"))
 		return
 	}
+	outChan := make(chan *ipld.Node)
+	ctx, cancel := context.WithCancel(context.Background())
```
hsanjuan: you have an api.ctx which should be used to derive the new context

api/rest/restapi.go
```diff
@@ -384,8 +385,31 @@ func (api *API) addFileHandler(w http.ResponseWriter, r *http.Request) {
...
+				logger.Warningf("mismatch. node cid: %s\nrpc cid: %s", node.String(), pinS.Cid)
+			}
+			if err != nil { // error will be triggered in importer
+				logger.Error(err)
+				cancel()
+				return
+			}
+		}
+	}()
+	err := dex.ImportToChannel(f, outChan, ctx)
```
hsanjuan: maybe I greenlighted this somewhere, but `ImportToChannel` has a very weird signature. It seems more natural like:

`func ImportToChannel(ctx, f) <-chan` (and the returned channel should be closed if context is cancelled).

ZenGround0: This is handled in PR #318, I'm not fixing it here too because there is not point modifying the dex repo and gx publishing now that PR #318 has brought this code to cluster.  I'll rebase and resolve conflicts after PR #318 is merged.

api/rest/restapi.go
```diff
@@ -384,8 +385,31 @@ func (api *API) addFileHandler(w http.ResponseWriter, r *http.Request) {
 		sendAcceptedResponse(w, errors.New("unsupported media type"))
 		return
 	}
+	outChan := make(chan *ipld.Node)
+	ctx, cancel := context.WithCancel(api.ctx)
+	go func() {
```
hsanjuan: Ok, I had a really long comment here until I've figured out that ImportToChannel is probably hanging and that's why we need a a goroutine here to consume the channel.

With the new ImportToChannel signature, I think it should return a channel immediately. It's ImportToChannel implementation's problem to make sure the channel is correctly fed (by launching a goroutine there and managing it with the given context).

The job in this api handler is just to consume that channel (which can then happen in the main thread) and send the response when finished.

Also, I think we do the file streaming using http 1.1. From the go docs:

```
// Depending on the HTTP protocol version and the client, calling
        // Write or WriteHeader may prevent future reads on the
        // Request.Body. For HTTP/1.x requests, handlers should read any
        // needed request body data before writing the response. Once the
        // headers have been flushed (due to either an explicit Flusher.Flush
        // call or writing enough data to trigger a flush), the request body
        // may be unavailable. For HTTP/2 requests, the Go HTTP server permits
        // handlers to continue to read the request body while concurrently
        // writing the response. However, such behavior may not be supported
        // by all HTTP/2 clients. Handlers should read before writing if
        // possible to maximize compatibility.
```
We probably want to stream some sort of information in the response as we are adding (in the future). But it seems we should not be doing that with http1.1 (and use http/2 instead).

api/rest/restapi.go
```diff
@@ -384,8 +385,31 @@ func (api *API) addFileHandler(w http.ResponseWriter, r *http.Request) {
...
+		for nodePtr := range outChan {
```
hsanjuan: we want to stop reading from this channel if the api context is cancelled, so probably need to rewrite this with a `select`, to make sure we end the loop and return an error if we shutdown in the middle of this operation.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -876,3 +881,31 @@ func (ipfs *Connector) SwarmPeers() (api.SwarmPeers, error) {
 	}
 	return swarm, nil
 }
+
+// BlockPut triggers an ipfs block put on the given data, inserting the block
+// into the ipfs daemon's repo.
+func (ipfs *Connector) BlockPut(data []byte) (api.Pin, error) {
+	pin := api.Pin{}
+
+	r := ioutil.NopCloser(bytes.NewReader(data))
```
hsanjuan: It is not obvious to me why a NopCloser is used here?

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -876,3 +881,31 @@ func (ipfs *Connector) SwarmPeers() (api.SwarmPeers, error) {
 	}
 	return swarm, nil
 }
+
+// BlockPut triggers an ipfs block put on the given data, inserting the block
+// into the ipfs daemon's repo.
+func (ipfs *Connector) BlockPut(data []byte) (api.Pin, error) {
```
hsanjuan: This is abusing a little bit the `api.Pin` type to serialize a Cid. Perhaps we'll be fine just returning a string (unless you think we might need to return more information and thus need a wrapper type in the future).

test/ipfs_mock.go
```diff
@@ -231,6 +237,29 @@ func (m *IpfsMock) handler(w http.ResponseWriter, r *http.Request) {
...
+		}
+		part, err := mpr.NextPart()
+		if err != nil {
+			goto ERROR
+		}
+		data, err := ioutil.ReadAll(part)
+		if err != nil {
+			goto ERROR
+		}
+		c := cid.NewCidV0(u.Hash(data)).String()
```
hsanjuan: Can we use CidV1? V0 is deprecated

api/rest/restapi.go
ZenGround0: 100% with you.  Im planning to fix this (no goroutine) in PR #318 after this is merged so i can rebase the changes in.  The reason im not doing it now is that I dont want to make changes to /publish the defunct dex repo which id need to do to get the correct ImportTo signature.

👍 on moving to http 2 as well

ipfsconn/ipfshttp/ipfshttp.go
ZenGround0: NewReaderFile takes in io.ReadCloser https://github.com/ipfs/go-ipfs-cmdkit/blob/master/files/readerfile.go#L20

hsanjuan: Got it now, thanks

api/rest/restapi.go
hsanjuan: ok!

api/rest/client/request.go
```diff
@@ -59,8 +59,8 @@ func (c *Client) doStreamRequest(method, path string, body io.Reader, headers ma
 	}
 
 	// Here are the streaming specific modifications
-	r.ProtoMajor = 1
-	r.ProtoMinor = 1
+	r.ProtoMajor = 2
```
hsanjuan: Let's leave a comment on why we do this and how this is important or not

api/rest/restapi.go
```diff
@@ -385,32 +383,38 @@ func (api *API) addFileHandler(w http.ResponseWriter, r *http.Request) {
...
+			"IPFSBlockPut",
+			node.RawData(),
+			&hash)
+		/* Verify that block put cid matches*/
+		if node.String() != hash { // node string is just cid string
+			logger.Warningf("mismatch. node cid: %s\nrpc cid: %s", node.String(), hash)
 		}
-	*/
+		if err != nil {
+			// TODO: think about how to handle this better.
```
hsanjuan: Thanks for leaving TODOs everywhere :)

PR #317, opened 2018-02-09
Sharding Component -- WIP

ZenGround0:
@hsanjuan There appears to be a fair amount of design work required here before we can implement so I would love to get your thoughts.
Is an AddNode function repeatedly called over RPC by the process reading from the importer's out-channel close to what you had in mind?  Is the AddNode pseudo code more or less what you were thinking?  Any ideas on how best to approach the getting storage limits alongside peer allocations or on framing the sharder's safety properties?

Feel free to push things directly to this branch.


ZenGround0:
@hsanjuan single node tests on my machine check out.  I will be testing and debugging in a live cluster this week.  This is ready for your review when you have the time.  Thank you!

shard/sharder.go
```diff
@@ -0,0 +1,352 @@
...
+	rpc "github.com/hsanjuan/go-libp2p-gorpc"
+	cid "github.com/ipfs/go-cid"
+	cbor "github.com/ipfs/go-ipld-cbor"
+	ipld "github.com/ipfs/go-ipld-format"
+	logging "github.com/ipfs/go-log"
+	peer "github.com/libp2p/go-libp2p-peer"
+	mh "github.com/multiformats/go-multihash"
+)
+
+var logger = logging.Logger("shard")
```
hsanjuan: `sharder`

shard/sharder.go
```diff
@@ -0,0 +1,352 @@
...
+// NewSharder returns a new sharder for use by an ipfs-cluster.  In the future
+// this may take in a shard-config
+func NewSharder(cfg *Config) (*Sharder, error) {
+	logger.Debugf("The alloc size provided: %d", cfg.AllocSize)
+	return &Sharder{allocSize: cfg.AllocSize,
+		currentShard: make(map[string]*cid.Cid),
+	}, nil
+}
+
+func (s *Sharder) unInit() bool {
```
hsanjuan: I think I know what this does, but a comment would help

shard/sharder.go
```diff
@@ -0,0 +1,352 @@
...
+  sharder responding to information received from informer calls?
+
+2. Shard node (cluster dag) format
+   In the current go-ipfs landscape downloading resolvers for non-standard
+   ipld formats (i.e. everything but unixfs, and I think dag-cbor) is
+   a difficult requirement to ask from nodes in the cluster.
+
+   For this reason I think we should use one of ipfs's standard ipld
+   formats.  Is dag-cbor ok to use? Probably preferrable over unixfs
+   protobuf if we can help it.
```
hsanjuan: yes, let's try get things working with dag-cbor, but make it in a way that we can switch the implementation easily.

shard/sharder.go
```diff
@@ -0,0 +1,352 @@
...
+		ReplicationFactorMin: 1,
+		ReplicationFactorMax: 1,
+	}.ToSerial()
+
+	return s.rpcClient.Call(targetPeer, "Cluster", "Pin", pinS, &struct{}{})
+}
+
+/* Open questions
+
+1. getAssignment()
```
hsanjuan: I might be underthinking this.

1. For every new shard, get a node from the allocator (if using the `reposize` strategy, the one reporting less reposize).
2. IPFS Block put all the blocks to that node
3. IPFS Block put the shard root node to that node
4. ipfs-cluster-pin the root node like you would pin any other CID (so it will be pinned recursively in several cluster peers)

The last step will pin with the desired replication factor just like any other CID, and will tell the necessary number of ipfs daemons to pin that tree. IPFS daemons which don't have the content will fetch it from the ipfs peer that has it.

You don't need to think how much space you take from a peer. It's going to be the shard size because each shard is written to a single place at first. For our prototype, we don't even have to worry about making this first peer the same as the ones in which the shard is pinned. You may well write to one peer and then the `Pin` operation may allocate them to others because it has no context that that shard is already written in one of them (it should be easy to address this later and manually allocate that peer).

Don't worry now about shards not distributed evenly etc. we can address not-up-to-date metrics, more clever allocation assignments later.

The final root node of all shards needs to be pinned, perhaps everywhere, but not recursively.

shard/sharder.go
```diff
@@ -0,0 +1,352 @@
...
+   a difficult requirement to ask from nodes in the cluster.
+
+   For this reason I think we should use one of ipfs's standard ipld
+   formats.  Is dag-cbor ok to use? Probably preferrable over unixfs
+   protobuf if we can help it.
+
+3. We'll need to think about serializing nodes across rpc.  This may be
+   simple I need to dig into this.  Depending on the node format that
+   files are imported to we might need multiple api types for multiple
+   formats.  We might be able to guarantee one format from our importer
```
hsanjuan: From the moment the data is chunked and we serialize it to pass it to the sharder, I don't see there is need to deserialize it anymore. We just need to carry the right metadata and block-put the binary slice we are carrying. I think this is how you do it now for the most part.

hsanjuan:
Let me know if you have questions @ZenGround0 , we can jump in a call if I went too fast over something.

api/rest/restapi.go
```diff
@@ -362,6 +363,93 @@ func (api *API) graphHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	// TODO: when complete this call should answer with a cid
+	// or better yet a pin-info describing the allocations
+	// of the resulting allocation
+	//
+	// Before returning this we will need to trigger a pin
+	// probably doing the same thing as Pin if its not a sharding call
+	//
+	// This is actually a bit tricky because right now we are not
```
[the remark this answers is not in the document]
ZenGround0: Ah, I think this is probably right but I'll need to check the add implementation to get a guarantee.  Good catch.

api/rest/restapi.go
```diff
@@ -362,6 +363,93 @@ func (api *API) graphHandler(w http.ResponseWriter, r *http.Request) {
 	sendResponse(w, err, graph)
 }
 
+func (api *API) addFile(ctx context.Context, outChan <-chan *ipld.Node, w http.ResponseWriter) error {
+	for nodePtr := range outChan {
```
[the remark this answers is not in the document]
ZenGround0: It should cause a channel closure in the importer.ToChannel goroutine but I'll incorporate context cancel handling here.  I'll do it now so even if I need to do a full rewrite I'll remember this.

api/rest/restapi.go
```diff
@@ -362,6 +363,93 @@ func (api *API) graphHandler(w http.ResponseWriter, r *http.Request) {
...
+		node := *nodePtr
+		nodeS := types.NodeSerial{
+			CidS: node.Cid().String(),
+			Data: node.RawData(),
+		}
+		err := api.rpcClient.Call("",
+			"Cluster",
+			"ShardAddNode",
+			nodeS,
+			&struct{}{})
```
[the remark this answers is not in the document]
ZenGround0: Yeah I want to do this soon.  At the very least just the root hash of the final cluster DAG node (file root if adding locally) so that the user gets some closure.  The silent finish is kind of unnerving right now (what happened, did everything work correctly?).

api/types.go
```diff
@@ -609,6 +617,48 @@ func (pins PinSerial) ToPin() Pin {
...
+// to be decoded into a block using NewBlockWithCid which can further be
+// decoded into an ipld node
+type NodeSerial struct {
+	CidS string
+	Data []byte
+}
+
+// ToIPLDNode converts a node serial to an ipld DAG node.  Data must be
+// encoded as Raw or DagCBOR (v0 or v1) for this to work
+func (nS NodeSerial) ToIPLDNode() (ipld.Node, error) {
```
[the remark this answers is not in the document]
ZenGround0: Hmm, I did notice this and put off thinking about breaking the pattern.  In light of your observation that we do not need the node itself, just the data and metadata, returning an error is much more sensible to avoid (ToIPLDNode will go away). Moving to this approach I am happy keeping the old error-less policy here for now.

Also happy to discuss changing the policy however.  I don't feel too strongly either way.

shard/sharder.go
ZenGround0: Correct me if I am wrong but as far as I can tell casting between ipld formats is not generally easy (or always possible).
However As it stands the dag-cbor impl is not too hard to swap out with dag-protobuf.  We could simply grab the links from the cbor shard node with `Links()` make a new protonode and call the Public facing merkledag/node:`AddRawLink`.  We could also construct `ipld.Links` directly from the cids in the sharder state and do this.

hsanjuan: I didn't mean that we can switch and have compatibility with what we had done before, I meant that we can switch for testing (from scratch) and then settle on what works best for the final solution.

shard/sharder.go
ZenGround0: Well said.  I wrote this before implementing and I converged pretty much onto the implementation you describe in 1 to 4.  

I understand that the space you take from a peer is just the shard size, in this section I was focusing (in retrospect looks like too much) on the details of the shard size potentially adapting during the sharding process.  Looking back most of the concerns here are orthogonal to sharding, i.e. how can nodes specify exactly what the storage they offer up can be used for: do they have the bandwidth for big shards, which nodes are make sense to accept allocs from for bandwidth or region/topology considerations, do they want to maintain extra freespace for more small pins in the future?

Ignoring the above as out of scope and taking on the simplifying assumption that all nodes in a cluster agree on one shard size, and it's a configured value, most of the allocation complexity goes away.  The only thing left to handle is what happens in the edge case all nodes together have the space to fit file data but no node has the space for the configured shard size (need fewer smaller shards).

Eventually it may make sense to stop making this assumption and allow nodes to specify richer inputs into the allocation process.  This will probably happen to address the broader questions asked in paragraph 2.  In the same way the reposize metric gets used for allocs perhaps the allocation configs of individual nodes will broadcast their allocation preferences, including their max shard size.  One weird thing about the current assumption for sharding is that as it stands right now it is ONLY the sharder config of the node doing the `ipfs-cluster-ctl add` that decides the shard size.  Therefore a node could specify a shard size of 5 MB in its config and end up with 100MB shards pinned to its repo.  

Also thanks for pointing out that we'd need to change pinning to include a particular PID to make the pin happen on the inserted node.  Right now we are doing step 4 but I forgot that this doesn't necessarily mean that the node with the blocks will do the pinning.  In fact there might be something weird happening where the node given the blocks is less likely to get allocated the pin because the freespace in its repo is artificially lowered by the inserted blocks.  Also supporting higher repl factors becomes pretty hard in the case of each node specifying their own shard size because then the same file may need to be sharded into different sized pieces to pack it into different nodes.

Whew sorry about the essay, I think we are on the same page, just wanted to get those thoughts out in the open.

shard/sharder.go
ZenGround0: Thanks for noticing this

ZenGround0:
Thanks @hsanjuan, this is all clear.  In particular thanks for the observation that we should be able to support multiple file shardings at the same time.  Moving forward with this my strategy is to introduce an `InitShardSession` RPC that returns a shard id and then call sharding RPCs with this id.

hsanjuan:
@ZenGround0 you don't need an `InitShardSession`. Just let AddNode receive a ShardId, and if you don't have it yet, init a new session.

ZenGround0:
@hsanjuan, ah thanks!

shard/sharder.go
hsanjuan: Forgot to submit this yesterday:

> Therefore a node could specify a shard size of 5 MB in its config and end up with 100MB shards pinned to its repo

Not only here, in many places we make the assumptions that cluster configurations are similar and make sense for the cluster. Shard size should be a request option anyway (for a start), and whoever triggers a sharding-add should be careful to not do it in a way that is not going to fit the cluster (I think that's a respectable expectation for our first approach, and that these considerations can be addressed in later rounds).

> Another thought: supporting higher repl factors becomes pretty hard in the case of each node specifying their own shard size because then the same file may need to be sharded into different sized pieces to pack it into different nodes. 

Having different shards sizes depending on available space might be a really good idea with things like FEC. But yeah, it's weird on a >1 replication factor. We will have to think of this, but probably the diverse-cluster with nodes of many types with completely different repo-capacity and almost full is not the main target usecase for our first iteration.

api/rest/restapi.go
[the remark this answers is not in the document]
ZenGround0: Latest commits address in large part, though concurrent request reading and response streaming is still elusive.

ZenGround0:
@hsanjuan this is ready for another round of review.  A quick summary of what's changed since you last checked in:
1. Your comments, mostly about unrefined rpc api data types and methods and lack of support for multiple sharding sessions, have been addressed
2. The sharding component now has a little mock rpc testing framework + 3 unit tests.  I can think of at least 2 more (force 2 shard nodes + ensure estimated size clusterdag size is correct) and would welcome more ideas 
3. clusterDAG shard-nodes and root-nodes are now automatically broken into multiple nodes if they exceed the IPFS block limit
4. ipfs-cluster-add has improved a lot.  The command now passes in most of the `ipfs add` flags and the importer endpoint was adjusted to take these in and reproduce ipfs's behavior.  The restapi file handler handles errors and context timeouts more systematically.  Finally the api streams output info back to the client.  This required making a cluster-specific fileAdder so we're also ready to make use of the modules you've been extracting.  While the api can stream the response back as soon as it gets output info, it turns out this breaks request reading even using HTTP 2.0 on the client.  Any ideas on how to fix this are very welcome.

Thanks in advance

api/types.go
```diff
@@ -609,6 +609,59 @@ func (pins PinSerial) ToPin() Pin {
...
+type ShardNodeSerial struct {
+	Cid  string
+	Data []byte
+	Size uint64
+	ID   string
+}
+
+// BlockWithFormat specifies a block of data and how it should be
+// formatted.  An empty format string will be interpreted as "v0".
+type BlockWithFormat struct {
```
hsanjuan: Is this mergeable with NodeSerial? Meaning, that we could use NodeSerial for passing a block and its data around?

I'm even wondering if worth merging with `ShardNodeSerial` too.

api/types.go
```diff
@@ -609,6 +609,59 @@ func (pins PinSerial) ToPin() Pin {
...
+type AllocateInfo struct {
+	Cid        string
+	Current    map[peer.ID]Metric
+	Candidates map[peer.ID]Metric
+}
+
+// GetCid decodes the cid string within AllocateInfo.  If the cid string is ""
+// then GetCid returns nil
+func (aI *AllocateInfo) GetCid() *cid.Cid {
+	if aI.Cid == "" {
```
hsanjuan: Might as well skip this check, `cid.Decode` likely returns nil for the empty string...

ipfs-cluster-ctl/main.go
```diff
@@ -166,6 +166,42 @@ is also TODO
...
+				},
+				cli.BoolFlag{
+					Name:  "quiet, q",
+					Usage: "write minimal output",
+				},
+				cli.BoolFlag{
+					Name:  "silent",
+					Usage: "write no output",
+				},
+				cli.BoolFlag{
```
hsanjuan: string flag. Mention options in help

ipfs-cluster-ctl/main.go
```diff
@@ -166,6 +166,42 @@ is also TODO
...
+					Name:  "trickle, t",
+					Usage: "use trickle dag layout for dag generation",
+				},
+				cli.StringFlag{
+					Name:  "chunker, s",
```
hsanjuan: mention which options are valid in help text

ipfscluster.go
```diff
@@ -152,6 +152,15 @@ type PinAllocator interface {
...
 }
 
+// Sharder controls the aggregation of ipfs file nodes into shards.  File
+// shards are grouped together and referenced by a cluster DAG node and
+// distributed among the cluster
+type Sharder interface {
+	Component
+	AddNode(size uint64, data []byte, c string, id string) error
+	Finalize(id string) error
+}
```
hsanjuan: This is a nice simple interface for a component that does a lot of stuff :+1:

ipld-importer/add.go
```diff
@@ -0,0 +1,439 @@
...
+type ignoreFileError struct {
+	fileName string
+}
+
+func (e *ignoreFileError) Error() string {
+	return fmt.Sprintf("%s is an ignored file", e.fileName)
+}
+
+// NewAdder Returns a new Adder used for a file add operation.
+func NewAdder(ctx context.Context, p pin.Pinner, bs bstore.GCBlockstore, ds ipld.DAGService) (*Adder, error) {
```
hsanjuan: Is this a simplified copy from the ipfs adder?

rpc_api.go
```diff
@@ -347,6 +366,20 @@ func (rpcapi *RPCAPI) ConsensusPeers(in struct{}, out *[]peer.ID) error {
 	return err
 }
 
+/*
+   Sharder methods
+*/
+
+// ShardAddNode runs Sharder.AddNode(node).
+func (rpcapi *RPCAPI) ShardAddNode(in api.ShardNodeSerial, out *struct{}) error {
```
hsanjuan: Let's call it `SharderAddNode` for consistency

rpc_api.go
```diff
@@ -347,6 +366,20 @@ func (rpcapi *RPCAPI) ConsensusPeers(in struct{}, out *[]peer.ID) error {
...
+	return rpcapi.c.sharder.AddNode(in.Size, in.Data, in.Cid, in.ID)
+}
+
+// ShardFinalize runs Sharder.Finalize().
+func (rpcapi *RPCAPI) ShardFinalize(in string, out *struct{}) error {
```
hsanjuan: `SharderFinalize`

shard/sharder.go
```diff
@@ -0,0 +1,421 @@
+package shard
```
hsanjuan: I think since we refer to this as the Sharder we should put it in a `sharder` folder and name the package `sharder` too.

ZenGround0:
@hsanjuan Yes apologies for the size, this one got bigger than I realized before I remembered to request review.  I recognize it was a mistake to wait this long and certainly am not trying to make a habit of it.  That said the packaging of distinct work into distinct commits should help this decompose during review, though I could also try to repackage it into multiple PRs if you'd prefer.

Moving to IRC to schedule a call

api/rest/restapi.go
```diff
@@ -362,60 +371,207 @@ func (api *API) graphHandler(w http.ResponseWriter, r *http.Request) {
...
+		// TODO: Unlike earlier approaches this is actually systematic
+		// and handles all possible errors.  This approach is still
+		// deferring the investigation into what error handling
+		// approaches give the best UX however.
+		select {
+		// Ensure we terminate reading from importer after cancellation
+		// but do not block
+		case <-ctx.Done():
+			err = errors.New("context timeout terminated add")
+			return enc.Encode(types.Error{Code: 1, Message: err.Error()})
```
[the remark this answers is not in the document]
ZenGround0: Because I built this with the goal of eventually streaming the response concurrent with reading the request I didn't want to assume that the header could be written at the time the error surfaced which is why errors are streamed in the body.

Maybe I am overcomplicating but from my reading of golang's http modules the client needs to know when to close the response body so I built a little application protocol over HTTP where the client knows to close on receipt of a JSON object encoded like this with an error code (1 or 2).  

If we decide against trying to get concurrent responding and request reading in the near future I can just change this to set an error response header

hsanjuan:
@ZenGround0 it's ok, let's not complicate it more :)

api/rest/restapi.go
```diff
@@ -362,60 +371,207 @@ func (api *API) graphHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+	return err
+}
+
+func (api *API) finishLocalAdd(args map[string]string) error {
+	// TODO: pin the root node from here, rpc.Call(pin, args["cid"])
+	return nil
+}
+
+func (api *API) consumeShardAdd(args map[string]string, outObj *types.NodeSerial) error {
```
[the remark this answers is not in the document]
ZenGround0: I don't think so, though on a related not the consumeImport approach is maybe not go-ish or cluster-ish enough for your tastes.  

To avoid writing the same code twice I parametrized the code that reads from importer channels with `consume` and `finish` functions.  Subsequent calls need to communicate data to future calls, but the data they need to communicate is depended on the `consume` `finish` function implementation.  Specifically the ShardAdd functions need to communicate the shardID across calls and the `consumeLocalAdd` function should communicate the root Cid so that `finishLocalAdd` can pin the root of the file.  

I could change to a specific struct that has both shardID and rootCid fields but it doesn't seem to be a better fit than a map[string]string from my point of view.

api/rest/client/methods.go
```diff
@@ -176,10 +176,15 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
 
 // AddMultiFile adds new files to the ipfs cluster, importing and potentially
 // sharding underlying dags across the ipfs repos of multiple cluster peers
-func (c *Client) AddMultiFile(multiFileR *files.MultiFileReader) error {
+func (c *Client) AddMultiFile(multiFileR *files.MultiFileReader, shard bool,
+	quiet bool, silent bool, trickle bool, chunker string,
```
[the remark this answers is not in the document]
ZenGround0: I like it.  We should bring this over to go-ipfs too 😄

api/types.go
ZenGround0: For sure, if you're ok with having unrelated zero valued fields along for the ride.  It's a trade off between lots of specific types perfectly tuned to match function call parameters and fewer types that are used in a bunch of different ways.  I think I've been doing the former lately when defining api types but I'm happy to move towards the latter.

ipfs-cluster-ctl/main.go
ZenGround0: Making trickle a bool means that `ipfs-cluster-ctl add` has subtlely different flag semantics than go-ipfs.  I agree a string flag for layout makes more sense in the long run but it comes with the potential downside of making our UI a little more annoying to use for people switching between our CLI and go-ipfs.  I'm fine with doing this just wanted to make the consequences of our choice here explicit.

ipld-importer/add.go
ZenGround0: Yes!  I cut away all of the GC and pinning stuff.  It also has some slight tweaks in the code that outputs printable objects for reporting progress.  I'll make a document level comment describing the source and the tweaks from the original.

api/types.go
ZenGround0: Thanks, indeed it does

api/rest/restapi.go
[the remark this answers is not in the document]
hsanjuan: I see, it's not pretty. I might take a stab at that code and see if I can improve something

api/types.go
hsanjuan: I think, since we only use this to pass data internally, and this data is essentially the same (a block slice with metadata that gets filled in as it goes down the pipeline), we can use the same type for everything. It's much simpler to see that the same type leaving the rest api is the one arriving to the ipfs connector. Essentially it hasn't been transformed, just carries more info in it.

hsanjuan:
Also, let's call the `ipld-importer` -> `importer`.

api/types.go
ZenGround0: Got it.  Do you think ShardNodeSerial should also be merged into NodeSerial?

hsanjuan: Yes

ZenGround0:
@hsanjuan Let me know how this is looking.  I think it is most of the way there.

hsanjuan:
I still don't like the `map[string]string` though I like the idea. But let's merge and take care of that later.

hsanjuan:
@ZenGround0 tests fail. Have a look and feel free to merge if unrelated.

PR #318, opened 2018-02-09
Bring importer interface to cluster

ZenGround0:
dependence on dex replaced with ipld-importer subpackage
slight improvement to importer tests and formatting
gx deps update: new raft, ipld format and ipfs added

Last remaining task in this direction is to implement a cluster adder.  Looking into this more I'm leaning towards including this in the go-ipld-importer repo for a few reasons, mainly keeping too many go-ipfs imports (mfs merkledag) out of cluster.  Also if we reuse go-ipfs's adder we have the benefit of always maintaining the same importing process and avoid forking away from future changes to importing semantics (though hopefully those are rare enough to not be a big issue).  Imho the file adder would ideally be extracted from coreunix, brought into the importers repo and grouped under the same level of abstraction as go-ipfs/tar/format.go where cluster, go-ipfs and friends could all use it as the unixfs specific importer.

Totally understand that this many not be feasible right now however.  Excited to sync up about go-ipld-importer soon


ipld-importer/import.go
```diff
@@ -0,0 +1,86 @@
...
+	// ignore errors caused by printing-blockstore Get not finding blocks
+	if err != nil && !strings.Contains(err.Error(), "dagservice: block not found") {
+		return err
+	}
+	return nil
+}
+
+// ToChannel imports file to ipfs ipld nodes, outputting nodes on the
+// provided channel
+func ToChannel(f files.File, outChan chan<- *ipld.Node, ctx context.Context) error {
```
hsanjuan: See my comment in the other review. Are there any reason for passing in the channel rather than having this return a newly created read-only channel?

ipld-importer/outgoing-dagservice.go
```diff
@@ -0,0 +1,74 @@
...
+	"errors"
+	"fmt"
+
+	cid "github.com/ipfs/go-cid"
+	ipld "github.com/ipfs/go-ipld-format"
+)
+
+var errUninit = errors.New("DAGService output channel uninitialized")
+
+// outDAGService will "add" a node by printing it.  pDAGService cannot Get nodes
```
hsanjuan: comment does not correspond to type

ipld-importer/import.go
ZenGround0: Not really.  From my  un-refined/new-to-go perspective it seemed clearer to have `ToChannel` block until the file processing was finished and have the receiver spawn a go-routine, rather than have the `ToChannel` function invisibly spawn a go-routine and return immediately.

Happy to change it if this is your preference.

ZenGround0: Just read the other PR, I'll change this.

ipld-importer/outgoing-dagservice.go
```diff
@@ -0,0 +1,76 @@
...
+
+	cid "github.com/ipfs/go-cid"
+	ipld "github.com/ipfs/go-ipld-format"
+)
+
+var errUninit = errors.New("DAGService output channel uninitialized")
+
+// outDAGService will "add" a node by sending through the outChan
+type outDAGService struct {
+	membership map[string]bool
```
hsanjuan: better use `map[string]struct{}`

ipld-importer/import.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return err
+	}
+	fileAdder.Pin = false
+
+	go func() {
+		defer close(outChan)
+		// add all files under the root, as in ipfs
+		for {
+			file, err := f.NextFile()
```
hsanjuan: this needs to finish when the ctx has been cancelled.

ipld-importer/import.go
```diff
@@ -0,0 +1,89 @@
...
+
+// ToPrint is a first step towards a streaming importer.  Verify that we
+// can hijack the blockstore abstraction to redirect blocks as they arrive
+// Closely follows go-ipfs/core/commands/add.go: Run func
+func ToPrint(f files.File) error {
+	dserv := &pDAGService{
+		membership: make(map[string]bool),
+	}
+
+	ctx := context.Background() // using background for now, should upgrade later
```
hsanjuan: maybe de-duplicate the code in this function since it's the same in both

ipld-importer/printing-dagservice.go
```diff
@@ -0,0 +1,63 @@
+package importer
```
hsanjuan: same comments as for the other dag service. you might want to use a `logger` rather than printLn. Or maybe in the future it is just worth to log as debug messages in the other dagService and simply delete this (I understand this is just for easier testing now).

hsanjuan:
What I mean is, let's delay the `writing our own adder` a little bit until things work well using the ipfs one.

ipld-importer/import.go
```diff
@@ -0,0 +1,89 @@
...
+				return err
+			}
+
+			if err := fileAdder.AddFile(file); err != nil {
+				return err
+			}
+		}
+
+		_, err = fileAdder.Finalize()
+		if !strings.Contains(err.Error(), "dagservice: block not found") {
```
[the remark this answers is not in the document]
ZenGround0: Because `Finalize()` finishes with a call to Get the root of the imported file from the dagservice.  This call only happens at the very end, and we expect it to error because the OutgoingDagservice always errors on Get.  Furthermore block gets are only called at this point in the code so we are not in danger of ignoring a false negative.

If we implement our own adder with its own `Finalize()` method this will be trivial to clean up.

ipld-importer/import.go
ZenGround0: It will, albeit in an unclear way because the `fileAdder.AddFile` is context aware and the outDAGService's `Add` function selects on the context's Done function alongside waiting for a consumer of the output channel, raising an error and exiting the loop at line 79.

Would you prefer that we check the context explicitly in the `ToChannel` function? (select ctx.Done(); default fileAdder.AddFile())

hsanjuan: Yeah I'd prefer that

ipld-importer/import.go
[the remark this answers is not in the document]
hsanjuan: Does adder.Finalize() do anything useful for us (outputDirs()?)? Most of the code in it is about mfs, and we don't even care about the result. Can we live without calling it?

ZenGround0:
Your question about folders relates to the comment about mfs.  I agree we don't need pinning or gc but we do need mfs as functions from that package are used to write unixfs formated directories to the dagservice.  

I think I see what you are asking about folders having "depth", but correct me if I'm wrong.  The way cmdkit multipart files work, only the very root directory, i.e. the part at the head of the stream can read files within that directory with a call to `NextFile()`.  Calling `NextFile()` on a subdirectory results in an EOF error.  All subfiles of the top level directory, including those nested within other directories are eventually pulled from the stream with a call to `NextFile()` from the top level root.  This seems to be a slight abuse of the interface but it makes our lives simpler.  Both ipfs/core/commands/add.go and the new cluster import.go iterate over the children of the top level root part and call `fileAdder.Addfile`.  When the input file is a multipart directory, this does not recursively cause the addition of the dirrectory's subfiles (https://github.com/ipfs/go-ipfs/blob/master/core/coreunix/add.go#L535) 

From this perspective I would say no, directories do not have depth, with the exception of the first one.  
Again let me know if I misunderstood your question.

ipld-importer/import.go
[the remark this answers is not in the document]
ZenGround0: No we cannot live without calling it.  As you seem to have guessed it outputs directories (I figured this out with the printing dagservice).  From my recent comment in the main thread we actually need mfs for this very reason.

hsanjuan:
Thanks for the detailed explanation @ZenGround0 . Ok, let's leave that funny error handling there for the moment until we can adjust the adder to our needs.

ipld-importer/import.go
ZenGround0: With upcoming changes (inserted select statement, not returning errors from goroutine) deduplication looks like it is not quite worth the cost.  Also I am leaning towards your suggestion of removing the printing-dagservice and the ToPrint call later on in favor of debug logging in to Channel.

PR #325, opened 2018-03-01
Priority Allocations + Non-recursive pinning  -- PROPOSAL

ZenGround0:
@hsanjuan this is my first stab at incorporating priority allocations into cluster to support the sharding use case of explicitly asking cluster to pin a cid on a certain peer so that blocks aren't moving around unnecessarily.  I have held off writing tests because I have a feeling that this design may not meet your standards.  A better solution would probably re-evaluate cluster's allocation approach more generally.  I went the quickest route and modified `pin` `allocate` `obtainAllocations` and the Allocator interface which you might find too intrusive to support this small edge case.  Interested to hear your thoughts.

This PR also tracks work extending the api.types Pin object to have a Recursive field to support non-recursive pinning.  This is to support non-recursive pinning of the clusterDAG root metadata nodes on all peers.  This shouldn't be too controversial a change and a commit will follow soon finishing this off.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -523,13 +523,16 @@ type Pin struct {
 	Allocations          []peer.ID
 	ReplicationFactorMin int
 	ReplicationFactorMax int
+	Recursive            bool
 }
 
-// PinCid is a shorcut to create a Pin only with a Cid.
+// PinCid is a shorcut to create a Pin only with a Cid.  Default is for pin to
+// be recursive
 func PinCid(c *cid.Cid) Pin {
 	return Pin{
 		Cid:         c,
 		Allocations: []peer.ID{},
+		Recursive:   true,
 	}
 }
 
@@ -540,6 +543,7 @@ type PinSerial struct {
 	Allocations          []string `json:"allocations"`
 	ReplicationFactorMin int      `json:"replication_factor_min"`
 	ReplicationFactorMax int      `json:"replication_factor_max"`
+	Recursive            bool     `json:"recursive"`
 }
 
 // ToSerial converts a Pin to PinSerial.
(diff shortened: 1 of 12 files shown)


api/types.go
```diff
@@ -523,13 +523,16 @@ type Pin struct {
 	Allocations          []peer.ID
 	ReplicationFactorMin int
 	ReplicationFactorMax int
+	Recursive            bool
```
hsanjuan: This is going to require a migration

hsanjuan:
@ZenGround0 generally OK with respects to code.

There are two features in this PR though, I'd prefer them separate. The pin part requires migration because otherwise all pins are going to default to `Recursive: false`.

In the allocator part I'm not sure this is the best way. I understand you want to force allocations when you pin something (to ensure that those are the allocations that the allocator gave you at some point earlier in the pipeline and to which you have written blocks).

* Do you need that your priority list of allocations comes back ordered by metric?
* Does it makes sense that your priority list is not considered if metrics have expired?
* Are you planning to pass one peer or several in this priority list?

What I'm trying to figure out is if we need to involve the allocator or we are just trying to manually set allocations (in which case we can set the `Allocations` field of the Pin object manually and use that to signal that we set one or several allocations by hand.

api/types.go
lanzafame: @hsanjuan What determines when a migration is required?

ZenGround0: @lanzafame if the state format changes so that updated versions of cluster will treat old state data in an incorrect way or be unable to directly deserialize old state files into the new format when loading then a migration is needed.  In this case a migration is needed because the state format (e.g. [MapState](https://github.com/ipfs/ipfs-cluster/blob/master/state/mapstate/map_state.go#L28)) uses PinSerial and the because the default value of booleans in golang is false while the default value of `ipfs pin`'s recursive argument is true.

hsanjuan:
We discussed in a call and I'm ok with this change. Also if merged to master. After the migration is added.

ZenGround0:
Putting this into two separate PRs to master.

PR #329, opened 2018-03-06
ipfs-cluster-ctl cmd: add no-status flag to pin add/rm

lanzafame:
@hsanjuan @ZenGround0 Thoughts on the naming of the shortflag `-nostat`?

### DIFF
--- a/ipfs-cluster-ctl/main.go
+++ b/ipfs-cluster-ctl/main.go
@@ -268,6 +268,10 @@ peers should pin this content.
 							Value: "",
 							Usage: "Sets a name for this pin",
 						},
+						cli.BoolFlag{
+							Name:  "no-status, ns",
+							Usage: "Prevents fetching pin status after pinning (faster, quieter)",
+						},
 					},
 					Action: func(c *cli.Context) error {
 						cidStr := c.Args().First()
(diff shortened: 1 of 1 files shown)


ipfs-cluster-ctl/main.go
```diff
@@ -305,7 +311,12 @@ in the cluster. The CID should disappear from the list offered by "pin ls",
 although unpinning operations in the cluster may take longer or fail.
 `,
 					ArgsUsage: "<CID>",
-					Flags:     []cli.Flag{},
+					Flags: []cli.Flag{
+						cli.BoolFlag{
+							Name:  "no-status, nostat",
+							Usage: "Prevents status reporting for faster unpinning",
```
hsanjuan: same

ipfs-cluster-ctl/main.go
```diff
@@ -305,7 +311,12 @@ in the cluster. The CID should disappear from the list offered by "pin ls",
 although unpinning operations in the cluster may take longer or fail.
 `,
 					ArgsUsage: "<CID>",
-					Flags:     []cli.Flag{},
+					Flags: []cli.Flag{
+						cli.BoolFlag{
+							Name:  "no-status, nostat",
+							Usage: "Prevents fetching pin status after pinning (faster, quieter)",
```
hsanjuan: `unpinning`

ZenGround0:
My personal preference is to omit a shortflag for this flag as `no-status` and `nostat` are close in pronunciation and length that having both feels a bit redundant to me.  I don't feel too strongly about this though so feel free to keep it as is if you disagree.

lanzafame:
@ZenGround0 I don't disagree, though considering the purpose of this is to cut 1s from a pin/unpin operation, it would make sense to me to have a very short 'short' flag, i.e. `-ns`. @hsanjuan Thoughts?

ZenGround0:
`-ns` SGTM

hsanjuan:
yeah I prefer `-ns` too

lanzafame:
Cool, I have just pushed the `-ns` short flag change up.

hsanjuan:
Fixes #329

PR #334, opened 2018-03-09
rest/client: double default timeout

### LINKED ISSUE #323: ipfs-cluster-ctl times out very quickly on longer operations
`Post http://127.0.0.1:9094/pins/recover?local=true: net/http: request canceled (Client.Timeout ele awaiting headers)`

My guess is that the Client is configure with some default timeouts.

* It should be configurable (as part of the rest/client config)
* There should be a flag to generally set timeouts in ipfs-cluster-ctl
* Default should be way higher than right now

lanzafame:
No logic behind how much I increased the value by. @hsanjuan suggestions welcome?

Resolves #323.

### DIFF
--- a/api/rest/client/client.go
+++ b/api/rest/client/client.go
@@ -14,7 +14,7 @@ import (
 
 // Configuration defaults
 var (
-	DefaultTimeout  = 60 * time.Second
+	DefaultTimeout  = 120 * time.Second
 	DefaultAPIAddr  = "/ip4/127.0.0.1/tcp/9094"
 	DefaultLogLevel = "info"
 )
--- a/ipfs-cluster-ctl/main.go
+++ b/ipfs-cluster-ctl/main.go
@@ -26,7 +26,7 @@ const Version = "0.3.4"
 
 var (
 	defaultHost     = "/ip4/127.0.0.1/tcp/9094"
-	defaultTimeout  = 60
+	defaultTimeout  = 120
 	defaultUsername = ""
 	defaultPassword = ""
 )


hsanjuan:
OK. Let's raise the default value for the flag in ipfs-cluster-ctl to two minutes too.  I don't think I realized it was configurable already when I wrote this.

lanzafame:
@hsanjuan trying to diagnose why the tests are failing and it appears that it is because the tests are hitting the 20m timeout. Running it locally currently, with a longer timeout to determine how much the timeout needs to be or which tests need speeding up.

lanzafame:
So with the timeout set to 30m, all the tests finished but failed the `TestClustersReplicationMinMaxRealloc` test failed, except when I ran that test by itself it passed, even looping over that test several times, it still passed (I thought it might be a slightly flaky test but it appears to be fine). Currently, running the full test suite again, just in case it produces a different result. Though all these tests seem unrelated to what this PR changes.

lanzafame:
I just ran `TestClustersReplicationMinMaxRealloc` and `TestClustersPin` (the test that failed on the second attempt at running the full test suite) individually with `-race` and it would appear both of them suffer from multiple race conditions, which might be the reason for the sporadic failures.

hsanjuan:
@lanzafame there are probably many race conditions around, many of which come from cluster.

But sometimes tests fail/hang for other reasons (libp2p heisenbugs). I'm slowly working to fix races but for the moment don't use `-race` (for your own sanity). For the rest, sometimes you'll have to retrigger tests until they pass. We'll revisit randomly failing tests when we upgrade to latest libp2p (and go).

PR #341, opened 2018-03-09
ToPin call and priority pinning

ZenGround0:
@hsanjuan the allocation functionality from #325 separated out from the recursive pin modification and set to merge into master.  Let me know if this needs further work.


cluster.go
```diff
@@ -983,15 +983,27 @@ func (c *Cluster) PinGet(h *cid.Cid) (api.Pin, error) {
...
+}
+
+// PinTo makes the cluster Pin a Cid as in Pin.  PinTo's argument already
+// contains a set of peers that should perform the pin.  If the max repl factor
+// is less than the size of the specified peerset then peers are chosen from
+// this set in allocation order.  If the min repl factor is greater than the
+// specified peerset then peers are allocated as in Pin.  PinTo is best effort.
+// If the peers selected for pinning are unavailable then PinTo will attempt to
+// allocate other peers and will not register an error.
+func (c *Cluster) PinTo(pin api.Pin) error {
```
hsanjuan: Let's have `Pin` do this directly?

allocate.go
```diff
@@ -75,7 +78,8 @@ func (c *Cluster) allocate(hash *cid.Cid, rplMin, rplMax int, blacklist []peer.I
 		rplMin,
 		rplMax,
 		currentMetrics,
-		candidatesMetrics)
+		candidatesMetrics,
+		priorityMetrics)
```
lanzafame: I am not sure if there is a current convention in ipfs codebases, but I usually follow the same pattern as a multiline struct initialization for multiline function calls, see below:
```go
s := &stuff{
    value1,
    value2,
}

s.DoStuff(
   param1,
   param2,
)
```
This way you don't have a two line addition from `git`'s perspective when you have a single line addition. Also, all parameters have the same indentation level. I often find my eyes skip over the first parameter if it is on the same line as the function call, and then all other params listed on consecutive lines.

Please feel free to disagree :)

allocate.go
```diff
@@ -136,7 +140,8 @@ func allocationError(hash *cid.Cid, needed, wanted int, candidatesValid []peer.I
 func (c *Cluster) obtainAllocations(
 	hash *cid.Cid,
 	rplMin, rplMax int,
-	currentValidMetrics, candidatesMetrics map[peer.ID]api.Metric) ([]peer.ID, error) {
+	currentValidMetrics, candidatesMetrics map[peer.ID]api.Metric,
```
lanzafame: Would it be worth considering a metric struct that contains the current, candidate and priority metrics maps? Or possibly even better (though I haven't read through all this code, so take this proposal with a grain of salt) would be to define a custom key type, e.g. 
```go
type MetricType int

const (
    Unknown MetricType = iota
    CurrentValid
    Candidates
    Priority
)

type MetricKey struct {
    peer.ID
    Type MetricType
}
```

usage:
```go
 func (c *Cluster) obtainAllocations(hash *cid.Cid, 	rplMin, rplMax int, metrics map[MetricKey]api.Metric) ([]peer.ID, error) {...}
```

@ZenGround0 Would love to hear your thoughts on this?

allocate.go
ZenGround0: I agree with this.  @hsanjuan is this already the convention or if not is it something you agree on making a convention in cluster code?

hsanjuan: I'm happy with trying to maintain this convention, but I'm also ok as long as code is `gofmt`'ed.

allocate.go
ZenGround0: @lanzafame While we could certainly do this I'm not sure I see the advantages of these two approaches over the existing approach.  Let me know if I'm overlooking advantages to these approaches in the paragraphs below.

A struct wrapping the dictionaries would make function arguments more concise but right now I don't necessarily see this as worth the trouble of introducing a new type.

As I understand it the content of the metric struct is independent of the cluster's consideration of this metric as priority, candidate or currently valid during allocation.  The later categorization depends on the context of the call to `Cluster.allocate`, i.e. the blacklist, priority-list, pinset of the cluster, whereas the metric struct is only in charge of reporting the state of the world at a certain peer to other peers in the cluster and can, potentially anyway, be used for more than just allocation decisions.  For this reason I don't think the MetricType typed field quite belongs in the metric.

@hsanjuan let me know if I'm missing something here.

Issue #338, opened 2018-03-09
ipfs-cluster-ctl pin/add pin/rm should have a --wait option


hsanjuan:
Just like we provide a --no-status option, there should be a --wait option which doesn't return until the pin has reached PINNED status or has been fully unpinned.

It should do this by regulary querying `status <cid>` on the pin and waiting for all PINNING to either be PINNED or PIN_ERROR.

Before exiting, it should print the final status.

I think this WaitForPin/Unpin function should be provided as an utility/method in `rest/client`, and re-used by ipfs-cluster-ctl. Open for proposals but `func client.WaitFor(targetStatus, checkInterval) <- channel Status` is one option.

lanzafame:
If `func client.WaitFor(targetStatus, checkInterval) <- channel Status` is returning a `Status` every `checkInterval` on the returned channel, what is the requirement for the `targetStatus` parameter?

lanzafame:
scratch that, I understand why now :)

hsanjuan:
Yeah, target would be Pinned/Unpinned :)

lanzafame:
![](https://github.trello.services/images/mini-trello-icon.png) [#338](https://trello.com/c/po1OdAfA/2-338)

hsanjuan:
#348

PR #348, opened 2018-03-14
rest/client: Add WaitFor utility function

lanzafame:
Putting this up early, as there are a few things I am unsure about.

1) api/rest/client/methods.go:193
```diff
+				// TODO(ajl): check I can override api.TrackerStatusBug here for
+				// error reporting purposes or return an error chan,
+				// not sure which is the best option.
+				statusCh <- api.TrackerStatusBug
```

2) The use of `goto` to simplify the for loop wait logic. See api/rest/client/methods.go:201 and api/rest/client/methods.go:207.

3) The `pin` test works but not in current configuration combined with the `unpin` test. 

Thoughts and suggestions would be greatly appreciated 👍

### DIFF
--- a/api/rest/client/methods.go
+++ b/api/rest/client/methods.go
@@ -2,9 +2,11 @@ package client
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"time"
 
 	cid "github.com/ipfs/go-cid"
 	peer "github.com/libp2p/go-libp2p-peer"
@@ -60,12 +62,16 @@ func (c *Client) Pin(ci *cid.Cid, replicationFactorMin, replicationFactorMax int
 	escName := url.QueryEscape(name)
 	err := c.do(
 		"POST",
-		fmt.Sprintf("/pins/%s?replication_factor_min=%d&replication_factor_max=%d&name=%s",
+		fmt.Sprintf(
+			"/pins/%s?replication_factor_min=%d&replication_factor_max=%d&name=%s",
 			ci.String(),
 			replicationFactorMin,
 			replicationFactorMax,
-			escName),
-		nil, nil)
+			escName,
+		),
+		nil,
+		nil,
+	)
 	return err
 }
 
@@ -171,4 +177,130 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
 	var graphS api.ConnectGraphSerial
 	err := c.do("GET", "/health/graph", nil, &graphS)
 	return graphS, err
+}
+
+// WaitFor is a utility function that allows for a caller to
+// wait for a paticular status for a CID. It returns a channel
+// upon which the caller can wait for the targetStatus.
+func (c *Client) WaitFor(ctx context.Context, fp StatusFilterParams) (api.GlobalPinInfo, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	sf := newStatusFilter()
+
+	go sf.pollStatus(ctx, c, fp)
+	go sf.filter(ctx, fp)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return api.GlobalPinInfo{}, ctx.Err()
+		case err := <-sf.Err:
+			return api.GlobalPinInfo{}, err
+		case <-sf.Done:
+			// drain status channel and return last status on channel
+			for {
+				if status, ok := <-sf.Out; !ok {
+					return status, nil
+				}
+			}
+		}
+	}
+}
+
+// StatusFilterParams contains the parameters required
+// to filter a stream of status results.
+type StatusFilterParams struct {
+	*cid.Cid
+	Local     bool
+	Target    api.TrackerStatus
+	CheckFreq time.Duration
+}
+
+type statusFilter struct {
+	In, Out chan api.GlobalPinInfo
+	Done    chan struct{}
+	Err     chan error
+}
+
+func newStatusFilter() *statusFilter {
+	return &statusFilter{
+		In:   make(chan api.GlobalPinInfo),
+		Out:  make(chan api.GlobalPinInfo),
+		Done: make(chan struct{}),
+		Err:  make(chan error),
+	}
+}
+
+func (sf *statusFilter) filter(ctx context.Context, fp StatusFilterParams) {
+	defer close(sf.Done)
+	defer close(sf.Out)
+
+	for {
+		select {
+		case <-ctx.Done():
+			sf.Err <- ctx.Err()
+			return
+		case gblPinInfo, more := <-sf.In:
+			if !more {
+				return
+			}
+			ok, err := statusReached(fp.Target, gblPinInfo)
+			if err != nil {
+				sf.Err <- err
+				return
+			}
+
+			if !ok {
+				sf.Out <- gblPinInfo
+				continue
+			}
+
+			sf.Out <- gblPinInfo
+			return
+		}
+	}
+}
+
+func (sf *statusFilter) pollStatus(ctx context.Context, c *Client, fp StatusFilterParams) {
+	ticker := time.NewTicker(fp.CheckFreq)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			sf.Err <- ctx.Err()
+			return
+		case <-ticker.C:
+			gblPinInfo, err := c.Status(fp.Cid, fp.Local)
+			if err != nil {
+				sf.Err <- err
+				return
+			}
+			logger.Debugf("pollStatus: status: %#v", gblPinInfo)
+			sf.In <- gblPinInfo
+		case <-sf.Done:
+			close(sf.In)
+			return
+		}
+	}
+}
+
+func statusReached(target api.TrackerStatus, gblPinInfo api.GlobalPinInfo) (bool, error) {
+	for _, pinInfo := range gblPinInfo.PeerMap {
+		switch pinInfo.Status {
+		case target:
+			continue
+		case api.TrackerStatusBug, api.TrackerStatusClusterError, api.TrackerStatusPinError, api.TrackerStatusUnpinError:
+			return false, fmt.Errorf("error has occurred while attempting to reach status: %s", target.String())
+		case api.TrackerStatusRemote:
+			if target == api.TrackerStatusPinned {
+				continue // to next pinInfo
+			}
+			return false, nil
+		default:
+			return false, nil
+		}
+	}
+	return true, nil
 }
--- a/ipfs-cluster-ctl/main.go
+++ b/ipfs-cluster-ctl/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -301,6 +302,15 @@ peers should pin this content.
 							Name:  "no-status, ns",
 							Usage: "Prevents fetching pin status after pinning (faster, quieter)",
 						},
+						cli.BoolFlag{
+							Name:  "wait, w",
+							Usage: "Wait for all nodes to report a status of pinned before returning",
+						},
+						cli.DurationFlag{
+							Name:  "wait-timeout, wt",
+							Value: 0,
+							Usage: "How long to --wait (in seconds), default is indefinitely",
+						},
 					},
 					Action: func(c *cli.Context) error {
 						cidStr := c.Args().First()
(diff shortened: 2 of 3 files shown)


api/rest/client/methods.go
```diff
@@ -172,3 +177,37 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
...
+				logger.Errorf("failed to get pin status: %v", err)
+				// TODO(ajl): check I can override api.TrackerStatusBug here for
+				// error reporting purposes or return an error chan,
+				// not sure which is the best option.
+				statusCh <- api.TrackerStatusBug
+			}
+			for _, pinInfo := range gblPinInfo.PeerMap {
+				if target != pinInfo.Status {
+					// one of the peers has a status that doesn't match target
+					goto wait
```
ZenGround0: I am not opposed to a well-placed goto but typically I like to see labels in ALL CAPS

api/rest/client/methods.go
```diff
@@ -172,3 +177,37 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
...
+
+	go func(statusCh chan api.TrackerStatus) {
+		for {
+			gblPinInfo, err := c.Status(ci, local)
+			if err != nil {
+				// TODO(ajl): check how logging is done in ipfs-cluster
+				logger.Errorf("failed to get pin status: %v", err)
+				// TODO(ajl): check I can override api.TrackerStatusBug here for
+				// error reporting purposes or return an error chan,
+				// not sure which is the best option.
```
ZenGround0: This is a good question.  @hsanjuan

api/rest/client/methods_test.go
```diff
@@ -224,3 +227,64 @@ func TestGetConnectGraph(t *testing.T) {
...
+				err = c.Unpin(ci)
+			}
+
+			t.Errorf("failed to perform setup %s: %v", tt.name, err)
+
+			statusCh := c.WaitFor(tt.args.ci, tt.args.local, tt.args.target, tt.args.checkFreq)
+			for {
+				select {
+				case stat := <-statusCh:
+					log.Printf("%#v\n", stat)
```
ZenGround0: I'm guessing this is just incomplete now because you are debugging but eventually it will be good to assert tt.target == stat here.

api/rest/client/methods.go
hsanjuan: IMHO, return an error chan

api/rest/client/methods.go
```diff
@@ -172,3 +177,37 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
...
+				statusCh <- api.TrackerStatusBug
+			}
+			for _, pinInfo := range gblPinInfo.PeerMap {
+				if target != pinInfo.Status {
+					// one of the peers has a status that doesn't match target
+					goto wait
+				}
+			}
+			// NOTE: this code shouldn't be reached unless all peers have the target status
+			statusCh <- target
```
hsanjuan: Should we send all statuses to the channel? Might be handy. But discard if previous hasn't been consumed.

api/rest/client/methods.go
```diff
@@ -172,3 +177,37 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
...
+			if err != nil {
+				// TODO(ajl): check how logging is done in ipfs-cluster
+				logger.Errorf("failed to get pin status: %v", err)
+				// TODO(ajl): check I can override api.TrackerStatusBug here for
+				// error reporting purposes or return an error chan,
+				// not sure which is the best option.
+				statusCh <- api.TrackerStatusBug
+			}
+			for _, pinInfo := range gblPinInfo.PeerMap {
+				if target != pinInfo.Status {
```
hsanjuan: You have to account for `Remote` status. Essentially, what you want to check for pinned is that "all pins have status PINNED except those with REMOTE or ERROR.". Perhaps worth returning an error if you see a pin in error state and aborting the whole thing. If no one is in ERROR, it amounts to checking that things are PINNED (excluding REMOTE), or UNPINNED.

api/rest/client/methods.go
hsanjuan: There is only one place jumping to the `wait` label, so `goto` not justified imho. Sleep and `continue`.

api/rest/client/methods.go
```diff
@@ -172,3 +177,37 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
...
+					// one of the peers has a status that doesn't match target
+					goto wait
+				}
+			}
+			// NOTE: this code shouldn't be reached unless all peers have the target status
+			statusCh <- target
+			return
+		wait:
+			time.Sleep(checkFreq)
```
hsanjuan: Maybe better to use a `time.Ticker`. Though you'd have to stop it during the request and restart after. Whatever ends up being cleaner.

api/rest/client/methods.go
ZenGround0: @lanzafame you may already know but you can continue the outer loop by placing a label there and executing `continue <LABEL>`

lanzafame:
[#338](https://trello.com/c/po1OdAfA/2-338)

api/rest/client/methods.go
```diff
@@ -172,3 +178,72 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
 	err := c.do("GET", "/health/graph", nil, &graphS)
 	return graphS, err
 }
+
+// WaitFor is a utility function that allows for a caller to
+// wait for a paticular status for a CID. It returns a channel
+// upon which the caller can wait for the targetStatus.
+func (c *Client) WaitFor(ctx context.Context, ci *cid.Cid, local bool, target api.TrackerStatus, checkFreq time.Duration) (<-chan api.TrackerStatus, <-chan error) {
```
hsanjuan: I realize I wasn't super clear. The returned channel should provide `GlobalPinInfo` objects (this is the result of the `ctl status` calls and thus the confusion).

api/rest/client/methods.go
```diff
@@ -172,3 +178,72 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
...
+
+	go func() {
+		ticker := time.NewTicker(checkFreq)
+		defer ticker.Stop()
+
+	OUTER:
+		for {
+			select {
+			case <-ctx.Done():
+				return
```
hsanjuan: put the ctx.Err() in the error channel

api/rest/client/methods.go
```diff
@@ -172,3 +178,72 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
...
 	return graphS, err
 }
+
+// WaitFor is a utility function that allows for a caller to
+// wait for a paticular status for a CID. It returns a channel
+// upon which the caller can wait for the targetStatus.
+func (c *Client) WaitFor(ctx context.Context, ci *cid.Cid, local bool, target api.TrackerStatus, checkFreq time.Duration) (<-chan api.TrackerStatus, <-chan error) {
+	statusCh := make(chan api.TrackerStatus)
+	errCh := make(chan error)
+
```
hsanjuan: Make sure channels are closed when the goroutine stops (because of error, because of completion)

api/rest/client/methods.go
```diff
@@ -172,3 +178,72 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
...
+		ticker := time.NewTicker(checkFreq)
+		defer ticker.Stop()
+
+	OUTER:
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				gblPinInfo, err := c.Status(ci, local)
```
hsanjuan: Use another function for this `reachedTarget(target, gblPinInfo) (bool, error)`.

Codeclimate will justly complain about too much complexity. Also, saves you using `goto` to outer loops.

hsanjuan: Also, you can put the glbPinInfo in the channel as you get it.

If target is reached, you close the channel are you're done.

api/rest/client/methods.go
```diff
@@ -172,3 +178,72 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
...
+				return
+			}
+		}
+	}()
+
+	return statusCh, errCh
+}
+
+// WaitForPinnedStatus ...
+func (c *Client) WaitForPinnedStatus(ctx context.Context, ci *cid.Cid, local bool, checkFreq time.Duration) (bool, error) {
```
hsanjuan: Taking into account all of the above, success here would mean:

* no errors received on error channel and no more items to read on it (because it's been closed)
* no more items to read from gPinInfo channel (because it's been closed too)

We might want to not return on errors, and just put them on the error channel and let the user decide whether to abort (via context) or keep waiting despite errors (maybe they're been addressed on the side). Let's leave that for next review round though.

ipfs-cluster-ctl/main.go
```diff
@@ -272,6 +273,15 @@ peers should pin this content.
...
 							Usage: "Prevents fetching pin status after pinning (faster, quieter)",
 						},
+						cli.BoolFlag{
+							Name:  "wait, w",
+							Usage: "Waits for all nodes to report a status of pinned before returning",
+						},
+						cli.DurationFlag{
+							Name:  "wait-timeout, wt",
+							Value: 0,
+							Usage: "Specify how long (in seconds) you are willing to wait for pinned status (default is indefinitely)",
```
hsanjuan: We have a sharness tests that limits the width of help lines to 120 chars (when doing `ipfs-cluster-ctl -h`). This probably hits it.

ipfs-cluster-ctl/main.go
```diff
@@ -326,11 +382,17 @@ although unpinning operations in the cluster may take longer or fail.
...
-						if !c.Bool("no-status") {
+
+						// handle boolean flags
+						switch {
+						case !c.Bool("no-status"):
 							time.Sleep(1000 * time.Millisecond)
 							resp, cerr := globalClient.Status(ci, false)
 							formatResponse(c, resp, cerr)
+						case c.Bool("wait"):
+							// TODO(ajl): globalClient.WaitForUnpinnedStatus
```
hsanjuan: don't forget

api/rest/client/methods.go
```diff
@@ -172,3 +178,106 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
...
+			return
+		case <-ticker.C:
+			gblPinInfo, err := c.Status(fp.Cid, fp.Local)
+			if err != nil {
+				sf.Err <- err
+				return
+			}
+			sf.In <- gblPinInfo
+		case <-sf.Done:
+			close(sf.In)
```
hsanjuan: you probably want to return here? Otherwise next tick will panic

api/rest/client/methods.go
```diff
@@ -172,3 +178,106 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
...
+		}
+	}
+}
+
+// WaitFor is a utility function that allows for a caller to
+// wait for a paticular status for a CID. It returns a channel
+// upon which the caller can wait for the targetStatus.
+func (c *Client) WaitFor(ctx context.Context, sf *StatusFilter, fp FilterParams) {
```
hsanjuan: Let WaitFor consume the channels and return `(globalPinInfo, error)`, where the first is the last status received and the second is an error (either from status or context).

api/rest/client/methods.go
```diff
@@ -172,3 +178,106 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
...
+	Target    api.TrackerStatus
+	CheckFreq time.Duration
+}
+
+func (sf *StatusFilter) Filter(ctx context.Context, fp FilterParams) {
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				sf.Err <- ctx.Err()
```
hsanjuan: If channels are exported they should be closed when the sender is not going to send anything anymore. Whoever uses them may just be hanging. I am leaning towards not exporting any channel (nor the `StatusFilter`), and just have `WaitFor` provide the functionality we need (it is easy to reverse this decision later if needed).

api/rest/client/methods.go
hsanjuan: Also, in that case, I think this could also work by just receiving FilterParams and creating the StatusFilter. Which in turns means that StatusFilter does not even need to be exported.

ipfs-cluster-ctl/main.go
```diff
@@ -291,11 +301,53 @@ peers should pin this content.
...
+							waitTimeout := c.Duration("wait-timeout")
+							ctx := context.Background()
+
+							if waitTimeout > checkFreq {
+								var cancel context.CancelFunc
+								ctx, cancel = context.WithTimeout(ctx, waitTimeout)
+								defer cancel()
+							}
+
+							sf := client.NewStatusFilter()
```
hsanjuan: Following from comments above, this whole code block would be reduced to:

```
fp := client.FilterParams{...}
st, err := client.WaitFor(fp)
if err != nil {
  /format err response
  return
}
formatResponse(c, st, nil)
```

sharness/t0030-ctl-pin.sh
```diff
@@ -14,13 +14,27 @@ test_expect_success IPFS,CLUSTER "pin data to cluster with ctl" '
     ipfs-cluster-ctl status "$cid" | grep -q -i "PINNED"
 '
 
+test_expect_success IPFS,CLUSTER "wait for data to pin to cluster with ctl" '
```
hsanjuan: :100:  awesome for including sharness tests!

api/rest/client/methods.go
```diff
@@ -172,3 +178,106 @@ func (c *Client) GetConnectGraph() (api.ConnectGraphSerial, error) {
...
+				return
+			case gblPinInfo := <-sf.In:
```
hsanjuan: since you are closing `sf.In`, better check `gblPinInfo, more := <- sf.In` and exit if more == false.

lanzafame:
@hsanjuan just squashed commits, should be good to merge now 👍

hsanjuan:
Thought I had merged this.. THANKS

Issue #339, opened 2018-03-09
Reduce/adjust sleep times in tests


hsanjuan:
We now have finer control over some consensus configuration options.

Currently tests take too long and probably most of that time we are just sleeping. We need to revisit all those delays and adjust them.

PR #349, opened 2018-03-14
WIP: Feat/restapi libp2p

hsanjuan:

PR #350, opened 2018-03-16
Feat #339: faster tests

hsanjuan:
This reduces main package test times from +15 minutes to ~4 in my machine. Let's see how it fares in travis and jenkins.

#339


hsanjuan:
Discovered that random test failures in our tests mostly come from libp2p "dial backoff" when bootstrapping test peers in parallel.  Once a dial backoff happens, libp2p mechanisms prevent any retry for at least 5 seconds (hardcoded in libp2p-swarm). Before, we got a second chance because we slept for longer, but now it was making at least one or two tests fail. Starting peers sequentially with a small sleep between them seems to help.

https://github.com/libp2p/go-libp2p-swarm/issues/64

hsanjuan:
Seems travis consistently passes now, and Jenkins mostly fails on windows only.

I'd like to merge #349 before I keep trying. Creating the hosts and connecting them before creating the peers might do the trick resolve the problems.

ci/Jenkinsfile
```diff
@@ -1,2 +1,2 @@
-golang([test: "go test -v -timeout 20m ./..."])
+golang([test: "go test -v -loglevel ERROR ./..."])
```
ZenGround0: why have a different loglevel between jenkins and travis?

cluster_config.go
```diff
@@ -105,10 +106,16 @@ type Config struct {
...
 	ReplicationFactorMin int
 
-	// MonitorPingInterval is frequency by which a cluster peer pings the
-	// monitoring component. The ping metric has a TTL set to the double
+	// MonitorPingInterval is the frequency by which a cluster peer pings
+	// the monitoring component. The ping metric has a TTL set to the double
 	// of this value.
 	MonitorPingInterval time.Duration
+
+	// PeerWatchInterval is the frequency that we watch for changes
```
ZenGround0: again minor typo but "frequency we use to watch for changes" or "frequency with which we watch for changes" is more correct.

coverage.sh
```diff
@@ -7,9 +7,9 @@ for dir in $dirs;
 do
         if ls "$dir"/*.go &> /dev/null;
         then
-            cmdflags="-timeout 20m -v -coverprofile=profile.out -covermode=count $dir"
```
ZenGround0: Are we using this script anymore?  From the makefile it looks like we are just calling `go test` directly?

ipfscluster_test.go
```diff
@@ -205,26 +205,37 @@ func createClusters(t *testing.T) ([]*Cluster, []*test.IpfsMock) {
 	for i := 1; i < nClusters; i++ {
 		cfgs[i].Bootstrap = []ma.Multiaddr{bootstrapAddr}
 	}
+	time.Sleep(200 * time.Millisecond)
 
 	// Start the rest
-	var wg sync.WaitGroup
+	// Doing this in parallel causes libp2p dial backoffs
```
ZenGround0: it looks like this changed from parallel to sequential which also makes sense in terms of avoiding collisions and backoffs.  Is the comment off or am I misinterpreting the code?

ci/Jenkinsfile
hsanjuan: temporary

coverage.sh
hsanjuan: yeah, travis.yml

ipfscluster_test.go
hsanjuan: oh, the comment just explains why it's not in parallel. ill change it

ZenGround0:
Jenkins failures are only on windows

My comment from above still stands though.  It would be great to 
1) write up the implicit knowledge encoded in testing wait intervals and the reason for doing particular waits before/after certain operations.  I'm thinking of this as the beginning of "A guide to testing in ipfs-cluster".  
2) find a good place for this documentation.  comments in testfile? readme in test folder? document in docs folder?
I'm ok with merging this as is without documenting these things now, but I think they will be very helpful for contributors starting out in the repo, particularly because contributors so often start with testing.  Maybe we could make an issue?

Let me know if you disagree with this.  Perhaps this implicit knowledge is already encoded well enough in the existing comments to enable new developers and I'm just not looking in the right places.

hsanjuan:
travis passes very consistently and jenkins more or less. So same as before, except everything takes 1/3 of the time at worst. Merging!

PR #352, opened 2018-03-18
State change for sharding support

ZenGround0:
@hsanjuan This is my stab at the new state format after our discussion last week.  Could I get some feedback on the representation of the different "PinTypes"?  Some questions:
* Are you in favor of the datatype's structure?
* Are any of the new names used confusing?
* Are you ok with a single Child field (queries for children of metadata nodes must go through ipfs)?
* Is the pattern of updating the Log, to account for the fact that shard pins will only know of their parents after they are pinned, acceptable?
* Any problems with the addition of a new trackerStatus to denote metapins?  Local and remote didn't seem quite right because potentially some shards are local and some are remote.
* should the `status` command display to the user a field to distinguish if a pin is metadata (cdag or shard) or data?

TODO:
[ ] ipfsconn method for retrieving clusterdag index + tests for this method
[ ] Finish Unpin functionality, including multiple parent awareness and replication factor resetting if max/min changes after one parent unpins
[ ] Pin and Unpin testing against each PinType
[ ] only display metadata pins when requested (ls -a)
[ ] state format migration to new PinSerial type

### DIFF
--- a/api/rest/client/methods.go
+++ b/api/rest/client/methods.go
@@ -77,9 +77,9 @@ func (c *Client) Unpin(ci *cid.Cid) error {
 
 // Allocations returns the consensus state listing all tracked items and
 // the peers that should be pinning them.
-func (c *Client) Allocations() ([]api.Pin, error) {
+func (c *Client) Allocations(pinType api.PinType) ([]api.Pin, error) {
 	var pins []api.PinSerial
-	err := c.do("GET", "/allocations", nil, &pins)
+	err := c.do("GET", fmt.Sprintf("/allocations?pintype=%s", pinType.String()), nil, &pins)
 	result := make([]api.Pin, len(pins))
 	for i, p := range pins {
 		result[i] = p.ToPin()
--- a/api/rest/restapi.go
+++ b/api/rest/restapi.go
@@ -646,13 +646,33 @@ func (api *API) unpinHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// filterOutPin returns true if the given pin should be filtered out according
+// to the input filter type
+func (api *API) filterOutPin(filter types.PinType, pin types.Pin) bool {
+	if filter == types.AllType {
+		return false
+	}
+	return pin.Type != filter
+}
+
 func (api *API) allocationsHandler(w http.ResponseWriter, r *http.Request) {
+	queryValues := r.URL.Query()
+	pintype := queryValues.Get("pintype")
+	filter := types.PinTypeFromString(pintype)
 	var pins []types.PinSerial
-	err := api.rpcClient.Call("",
+	err := api.rpcClient.Call(
+		"",
 		"Cluster",
 		"Pins",
 		struct{}{},
-		&pins)
+		&pins,
+	)
+	for i, pinS := range pins {
+		if api.filterOutPin(filter, pinS.ToPin()) {
+			// remove this pin from output
+			pins = append(pins[:i], pins[i+1:]...)
+		}
+	}
 	sendResponse(w, err, pins)
 }
 
@@ -820,7 +840,8 @@ func parseCidOrError(w http.ResponseWriter, r *http.Request) types.PinSerial {
 	}
 
 	pin := types.PinSerial{
-		Cid: hash,
+		Cid:  hash,
+		Type: types.DataType,
 	}
 
 	queryValues := r.URL.Query()
--- a/api/types.go
+++ b/api/types.go
@@ -48,6 +48,9 @@ const (
 	TrackerStatusUnpinned
 	// The IPFS deamon is not pinning the item but it is being tracked
 	TrackerStatusRemote
+	// The IPFS daemon is not pinning the item through this cid but it is
+	// tracked in a cluster dag
+	TrackerStatusSharded
 )
 
 // TrackerStatus represents the status of a tracked Cid in the PinTracker
@@ -515,22 +518,109 @@ func StringsToPeers(strs []string) []peer.ID {
 	return peers
 }
 
+// CidsToStrings encodes cid.Cids to strings.
+func CidsToStrings(cids []*cid.Cid) []string {
+	strs := make([]string, len(cids))
+	for i, c := range cids {
+		strs[i] = c.String()
+	}
+	return strs
+}
+
+// StringsToCidSet decodes cid.Cids from strings.
+func StringsToCidSet(strs []string) *cid.Set {
+	cids := cid.NewSet()
+	for _, str := range strs {
+		c, err := cid.Decode(str)
+		if err != nil {
+			logger.Error(str, err)
+		}
+		cids.Add(c)
+	}
+	return cids
+}
+
+// PinType values
+const (
+	DataType = iota + 1
+	MetaType
+	CdagType
+	ShardType
+)
+
+// AllType is a PinType used for filtering all pin types
+const AllType = -1
+
+// PinType specifies which of four possible interpretations a pin represents.
+// DataType pins are the simplest and represent a pin in the pinset used to
+// directly track user data.  ShardType pins are metadata pins that track
+// many nodes in a user's data DAG.  ShardType pins have a parent pin, and in
+// general can have many parents.  ClusterDAG, or Cdag for short, pins are also
+// metadata pins that do not directly track user data DAGs but rather other
+// metadata pins.  CdagType pins have at least one parent.  Finally MetaType
+// pins always track the cid of the root of a user-tracked data DAG.  However
+// MetaType pins are not stored directly in the ipfs pinset.  Instead the
+// underlying DAG is tracked via the metadata pins underneath the root of a
+// CdagType pin
+type PinType int
+
+// PinTypeFromString is the inverse of String.  It returns the PinType value
+// corresponding to the input string
+func PinTypeFromString(str string) PinType {
+	switch str {
+	case "pin":
+		return DataType
+	case "meta-pin":
+		return MetaType
+	case "clusterdag-pin":
+		return CdagType
+	case "shard-pin":
+		return ShardType
+	case "all":
+		return AllType
+	default:
+		return PinType(0) // invalid string
+	}
+}
+
+// String returns a printable value to identify the PinType
+func (pT *PinType) String() string {
+	switch *pT {
+	case DataType:
+		return "pin"
+	case MetaType:
+		return "meta-pin"
+	case CdagType:
+		return "clusterdag-pin"
+	case ShardType:
+		return "shard-pin"
+	case AllType:
+		return "all"
+	default:
+		panic("String() called on invalid shard type")
+	}
+}
+
 // Pin is an argument that carries a Cid. It may carry more things in the
 // future.
 type Pin struct {
 	Cid                  *cid.Cid
 	Name                 string
+	Type                 PinType
 	Allocations          []peer.ID
 	ReplicationFactorMin int
 	ReplicationFactorMax int
 	Recursive            bool
+	Parents              *cid.Set
+	Clusterdag           *cid.Cid
 }
 
 // PinCid is a shortcut to create a Pin only with a Cid.  Default is for pin to
-// be recursive
+// be recursive and the pin to be of DataType
 func PinCid(c *cid.Cid) Pin {
 	return Pin{
 		Cid:         c,
+		Type:        DataType,
 		Allocations: []peer.ID{},
 		Recursive:   true,
 	}
@@ -540,10 +630,13 @@ func PinCid(c *cid.Cid) Pin {
 type PinSerial struct {
 	Cid                  string   `json:"cid"`
 	Name                 string   `json:"name"`
+	Type                 int      `json:"type"`
 	Allocations          []string `json:"allocations"`
 	ReplicationFactorMin int      `json:"replication_factor_min"`
 	ReplicationFactorMax int      `json:"replication_factor_max"`
 	Recursive            bool     `json:"recursive"`
+	Parents              []string `json:"parents"`
+	Clusterdag           string   `json:"clusterdag"`
 }
 
 // ToSerial converts a Pin to PinSerial.
@@ -552,17 +645,28 @@ func (pin Pin) ToSerial() PinSerial {
 	if pin.Cid != nil {
 		c = pin.Cid.String()
 	}
+	cdag := ""
+	if pin.Clusterdag != nil {
+		cdag = pin.Clusterdag.String()
+	}
 
 	n := pin.Name
 	allocs := PeersToStrings(pin.Allocations)
+	var parents []string
+	if pin.Parents != nil {
+		parents = CidsToStrings(pin.Parents.Keys())
+	}
 
 	return PinSerial{
 		Cid:                  c,
 		Name:                 n,
 		Allocations:          allocs,
+		Type:                 int(pin.Type),
 		ReplicationFactorMin: pin.ReplicationFactorMin,
 		ReplicationFactorMax: pin.ReplicationFactorMax,
 		Recursive:            pin.Recursive,
+		Parents:              parents,
+		Clusterdag:           cdag,
 	}
 }
 
@@ -578,6 +682,10 @@ func (pin Pin) Equals(pin2 Pin) bool {
 	}
 
 	if pin1s.Name != pin2s.Name {
+		return false
+	}
+
+	if pin1s.Type != pin2s.Type {
 		return false
 	}
 
@@ -599,6 +707,18 @@ func (pin Pin) Equals(pin2 Pin) bool {
 	if pin1s.ReplicationFactorMin != pin2s.ReplicationFactorMin {
 		return false
 	}
+
+	if pin1s.Clusterdag != pin2s.Clusterdag {
+		return false
+	}
+
+	sort.Strings(pin1s.Parents)
+	sort.Strings(pin2s.Parents)
+
+	if strings.Join(pin1s.Parents, ",") != strings.Join(pin2s.Parents, ",") {
+		return false
+	}
+
 	return true
 }
 
@@ -608,14 +728,24 @@ func (pins PinSerial) ToPin() Pin {
 	if err != nil {
 		logger.Error(pins.Cid, err)
 	}
+	var cdag *cid.Cid
+	if pins.Clusterdag != "" {
+		cdag, err = cid.Decode(pins.Clusterdag)
+		if err != nil {
+			logger.Error(pins.Clusterdag, err)
+		}
+	}
 
 	return Pin{
 		Cid:                  c,
 		Name:                 pins.Name,
 		Allocations:          StringsToPeers(pins.Allocations),
+		Type:                 PinType(pins.Type),
 		ReplicationFactorMin: pins.ReplicationFactorMin,
 		ReplicationFactorMax: pins.ReplicationFactorMax,
 		Recursive:            pins.Recursive,
+		Parents:              StringsToCidSet(pins.Parents),
+		Clusterdag:           cdag,
 	}
 }
 
--- a/cluster.go
+++ b/cluster.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/ipfs/ipfs-cluster/api"
 	"github.com/ipfs/ipfs-cluster/consensus/raft"
+	"github.com/ipfs/ipfs-cluster/sharder"
 	"github.com/ipfs/ipfs-cluster/state"
 
 	rpc "github.com/hsanjuan/go-libp2p-gorpc"
@@ -19,6 +20,7 @@ import (
 	ipnet "github.com/libp2p/go-libp2p-interface-pnet"
 	peer "github.com/libp2p/go-libp2p-peer"
 	peerstore "github.com/libp2p/go-libp2p-peerstore"
+	p2praft "github.com/libp2p/go-libp2p-raft"
 	swarm "github.com/libp2p/go-libp2p-swarm"
 	basichost "github.com/libp2p/go-libp2p/p2p/host/basic"
 	ma "github.com/multiformats/go-multiaddr"
@@ -960,8 +962,8 @@ func (c *Cluster) Pins() []api.Pin {
 		logger.Error(err)
 		return []api.Pin{}
 	}
-
 	return cState.List()
+
 }
 
 // PinGet returns information for a single Cid managed by Cluster.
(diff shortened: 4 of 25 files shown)


api/types.go
```diff
@@ -515,19 +518,65 @@ func StringsToPeers(strs []string) []peer.ID {
...
 type Pin struct {
 	Cid                  *cid.Cid
 	Name                 string
+	Type                 PinType
 	Allocations          []peer.ID
 	ReplicationFactorMin int
 	ReplicationFactorMax int
 	Recursive            bool
+	Parents              []*cid.Cid
+	Child                *cid.Cid
```
hsanjuan: Maybe `Cdag` ? For consistency? It makes me wonder if `cdag` is a good name, or we should write `clusterdag` in full.

api/types.go
```diff
@@ -592,13 +656,23 @@ func (pin Pin) Equals(pin2 Pin) bool {
 		return false
 	}
 
+	sort.Strings(pin1s.Parents)
```
hsanjuan: Move the parents tests to the last position, that way we don't need to do it if something else is different. Thus marginally cheaper.

api/types.go
```diff
@@ -608,14 +682,21 @@ func (pins PinSerial) ToPin() Pin {
 	if err != nil {
 		logger.Error(pins.Cid, err)
 	}
+	child, err := cid.Decode(pins.Child)
+	if err != nil {
+		logger.Error(pins.Child, err)
```
hsanjuan: This is potentially going to complain all the time with pins which rightfully don't have a child. Should just not try to decode if child is nil.

api/types.go
```diff
@@ -515,19 +518,65 @@ func StringsToPeers(strs []string) []peer.ID {
...
+// directly track user data.  ShardType pins are metadata pins that track
+// many nodes in a user's data DAG.  ShardType pins have a parent pin, and in
+// general can have many parents.  ClusterDAG, or Cdag for short, pins are also
+// metadata pins that do not directly track user data DAGs but rather other
+// metadata pins.  CdagType pins have at least one parent.  Finally MetaType
+// pins always track the cid of the root of a user-tracked data DAG.  However
+// MetaType pins are not stored directly in the ipfs pinset.  Instead the
+// underlying DAG is tracked via the metadata pins underneath the root of a
+// CdagType pin
+type PinType int
```
hsanjuan: Add a `String()` method for this, it will be handy for UX.

cluster.go
```diff
@@ -995,8 +995,18 @@ func (c *Cluster) PinGet(h *cid.Cid) (api.Pin, error) {
 // the cluster.  Priority allocations are best effort.  If any priority peers
 // are unavailable then Pin will simply allocate from the rest of the cluster.
 func (c *Cluster) Pin(pin api.Pin) error {
-	_, err := c.pin(pin, []peer.ID{}, pin.Allocations)
-	return err
+	switch pin.Type {
```
hsanjuan: This switch/logic should probably go into the unexported `pin()`, which is what `repinFromPeer` calls (if I remember well). I'm pretty sure it's going to cause weird side effects otherwise.

cluster.go
```diff
@@ -1056,16 +1066,69 @@ func (c *Cluster) pin(pin api.Pin, blacklist []peer.ID, prioritylist []peer.ID)
...
+	}
+	pin := cState.Get(h)
+
+	switch pin.Type {
+	case api.DataType:
+		return c.consensus.LogUnpin(pin)
+	case api.ShardType:
+		warn := "unpinning shard cid %s without unpinning parent"
+		logger.Warningf(warn, h.String())
+		return c.consensus.LogUnpin(pin)
```
hsanjuan: Not sure if it should be allowed at all

cluster.go
```diff
@@ -1056,16 +1066,69 @@ func (c *Cluster) pin(pin api.Pin, blacklist []peer.ID, prioritylist []peer.ID)
 // of underlying IPFS daemon unpinning operations.
 func (c *Cluster) Unpin(h *cid.Cid) error {
 	logger.Info("IPFS cluster unpinning:", h)
+	cState, err := c.consensus.State()
```
hsanjuan: I am not 100% sure this (the `Cluster` component) is the right place to have this logic. Might as well be part of LogOp Apply(), although I see advantages of keeping it here. Let me think about this until next round.

cluster.go
```diff
@@ -1056,16 +1066,69 @@ func (c *Cluster) pin(pin api.Pin, blacklist []peer.ID, prioritylist []peer.ID)
...
+		return errors.New("unrecognized pin type")
 	}
+}
 
-	err := c.consensus.LogUnpin(pin)
+// unpinClusterDag unpins the clusterDAG metadata node and the shard metadata
+// nodes that it references.  It handles the case where multiple parents
+// reference the same metadata node, only unpinning those nodes without
+// existing references
+func (c *Cluster) unpinClusterDag(metaPin api.Pin) {
```
hsanjuan: I foresee this will have to return an error :)

cluster.go
```diff
@@ -1056,16 +1066,69 @@ func (c *Cluster) pin(pin api.Pin, blacklist []peer.ID, prioritylist []peer.ID)
...
+
+	// Read through links in clusterDAG node (maybe use sharder for this, probably can use ipld directly or maybe abstract things into a utility helpfer function?
+
+	//
+	return
+}
+
+// UpdatePin updates the shared state's view on the data included with the cid.
+// Currently only supports updating the parent cid of shard nodes
+func (c *Cluster) UpdatePin(h, parent *cid.Cid) error {
```
hsanjuan: I don't think this needs to be exported. Maybe callable from RPC but not exported. Also, should be able to just call pin().

cluster.go
```diff
@@ -1056,16 +1066,69 @@ func (c *Cluster) pin(pin api.Pin, blacklist []peer.ID, prioritylist []peer.ID)
...
+		return errors.New("cannot update pin uncommitted to state")
+	}
+	pin := cState.Get(h)
+	if pin.Type != api.ShardType {
+		return errors.New("cannot update on non-shard type")
+	}
+	if pin.Parents == nil {
+		pin.Parents = make([]*cid.Cid, 0)
+	}
+	pin.Parents = append(pin.Parents, parent)
```
hsanjuan: `append` does the magic of creating the slice for you if it's nil. Thus, the previous check and `make` is not necessary.

consensus/raft/log_op.go
```diff
@@ -61,6 +62,15 @@ func (op *LogOp) ApplyTo(cstate consensus.State) (consensus.State, error) {
 			op.Cid,
 			&struct{}{},
 			nil)
+	case LogOpUpdate:
+		err = state.Rm(op.Cid.ToPin().Cid)
+		if err != nil {
+			goto ROLLBACK
+		}
+		err = state.Add(op.Cid.ToPin())
```
hsanjuan: Removing before adding has no benefits.. adds just replace. My impression is that this is equivalent to LogOpPin.

cluster.go
ZenGround0: Eventually we will likely want to provide a way for users to evict shards and clusterdags manually from the state at least in the case a crash has occurred and the state should be cleaned.  

Right now given our assumption that data dags are being streamed and cannot fit entirely on a peer during sharding, shard components must be pinned without having a metaPin reference linked to them when shards are complete but the whole clusterdag is not.  We could also avoid this by just block putting everything and waiting to Pin until everything is ready.  Of course all of this belongs in a future discussion around data safety and recovery.  For now I am fine with just erroring when this happens and using `state cleanup` if unlinked shard pins make it into the state.

cluster.go
ZenGround0: Let me know if you would prefer this in Apply().

ZenGround0:
@hsanjuan mind giving this a look?

cluster.go
```diff
@@ -1031,6 +1036,66 @@ func (c *Cluster) pin(pin api.Pin, blacklist []peer.ID, prioritylist []peer.ID)
...
+		if len(pin.Parents) > 1 {
+			return false, errors.New("up to one parent per commitment of a shard")
+		}
+		// In general multiple clusterdags may reference the same shard
+		// and sharder sessions typically update a shard pin's metadata.
+		// Hence we check for an existing shard and carefully update.
+		var empty bool // skip check when state does not yet exist
+		cState, err := c.consensus.State()
+		if err != nil {
+			if strings.Contains(err.Error(), "no state has been agreed upon yet") {
```
hsanjuan: Can we make "no estate has been agreed upon yet" an `ErrNoState` in consensus and check against it?

cluster.go
```diff
@@ -1031,6 +1036,66 @@ func (c *Cluster) pin(pin api.Pin, blacklist []peer.ID, prioritylist []peer.ID)
...
+			}
+		}
+		if !empty && cState.Has(pin.Cid) {
+			existing := cState.Get(pin.Cid)
+			// For now all repins of the same shard must use the same
+			// replmax and replmin.  It is unclear what the best UX is here
+			// especially if the same Shard is referenced in multiple
+			// clusterdags.  This simplistic policy avoids complexity and
+			// suits existing needs for shard pins.
+			if existing.ReplicationFactorMin != rplMin ||
```
hsanjuan: We should take the most conservative approach (that is, the highest value for min and max).

api/rest/client/methods.go
```diff
@@ -77,9 +77,9 @@ func (c *Client) Unpin(ci *cid.Cid) error {
 
 // Allocations returns the consensus state listing all tracked items and
 // the peers that should be pinning them.
-func (c *Client) Allocations() ([]api.Pin, error) {
+func (c *Client) Allocations(hidden bool) ([]api.Pin, error) {
```
hsanjuan: I'm not much into this naming. `all`, `hidden`, `a`... if we are going to filter, have a filter type int as receiver with values for `all/meta/pin/cdag/shard`.

`Allocations(pinType)`

where -1 means all.

api/rest/client/methods.go
```diff
@@ -77,9 +77,9 @@ func (c *Client) Unpin(ci *cid.Cid) error {
...
 	var pins []api.PinSerial
-	err := c.do("GET", "/allocations", nil, &pins)
+	err := c.do("GET", fmt.Sprintf("/allocations?a=%t", hidden), nil, &pins)
```
hsanjuan: rename query param to `pintype`

api/rest/restapi.go
```diff
@@ -647,12 +647,27 @@ func (api *API) unpinHandler(w http.ResponseWriter, r *http.Request) {
...
+			"Cluster",
+			"Pins",
+			struct{}{},
+			&pins,
+		)
+	} else {
+		err = api.rpcClient.Call(
+			"",
+			"Cluster",
+			"PinsQuiet",
```
hsanjuan: Handle all pintype filtering here in the REST API. For the moment I want this to be an API feature because it is solely a presentation helper (I could perfectly live without it) and I don't want it to change internal APIs just because the user wants less things on the screen.

When the time comes that our states are so big that this becomes a feature for efficiency, then it should be addressed in the state component direclty and easily removed from here.

api/types.go
```diff
@@ -561,6 +561,22 @@ const (
...
+func (pT PinType) String() string {
+	switch pT {
+	case DataType:
+		return "pin"
+	case MetaType:
+		return "meta-pin"
+	case CdagType:
+		return "clusterdag-pin"
+	case ShardType:
+		return "clusterdag-shard-pin"
```
hsanjuan: I can live with "shard-pin".

cluster.go
```diff
@@ -954,14 +955,26 @@ func (c *Cluster) RecoverLocal(h *cid.Cid) (api.PinInfo, error) {
...
 		return []api.Pin{}
 	}
+	pins := cState.List()
+	if !quiet {
+		return pins
+	}
+	// Filter out shard and clusterdag internal pins
+	quietPins := make([]api.Pin, 0)
+	for _, pin := range pins {
+		if pin.Type == api.CdagType || pin.Type == api.ShardType {
```
hsanjuan: As pointed above, let the filtering happen on the API only as it is purely a cosmetic feature for the moment.

rpc_api.go
```diff
@@ -42,18 +42,20 @@ func (rpcapi *RPCAPI) Unpin(ctx context.Context, in api.PinSerial, out *struct{}
...
-	return rpcapi.c.UpdatePin(pin.Cid, pin.Parents[0])
+	*out = cidSerialList
+	return nil
 }
 
-// Pins runs Cluster.Pins().
-func (rpcapi *RPCAPI) Pins(ctx context.Context, in struct{}, out *[]api.PinSerial) error {
-	cidList := rpcapi.c.Pins()
+// PinsQuiet runs Cluster.Pins(true)
+func (rpcapi *RPCAPI) PinsQuiet(ctx context.Context, in struct{}, out *[]api.PinSerial) error {
```
hsanjuan: Per comments above, this can go away

cluster.go
```diff
@@ -1021,6 +1036,66 @@ func (c *Cluster) pin(pin api.Pin, blacklist []peer.ID, prioritylist []peer.ID)
 	if err := isReplicationFactorValid(rplMin, rplMax); err != nil {
 		return false, err
 	}
+	switch pin.Type {
+	case api.DataType:
+		// Fall through
```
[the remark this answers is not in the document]
ZenGround0: Oh my comment is misleading.  I want the DataType pins to leave the switch without executing any code.

cluster.go
```diff
@@ -1056,16 +1131,94 @@ func (c *Cluster) pin(pin api.Pin, blacklist []peer.ID, prioritylist []peer.ID)
...
+	if !cState.Has(h) {
+		return errors.New("cannot unpin pin uncommitted to state")
+	}
+	pin := cState.Get(h)
+
+	switch pin.Type {
+	case api.DataType:
+		return c.consensus.LogUnpin(pin)
+	case api.ShardType:
+		err := "unpinning shard cid %s before unpinning parent"
```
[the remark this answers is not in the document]
ZenGround0: Glad you asked this.  It is possible right now in the case of a cluster failure during sharding.  Because sharding is currently approaching a prototype that will not have any data safety guarantees or builtin recovery mechanisms this might be useful in the near future.  The alternative will be cleaning up a failed sharding session by cleaning up the state and starting fresh.  I am happy to change this to a warning if you prefer.  Might also be nice to include the `-f` flag into the semantics of unpin so that certain pin types can only be unpinned if done with the force flag.

cluster.go
ZenGround0: Thanks for bringing this up.  I've thought a lot about this and am still somewhat lost.  I thought a lot about taking the most conservative approach but I thought that would look like the highest value for min and the lowest value for max, after all users specify a max value to ensure that cids will not be pinned more than some threshold.  If we take the highest value for max, aren't we violating someone's initial assumption?

If we take the lowest max and the highest min we can potentially get into a situation where our highest min is higher than our lowest max and so we won't be able to satisfy anyone's constraints!   There are plenty of rules we could come up with, take the last, take the first, but I don't know which one is correct and I think that ends up being a UX question.

I kept the rule enforcing that pins stay consistent to enforce a rigid pattern of use that while potentially restrictive would prevent the possibility of confusing situations or implicitly redefining the ideas behind replmin and replmax.  I also did this because enforcing other invariants on repl factors requires more complexity during the unpin process to ensure that repl factors represent the invariant over the *current* set of parents.  I was not sure if this complexity was warranted.

If you'd still like high min and high max or some other invariant I'm happy to do it.  Any other guidance or thoughts is also very appreciated.

cluster.go
```diff
@@ -1021,6 +1036,66 @@ func (c *Cluster) pin(pin api.Pin, blacklist []peer.ID, prioritylist []peer.ID)
...
+			return false, errors.New("cdag nodes are referenced once")
+		}
+	case api.MetaType:
+		if len(pin.Allocations) != 0 {
+			return false, errors.New("meta pin should not specify allocations")
+		}
+		return true, c.consensus.LogPin(pin)
+	default:
+		return false, errors.New("unrecognized pin type")
+	}
```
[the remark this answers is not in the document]
ZenGround0: validatePin() sounds good.  What do you mean take care of setting the parents right "in the Shardtype"?

cluster.go
```diff
@@ -1021,6 +1036,66 @@ func (c *Cluster) pin(pin api.Pin, blacklist []peer.ID, prioritylist []peer.ID)
...
+			// suits existing needs for shard pins.
+			if existing.ReplicationFactorMin != rplMin ||
+				existing.ReplicationFactorMax != rplMax {
+				return false, errors.New("shard update with wrong repl factors")
+			}
+			// Pin with correct parents
+			if len(pin.Parents) == 1 &&
+				!containsCid(existing.Parents, pin.Parents[0]) {
+
+				pin.Parents = append(existing.Parents, pin.Parents[0])
```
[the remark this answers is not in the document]
ZenGround0: If you prefer this I will change to use `cid.Set`, but in my opinion `containsCid` seems cleaner than adding all parents to the set in a for loop and then converting back to slices at the end.

cluster.go
```diff
@@ -1021,6 +1036,66 @@ func (c *Cluster) pin(pin api.Pin, blacklist []peer.ID, prioritylist []peer.ID)
...
 	}
+	switch pin.Type {
+	case api.DataType:
+		// Fall through
+	case api.ShardType:
+		if !pin.Recursive {
+			return false, errors.New("must pin shards recursively")
+		}
+		if len(pin.Parents) > 1 {
+			return false, errors.New("up to one parent per commitment of a shard")
```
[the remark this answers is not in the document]
ZenGround0: Thanks this will need some rethinking

cluster.go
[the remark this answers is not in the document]
ZenGround0: Nice this makes new parent behavior easier!

ZenGround0:
Hey @hsanjuan check some of the outdated comments as I've left comments that need feedback from you on them.  Thanks!

cluster.go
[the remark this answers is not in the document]
hsanjuan: oh god did I ask you to disallow it on the first review and now I was asking you to re-allow it?

cluster.go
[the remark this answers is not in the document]
hsanjuan: ah yes :)

api/rest/client/methods.go
```diff
@@ -77,9 +77,9 @@ func (c *Client) Unpin(ci *cid.Cid) error {
 
 // Allocations returns the consensus state listing all tracked items and
 // the peers that should be pinning them.
-func (c *Client) Allocations(hidden bool) ([]api.Pin, error) {
+func (c *Client) Allocations(pinType api.FilterType) ([]api.Pin, error) {
 	var pins []api.PinSerial
-	err := c.do("GET", fmt.Sprintf("/allocations?a=%t", hidden), nil, &pins)
+	err := c.do("GET", fmt.Sprintf("/allocations?pintype=%d", int(pinType)), nil, &pins)
```
hsanjuan: use a string rather than a number?

api/rest/restapi.go
```diff
@@ -646,27 +646,42 @@ func (api *API) unpinHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// filterOutPin returns true if the given pin should be filtered out according
+// to the input filter type
+func (api *API) filterOutPin(filter types.FilterType, pin types.Pin) bool {
+	switch filter {
+	case types.FilterPin:
```
hsanjuan: I have a question: why do you need a FilterType if you already have `PinType` defined and they' re the same thing (+ `-1` meaning all)

cluster.go
```diff
@@ -1012,6 +1001,70 @@ func (c *Cluster) Pin(pin api.Pin) error {
...
+		// replmax and replmin.  It is unclear what the best UX is here
+		// especially if the same Shard is referenced in multiple
+		// clusterdags.  This simplistic policy avoids complexity and
+		// suits existing needs for shard pins.
+		if existing.ReplicationFactorMin != rplMin ||
+			existing.ReplicationFactorMax != rplMax {
+			return errors.New("shard update with wrong repl factors")
+		}
+		// Pin with union of old and new parents
+		parents := cid.NewSet()
```
hsanjuan: what I meant is that the `Parents` field in the `Pin` struct should be of type cid.Set. ToSerial() would convert the set to an array (so the serial representation is the same). ToPin() would add the elements in the array to the set. Then you can work directly with the set.

hsanjuan: Since the serial form doesn't change, you won't need a state migration.

api/rest/restapi.go
ZenGround0: I thought this was what you you were asking for with "have a filter type int as a receiver" :).  I will gladly switch to strings.

cluster.go
ZenGround0: I like it!

cluster.go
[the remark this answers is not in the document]
ZenGround0: haha not exactly. Last review you said "unsure we should allow this at all" as a comment to the warning I had in place.

cluster.go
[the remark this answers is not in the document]
ZenGround0: Ah I'll modify this then.

cluster.go
ZenGround0: @hsanjuan could you respond to this comment please?

api/rest/restapi.go
hsanjuan: I mean, using `api.PinType`. Strings is only for the query parameter.

api/types.go
```diff
@@ -548,6 +548,8 @@ const (
 	ShardType
 )
 
+const AllType = -1
```
hsanjuan: Group with PinTypes or add a comment. Golint doesn't complain?

api/rest/client/methods.go
```diff
@@ -77,9 +77,9 @@ func (c *Client) Unpin(ci *cid.Cid) error {
 
 // Allocations returns the consensus state listing all tracked items and
 // the peers that should be pinning them.
-func (c *Client) Allocations(pinType api.FilterType) ([]api.Pin, error) {
+func (c *Client) Allocations(pinType string) ([]api.Pin, error) {
```
hsanjuan: `type types.PinType`. Receive a PinType, send a string on the query (because that's user friendly)

api/types.go
```diff
@@ -561,9 +563,24 @@ const (
 // CdagType pin
 type PinType int
 
+func (pT *PinType) Parse(str string) {
```
hsanjuan: I don't know which one is better, but we have `TrackerStatusFromString(str string) TrackerStatus`, then for consisntency this should be `PinTypeFromString(...)`

hsanjuan:
@ZenGround0 made comments in outdated commit, but they still stand

api/types.go
ZenGround0: It did which is why I added a comment!

api/types.go
ZenGround0: Im cool with that 😎

PR #356, opened 2018-03-21
Fix #269: Add IPFS() method to api/rest/client

### LINKED ISSUE #269: rest/client: IpfsShell() method returns an go-ipfs-api to the proxy port
Easily obtaining an ipfs-client facilitates that the integrators can readily access functions offered by the underlying ipfs-nodes.

Would need a `ProxyPort` config option.

hsanjuan:
```
    The IPFS() methods returns an ipfs Shell pointing to the ipfs-cluster
    proxy endpoint. The location can be customized (via ProxyAddr configuration
    option) or it is assumed to be the same as the PeerAddr/APIAddr, with
    a different port (the default).
```

NEEDs REBASING once #349 is merged.


api/rest/client/client.go
```diff
@@ -164,6 +178,52 @@ func (c *Client) setupHostname() error {
...
+		return err
+	}
+	var paddr ma.Multiaddr
+	switch {
+	case c.config.PeerAddr != nil:
+		paddr = ma.Split(c.config.PeerAddr)[0].Encapsulate(port)
+	case c.config.APIAddr != nil: // Host/Port setupHostname sets APIAddr
+		paddr = ma.Split(c.config.APIAddr)[0].Encapsulate(port)
+	default:
+		return errors.New("cannot guess default proxy address")
```
ZenGround0: This message throws me off a bit.  Are we saying that there are many potential defaults that we tried to guess and we can't finish guessing, or are we saying that there is no default and so there is no guessing to be done?  When I read this I wonder "why can't it guess the default, surely the code knows what the default value is?"

Perhaps "cannot find proxy address"?

api/rest/client/client.go
```diff
@@ -55,6 +58,12 @@ type Config struct {
 	// cluster peer, this corresponds to the cluster secret.
 	ProtectorKey []byte
 
+	// ProxyAddr allows to obtain a go-ipfs-api pointing to the
+	// ipfs proxy endpoint of ipfs-cluster. If empty, the location
+	// it will be guessed from the prevalent in PeerAddr/APIAddr/Host,
```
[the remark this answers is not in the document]
hsanjuan: ok

api/rest/client/client.go
hsanjuan: ok. This should be dead code, as when setupProxy() either PeerAddr or APIAddr will be set (or something will have errored before.

PR #359, opened 2018-03-25
Feat/sharding use pin feats

ZenGround0:
This integrates pinning features from previous PRs to complete the sharding prototypes functionality.  It also cleans up a lot of TODOs that are now more clearly out of scope.

This has not yet been tested so stay tuned.

### DIFF
--- a/ipfs-cluster-ctl/main.go
+++ b/ipfs-cluster-ctl/main.go
@@ -205,6 +205,16 @@ chunker: 'rabin-<min>-<avg>-<max>'.  Default is 'size-262144'`,
 					Name:  "hidden, H",
 					Usage: "include files that are hidden.  Only takes effect on recursive add",
 				},
+				cli.IntFlag{
+					Name:  "replication-min, rmin",
+					Value: 0,
+					Usage: "Sets the minimum replication factor for pinning this file",
+				},
+				cli.IntFlag{
+					Name:  "replication-max, rmax",
+					Value: 0,
+					Usage: "Sets the maximum replication factor for pinning this file",
+				},
 			},
 			Action: func(c *cli.Context) error {
 				paths := make([]string, c.NArg(), c.NArg())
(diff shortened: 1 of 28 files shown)


ipfs-cluster-ctl/main.go
```diff
@@ -215,12 +225,23 @@ chunker: 'rabin-<min>-<avg>-<max>'.  Default is 'size-262144'`,
 				// Files are all opened but not read until they are sent.
 				multiFileR, err := parseFileArgs(paths, c.Bool("recursive"))
 				checkErr("serializing all files", err)
+				// Default repl factor for file adding is 1
+				replMin := c.Int("replication-min")
+				if replMin == 0 {
+					replMin = 1
```
hsanjuan: why 1? I think our default is -1 (pin everywhere) in pin/add...

ZenGround0: Good question.  I realized the default for `pin add` is -1.  I would agree that it is not great to have separate defaults for separate commands, but at the time of writing I was considering that the common case for adding sharded files would often be space constrained and so starting out with shards on only one machine might be the better default.

I'm now realizing that this isn't even only set when `--shard` is passed in.  I am good with removing this entirely and using the -1 default.  In hindsight this just seems like an example of me making a choose that was too specific given our limited knowledge of how this command will be used.

hsanjuan: actually, isn't the default 0 - meaning, take whatever the configuration says?

ZenGround0: yes, and the configuration says -1 by default

hsanjuan: ok, but then it should be 0 by default in cluster-ctl

ZenGround0:
@hsanjuan I'm closing this so I can open a PR to v1-sharding now that the state format PR is merged.

Issue #319, opened 2018-02-16
Re-use ipfs/go-fs-lock


hsanjuan:
@dignifiedquire has been kind enough to extract this from go-ipfs: https://github.com/ipfs/go-ipfs/pull/4631

I think we directly copied this code to cluster, so now we can re-use it too.

PR #360, opened 2018-03-26
Use go-fs-lock pkg

lanzafame:
This PR purely, replaces the go4.org/lock dep with github.com/ipfs/go-fs-lock and adds the new dep to package.json.

Resolves: #319.


hsanjuan:
will merge with green tests. thanks

PR #361, opened 2018-03-26
Fix: Recover does not pin recursively

hsanjuan:
First bug from testing 0.3.5-rc1 (good that I did an rc).

@ZenGround0 the pin tracker will need some changes with sharding... probably should carry the pin types of the tracked things.

### DIFF
--- a/pintracker/maptracker/maptracker.go
+++ b/pintracker/maptracker/maptracker.go
@@ -404,9 +404,11 @@ func (mpt *MapPinTracker) Recover(c *cid.Cid) (api.PinInfo, error) {
 	var err error
 	switch p.Status {
 	case api.TrackerStatusPinError:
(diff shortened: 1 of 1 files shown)


hsanjuan:
@ZenGround0 will storing the pinType in the tracker be enough ?

ZenGround0:
No not alone because as it stands validation checks in the `pin` function would raise complaints that the right fields are not set.  We could make changes to the pin function to accommodate this however.

hsanjuan:
@ZenGround0 but the tracker is not involved in those checks, it only needs to tell the ipfs connector how to pin something?

ZenGround0:
Oh, thank you.  I was mixing up repin-from-peer and the pin tracker repinning on ipfs

PR #363, opened 2018-03-28
 Fix: --wait flag in ipfs-cluster-ctl and WaitFor function

hsanjuan:
@lanzafame this was very visibly not working when I tried it. Please test a bit better next time.

### DIFF
--- a/api/rest/client/methods.go
+++ b/api/rest/client/methods.go
@@ -251,12 +251,10 @@ func (sf *statusFilter) filter(ctx context.Context, fp StatusFilterParams) {
 				return
 			}
 
+			sf.Out <- gblPinInfo
 			if !ok {
-				sf.Out <- gblPinInfo
 				continue
 			}
-
-			sf.Out <- gblPinInfo
 			return
 		}
 	}
--- a/api/rest/client/methods.go
+++ b/api/rest/client/methods.go
@@ -191,18 +191,19 @@ func (c *Client) WaitFor(ctx context.Context, fp StatusFilterParams) (api.Global
 	go sf.pollStatus(ctx, c, fp)
 	go sf.filter(ctx, fp)
 
+	var status api.GlobalPinInfo
+
 	for {
 		select {
 		case <-ctx.Done():
 			return api.GlobalPinInfo{}, ctx.Err()
 		case err := <-sf.Err:
 			return api.GlobalPinInfo{}, err
-		case <-sf.Done:
-			// drain status channel and return last status on channel
-			for {
-				if status, ok := <-sf.Out; !ok {
-					return status, nil
-				}
+		case st, ok := <-sf.Out:
+			if ok {
+				status = st
+			} else { // channel closed
+				return status, nil
 			}
 		}
 	}
@@ -211,7 +212,7 @@ func (c *Client) WaitFor(ctx context.Context, fp StatusFilterParams) (api.Global
 // StatusFilterParams contains the parameters required
 // to filter a stream of status results.
 type StatusFilterParams struct {
-	*cid.Cid
+	Cid       *cid.Cid
 	Local     bool
 	Target    api.TrackerStatus
 	CheckFreq time.Duration
--- a/api/rest/client/methods.go
+++ b/api/rest/client/methods.go
@@ -200,11 +200,10 @@ func (c *Client) WaitFor(ctx context.Context, fp StatusFilterParams) (api.Global
 		case err := <-sf.Err:
 			return api.GlobalPinInfo{}, err
 		case st, ok := <-sf.Out:
-			if ok {
-				status = st
-			} else { // channel closed
+			if !ok { // channel closed
 				return status, nil
 			}
+			status = st
 		}
 	}
 }
--- a/api/rest/client/methods_test.go
+++ b/api/rest/client/methods_test.go
@@ -389,7 +389,7 @@ func TestWaitFor(t *testing.T) {
 	tapi.SetClient(rpcC)
 
 	testF := func(t *testing.T, c *Client) {
-		ci, _ := cid.Decode(test.SlowCid)
+		ci, _ := cid.Decode(test.TestCid1)
 
 		var wg sync.WaitGroup
 		wg.Add(1)
--- a/api/rest/client/methods_test.go
+++ b/api/rest/client/methods_test.go
@@ -1,8 +1,15 @@
 package client
 
 import (
+	"context"
+	"sync"
 	"testing"
+	"time"
 
+	rpc "github.com/hsanjuan/go-libp2p-gorpc"
+	peer "github.com/libp2p/go-libp2p-peer"
+
+	"github.com/ipfs/ipfs-cluster/api"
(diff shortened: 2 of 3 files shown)


lanzafame:
@hsanjuan Agreed. Could you help me see where I went wrong with the following sharness tests, as they shouldn't have passed.

From `sharness/t0030-ctl-pin.sh`:
```
expecting success:
    cid=`docker exec ipfs sh -c "dd if=/dev/urandom bs=1024 count=2048 | ipfs add -q"`
    ipfs-cluster-ctl pin add --wait "$cid" | grep -q -i "PINNED" &&
    ipfs-cluster-ctl pin ls "$cid" | grep -q "$cid" &&
    ipfs-cluster-ctl status "$cid" | grep -q -i "PINNED"

expecting success:
    cid=`ipfs-cluster-ctl --enc=json pin ls | jq -r ".[] | .cid" | head -1`
    ipfs-cluster-ctl pin rm --wait "$cid" | grep -q -i "UNPINNED" &&
    !(ipfs-cluster-ctl pin ls "$cid" | grep -q "$cid") &&
    ipfs-cluster-ctl status "$cid" | grep -q -i "UNPINNED"

expecting success:
    cid=`docker exec ipfs sh -c "dd if=/dev/urandom bs=1024 count=2048 | ipfs add -q"`
    ipfs-cluster-ctl pin add --wait --wait-timeout 2s "$cid" | grep -q -i "PINNED" &&
    ipfs-cluster-ctl pin ls "$cid" | grep -q "$cid" &&
    ipfs-cluster-ctl status "$cid" | grep -q -i "PINNED"

expecting success:
    cid=`ipfs-cluster-ctl --enc=json pin ls | jq -r ".[] | .cid" | head -1`
    ipfs-cluster-ctl pin rm --wait --wait-timeout 2s "$cid" | grep -q -i "UNPINNED" &&
    !(ipfs-cluster-ctl pin ls "$cid" | grep -q "$cid") &&
    ipfs-cluster-ctl status "$cid" | grep -q -i "UNPINNED"
```

Also, `api/rest/client/methods_test.go` is failing to build because a reference to `test.SlowCid` which doesn't exist?

ipfs-cluster-ctl/main.go
```diff
@@ -712,3 +674,54 @@ func parseCredentials(userInput string) (string, string) {
...
+		status, cerr = waitFor(ci, target, timeout)
+		checkErr("waiting for pin status", cerr)
+	}
+
+	if noStatus {
+		return
+	}
+
+	if status.Cid == nil { // no status from "wait"
+		time.Sleep(1000 * time.Millisecond)
```
lanzafame: `s/1000*time.Millisecond/1*time.Second`

ipfs-cluster-ctl/main.go
```diff
@@ -398,34 +380,14 @@ although unpinning operations in the cluster may take longer or fail.
...
-							}
-
-							status, cerr := globalClient.WaitFor(ctx, fp)
-							formatResponse(c, status, cerr)
-						}
-
+						waitAndFormatStatusResponse(
+							c,
+							c.Bool("wait"),
+							c.Bool("no-status"),
```
lanzafame: It seems weird that `waitAndFormatStatusResponse` would handle the `--no-status` flag. Maybe `handleResponseFormatFlags` would be a better name and then retrieve `wait` and `no-status` from `*cli.Context`. Or the other way, could be to handle `no-status` outside of `waitAndFormatStatusResponse`. Not sure which would be better? @hsanjuan

api/rest/client/methods_test.go
```diff
@@ -308,3 +317,115 @@ func TestGetConnectGraph(t *testing.T) {
...
+	rpcC := rpc.NewClientWithServer(nil, "wait", rpcS)
+	err := rpcS.RegisterName("Cluster", &waitService{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	tapi.SetClient(rpcC)
+
+	testF := func(t *testing.T, c *Client) {
+		ci, _ := cid.Decode(test.SlowCid)
```
lanzafame: `test.SlowCid` doesn't exist. Related question, where do the `test.TestCid1`, etc, come from when running the tests?

api/rest/client/methods.go
```diff
@@ -251,12 +252,10 @@ func (sf *statusFilter) filter(ctx context.Context, fp StatusFilterParams) {
 				return
 			}
 
+			sf.Out <- gblPinInfo
```
lanzafame: 👍

api/rest/client/methods.go
```diff
@@ -191,18 +191,19 @@ func (c *Client) WaitFor(ctx context.Context, fp StatusFilterParams) (api.Global
...
 		case err := <-sf.Err:
 			return api.GlobalPinInfo{}, err
-		case <-sf.Done:
-			// drain status channel and return last status on channel
-			for {
-				if status, ok := <-sf.Out; !ok {
-					return status, nil
-				}
+		case st, ok := <-sf.Out:
+			if ok {
```
lanzafame: Could be simplified to:
```go
if !ok {
    return status, nil
}
status = st
```

ipfs-cluster-ctl/main.go
hsanjuan: did the first

hsanjuan:
> @hsanjuan Agreed. Could you help me see where I went wrong with the following sharness tests, as they shouldn't have passed.

The first case in a switch was `case !c.Bool("no-status")`. Since that flag is set to false by default, it waited 1 second and print a status (in which everything was pinned already because it's fast enough). So the output would look like you had been waiting.

While those tests verify that the flags are actually valid and usable and that we can call the command with them, they don't fully verify the behavior. I haven't come up with a straightforward way to fully verify `--wait` with sharness so I left those as they are. They still will catch some errors.

ipfs-cluster-ctl/main.go
```diff
@@ -712,3 +670,53 @@ func parseCredentials(userInput string) (string, string) {
 		return "", ""
 	}
 }
+
+func handlePinResponseFormatFlags(
+	c *cli.Context,
+	ci *cid.Cid,
+	target api.TrackerStatus,
+	timeout time.Duration,
```
lanzafame: We could access the timeout value via `cli.Context`?

ipfs-cluster-ctl/main.go
```diff
@@ -712,3 +670,53 @@ func parseCredentials(userInput string) (string, string) {
...
+	if c.Bool("wait") {
+		status, cerr = waitFor(ci, target, timeout)
+		checkErr("waiting for pin status", cerr)
+	}
+
+	if c.Bool("no-status") {
+		return
+	}
+
+	if status.Cid == nil { // no status from "wait"
```
lanzafame: Could this go before the `no-status` check as it is related to the `wait` flag?

lanzafame: Does having `--wait` and `--no-status` set at the same time make sense though?

lanzafame: sorry nvm, I totally not understanding that correctly

PR #368, opened 2018-03-28
Feat/sharding use pin feats

ZenGround0:
Same as PR #359 but merging to the main sharding branch.  Also corrected to keep the replication factor default equal to 0.


ZenGround0:
@hsanjuan Let me know if there is anything else you want changed on this PR before merging.  Thanks!

PR #385, opened 2018-04-20
Refactor add & general cleanup

ZenGround0:
addFile function is now a Cluster method accessed by RPC
residue from attempting to stream responses removed
ipfs-cluster-ctl ls bug fixed
problem with importer/add not printing resolved
new test now checks for this

License: MIT


ZenGround0:
@hsanjuan tests are on there way.  I'm planning to add them to another PR.

add.go
```diff
@@ -0,0 +1,231 @@
...
+	"net/url"
+	"strconv"
+
+	"github.com/ipfs/ipfs-cluster/api"
+	"github.com/ipfs/ipfs-cluster/importer"
+
+	"github.com/ipfs/go-ipfs-cmdkit/files"
+)
+
+func (c *Cluster) consumeLocalAdd(
```
lanzafame: @ZenGround0 Why not have `consumeLocalAdd` and `consumeShardAdd` return `(string, error)` and not pass `args` to those two functions? It seems unnecessary to pass the state round when it could just be handled within `consumeImport`

lanzafame: Actually, both `consumeLocalAdd` and `consumeShardAdd` could be slimmed down to just receive a `*api.NodeWithMeta` constructed outside of those two functions instead of inside.

add.go
```diff
@@ -0,0 +1,231 @@
...
+	// v0 (v1.cbor, v1.protobuf) we'll need to update this
+	outObj.Format = ""
+	args["cid"] = outObj.Cid // root node stored on last call
+	var hash string
+	err := c.rpcClient.Call(
+		"",
+		"Cluster",
+		"IPFSBlockPut",
+		*outObj,
+		&hash)
```
lanzafame: nitpick: put the closing `)` on the next line.

add.go
```diff
@@ -0,0 +1,231 @@
...
+	outObj.ID = shardID
+	outObj.ReplMax = replMax
+	outObj.ReplMin = replMin
+	var retStr string
+	err := c.rpcClient.Call(
+		"",
+		"Cluster",
+		"SharderAddNode",
+		*outObj,
+		&retStr)
```
lanzafame: nitpick: put the closing `)` on the next line.

api/rest/restapi.go
```diff
@@ -371,252 +368,35 @@ func (api *API) graphHandler(w http.ResponseWriter, r *http.Request) {
...
+	fI := types.FileInfo{
+		Reader: reader,
+		Params: queryValues,
+	}
+	var toPrint []types.AddedOutput
+	err = api.rpcClient.Call("",
+		"Cluster",
+		"AddFile",
+		fI,
+		&toPrint)
```
lanzafame: nitpick: put the closing `)` on the next line.

add.go
```diff
@@ -0,0 +1,231 @@
...
+
+	var consume func(map[string]string, *api.NodeWithMeta, int, int) error
+	var finish func(map[string]string, int, int) error
+	if shard == "true" {
+		consume = c.consumeShardAdd
+		finish = c.finishShardAdd
+	} else {
+		consume = c.consumeLocalAdd
+		finish = c.finishLocalAdd
+
```
lanzafame: blank line should be after the `}` maybe?

add.go
```diff
@@ -0,0 +1,231 @@
...
+	logger.Debugf("succeeding sharding import")
+	return toPrint, nil
+}
+
+// AddFile adds a file to the ipfs daemons of the cluster.  The ipfs importer
+// pipeline is used to DAGify the file.  Depending on input parameters this
+// DAG can be added locally to the calling cluster peer's ipfs repo, or
+// sharded across the entire cluster.
+func (c *Cluster) AddFile(
+	reader *multipart.Reader,
```
lanzafame: nitpick: I wouldn't have considered function declaration _that_ long, but changing this is up to you.

add.go
```diff
@@ -0,0 +1,231 @@
...
+		// Ensure we terminate reading from importer after cancellation
+		// but do not block
+		case <-ctx.Done():
+			err = errors.New("context timeout terminated add")
+			return nil, err
+		// Terminate session when importer throws an error
+		case err, ok := <-errChan:
+			if !ok {
+				openChs--
+				errChan = nil
```
lanzafame: Why do you set the `channels` to `nil`, if they have been closed by another goroutine? I am curious more than anything else.

add.go
ZenGround0: I agree you could do these things.  I went this route to make what is happening more explicit to the reader.  The string passed to `finish` has a totally different meaning depending on whether this is local or sharded.  Existing is probably overkill though and might over complicate things, I'll try to simplify.

add.go
ZenGround0: Good question.  My understanding is that closed channels are always available, and nil channels never are.  The `openChs` count requires listening on a closed channel exactly once.  See https://stackoverflow.com/questions/13666253/breaking-out-of-a-select-statement-when-all-channels-are-closed?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa for more.

add.go
ZenGround0: Yeah I get that this is the issue, @hsanjuan also did not like this.  I will refactor so that consume returns a string which I think appeals to both of your styles better. 

I'm not sure if I'm understanding your confusion properly but the reason we can't just send information into the consume function is that in general certain state (root cid/shard session id) must persist between calls to consume or finish.  True, we could read the outObj pointer after consume finished in consumeImport but then we need to read specific fields based on whether this is local or sharded call and that breaks the abstraction I was going for.

ZenGround0:
@hsanjuan add logic is now refactored to a subpackage and imported on api and cluster component to avoid pointers over RPC

api/types.go
```diff
@@ -749,6 +750,13 @@ func (pins PinSerial) ToPin() Pin {
 	}
 }
 
+// FileInfo carries a reader pointing to a file and the parameters specified
+// for adding the file to ipfs-cluster
+type FileInfo struct {
```
hsanjuan: This should not be necessary anymore?

importer/import.go
```diff
@@ -10,6 +10,13 @@ import (
 	"github.com/ipfs/go-ipfs-cmdkit/files"
 )
 
+func shouldIgnore(err error) bool {
+	if strings.Contains(err.Error(), "dagservice: block not found") {
```
hsanjuan: I am unable to find where this error string is generated. I would like that our comparison is something like `if err == module.ErrNotFound`. But I've been looking around for a while and I don't find the place where this error is constructed like this. Can you trace this? I only find references to `blockstore: block not found`.

ZenGround0: Wow this is hard to find.  I will let you know if I make any progress here.

ZenGround0: Omg it was me the whole time

ZenGround0:
@hsanjuan I think this is good to go, will you give it another look?

importer/import.go
hsanjuan: haha and I did not have the branch checked out so my greps wouldn't find it!

PR #386, opened 2018-04-20
Tests to front end of adding pipeline

ZenGround0:
api client test
api test
one sharness test
refactoring of testingData for access from other packages
always wrap files added by cluster
remove unused flag 'progress'
changes to support hidden flag

@hsanjuan -- will rebase on finished PR #385.  
Still needs more sharness tests and tests for the functions in add.go.

Let me know if you disagree with the policy to wrap all files added in one session under a common root.  It makes sense to me because the files will be sharded together in the sharding case so they are associated with the info in the wrapping root anyway.  Also let me know if you'd like more extensive testing of the api.  I tested to the level other api functions seem to be tested but could step this up.

License: MIT


ZenGround0:
@hsanjuan One thing that has come up in testing is a problem when a cid is pinned normally and with the sharding approach in the same cluster.  There are lots of solutions but I need a spec for what we want.  Do you prefer barring identical CIDs from being pinned in multiple ways (i.e. `ipfs-cluster-ctl add <file> && ipfs-cluster-ctl --shard add <file>` fails on the second command), or maintaining multiple pinning schemes for the same CID and changing the state interface (i.e. mapstate.Get(cid) now returns a list of pins)?

hsanjuan:
@ZenGround0 

> a problem when a cid is pinned normally and with the sharding approach in the same cluster

Let's jump in a call, or discuss on monday standup

api/rest/client/request.go
```diff
@@ -58,10 +58,8 @@ func (c *Client) doStreamRequest(method, path string, body io.Reader, headers ma
 		r.Header.Set(k, v)
 	}
 
-	// Here are the streaming specific modifications
-	// Using HTTP 2.0 to enable parallel reading req and writing resp
-	r.ProtoMajor = 2
-	r.ProtoMinor = 0
+	r.ProtoMajor = 1
+	r.ProtoMinor = 1
```
hsanjuan: what was the issue here for not maintaining http/2.0

api/rest/restapi_test.go
```diff
@@ -148,6 +155,53 @@ func TestAPIPeerAddEndpoint(t *testing.T) {
 	}
 }
 
+func TestAPIAddFileEndpoint(t *testing.T) {
```
hsanjuan: Either make 3 tests, or make 3 subtests (`t.Run()`)

Also, I think you need to rebase sharing on master. All the restapi tests are using this `testBothEndpoints` format that runs the test against both the libp2p and the http endpoints, so it should be used here too.

ipfscluster_test.go
```diff
@@ -1525,3 +1526,202 @@ func TestClustersGraphUnhealthy(t *testing.T) {
...
+			allStatus := c.tracker.StatusAll()
+			if len(allStatus) != numPins {
+				t.Errorf("Expected %d pins from local file add but got %d", numPins, len(allStatus))
+			}
+		}
+	}
+	runF(t, clusters, fPinnedThunk(rootCid, 1))
+
+	// Add the same file but as a trickle DAG
+	params["layout"] = []string{"trickle"}
```
hsanjuan: Make a different test, or subtest

ipfscluster_test.go
```diff
@@ -1525,3 +1526,202 @@ func TestClustersGraphUnhealthy(t *testing.T) {
...
+	params := make(map[string][]string)
+	params["shard"] = []string{"true"}
+	params["quiet"] = []string{"false"}
+	params["silent"] = []string{"false"}
+	params["layout"] = []string{""}
+	params["chunker"] = []string{""}
+	params["raw"] = []string{"false"}
+	params["hidden"] = []string{"false"}
+	params["replMin"] = []string{"-1"}
+	params["replMax"] = []string{"-1"}
```
hsanjuan: You may want to have a function to create a default params map

ipfscluster_test.go
```diff
@@ -1525,3 +1526,202 @@ func TestClustersGraphUnhealthy(t *testing.T) {
...
+
+	_, err = clusters[j].AddFile(multiPartR, params)
```
hsanjuan: I probably missed this in some other review. It looks suboptimal that AddFile receives a map of parameters of strings. We should have a type to wrap the parameters with the right types.

ipfscluster_test.go
```diff
@@ -1525,3 +1526,202 @@ func TestClustersGraphUnhealthy(t *testing.T) {
...
+
+	rootCid, _ := cid.Decode(test.TestDirBalancedRootCID)
+	m, err := mockFromCluster(clusters[j], clusters, mock)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	fShardsPinned := func(t *testing.T, c *Cluster) {
+		// Verify that consensus state has the expected pinset
+		pins := c.Pins()
```
hsanjuan: maybe you want to do PinGet(rootCid) directly, rather than looping all pins.

ipfscluster_test.go
```diff
@@ -1525,3 +1526,202 @@ func TestClustersGraphUnhealthy(t *testing.T) {
...
+		}
+		cdag := pins[iRoot].Clusterdag
+		if cdag == nil {
+			t.Fatalf("no clusterdag pinned for data root")
+		}
+
+		// Check that the clusterdag pointed to by data root is pinned
+		iCdag := -1
+		for i, pin := range pins {
+			if pin.Cid.Equals(cdag) {
```
hsanjuan: same comment, `PinGet`

ipfscluster_test.go
```diff
@@ -1525,3 +1526,202 @@ func TestClustersGraphUnhealthy(t *testing.T) {
...
+		}
+		cdagNode, err := sharder.CborDataToNode(cdagBytes, "cbor")
+		if err != nil {
+			t.Fatalf("cdag bytes cannot be parsed to ipld-cbor node")
+		}
+
+		// Check that all shards in the clusterDAG are pinned
+		for _, shardLink := range cdagNode.Links() {
+			iShard := -1
+			for i, pin := range pins {
```
hsanjuan: same

ipfscluster_test.go
```diff
@@ -1525,3 +1526,202 @@ func TestClustersGraphUnhealthy(t *testing.T) {
...
+	params["replMax"] = []string{"-1"}
+
+	_, err = clusters[j].AddFile(multiPartR, params)
+	if err != nil {
+		t.Error(err)
+	}
+	delay()
+
+	rootCid, _ := cid.Decode(test.TestDirBalancedRootCID)
+	m, err := mockFromCluster(clusters[j], clusters, mock)
```
hsanjuan: isn't the mock you are looking for simply `mock[j]` ?

api/rest/client/request.go
ZenGround0: I couldn't get simultaneous streaming of request and response working which was the whole reason I moved to 2.0 in the first place.  I could bring it back if you like but I couldn't think of a reason to keep it so went back to the standard choice.

GitCop:

 * Commit: 72ee994b8f492aaee0c3a28493fa70e6df08a880
    **License: MIT  
Signed-off-by:**

---

api/rest/client/request.go
hsanjuan: SGTM

api/rest/restapi.go
```diff
@@ -497,20 +497,32 @@ func (api *API) graphHandler(w http.ResponseWriter, r *http.Request) {
...
-	params := r.URL.Query()
+	urlParams := r.URL.Query()
+	layout := urlParams.Get("layout")
+	chunker := urlParams.Get("chunker")
+	raw, _ := strconv.ParseBool(urlParams.Get("raw"))
+	hidden, _ := strconv.ParseBool(urlParams.Get("hidden"))
+	silent, _ := strconv.ParseBool(urlParams.Get("silent"))
+	shard, _ := strconv.ParseBool(urlParams.Get("shard"))
+	replMin, _ := strconv.Atoi(urlParams.Get("repl_min"))
+	replMax, _ := strconv.Atoi(urlParams.Get("repl_max"))
```
hsanjuan: It might be worth doing error checking for these params, making sure they could be parsed and that they have allowable values.

api/types.go
```diff
@@ -494,31 +494,6 @@ func (addrsS MultiaddrsSerial) ToMultiaddrs() []ma.Multiaddr {
 	return addrs
 }
 
-
-// PeersToStrings IDB58Encodes a list of peers.
-func PeersToStrings(peers []peer.ID) []string {
```
hsanjuan: Where did these go? They are used in tests. Ah they were dups. ok

sharness/t0031-ctl-add.sh
```diff
@@ -26,7 +26,11 @@ test_expect_success IPFS,CLUSTER "add sharded small file to cluster" '
 '
 
 test_expect_success IPFS,CLUSTER "add same file sharded and unsharded" '
-    echo "complete ME"
+    output=`ipfs-cluster-ctl add --shard ../test_data/small_file | tail -1` &&
+    cid=${output:7:47} &&
```
hsanjuan: This looks weird

api/rest/restapi.go
```diff
@@ -497,24 +497,35 @@ func (api *API) graphHandler(w http.ResponseWriter, r *http.Request) {
...
 
-	params := r.URL.Query()
+	urlParams := r.URL.Query()
+	layout := urlParams.Get("layout")
+	chunker := urlParams.Get("chunker")
+	raw, _ := strconv.ParseBool(urlParams.Get("raw"))
+	hidden, _ := strconv.ParseBool(urlParams.Get("hidden"))
+	shard, _ := strconv.ParseBool(urlParams.Get("shard"))
+	replMin, _ := strconv.Atoi(urlParams.Get("repl_min"))
+	replMax, _ := strconv.Atoi(urlParams.Get("repl_max"))
```
hsanjuan: It might be worth doing error checking for these params, making sure they could be parsed and that they have allowable values.

sharness/t0031-ctl-add.sh
```diff
@@ -0,0 +1,40 @@
...
+    [[ -z "$(ipfs-cluster-ctl pin ls)" ]] &&
+    ipfs-cluster-ctl pin ls -a | grep -q "$cid" &&
+    [[ $(ipfs-cluster-ctl pin ls -a | wc -l) -eq "3" ]]  &&         
+    ipfs-cluster-ctl pin rm $cid &&
+    [[ -z "$(ipfs-cluster-ctl pin ls -a)" ]] 
+'
+
+test_expect_success IPFS,CLUSTER "add same file sharded and unsharded" '
+    output=`ipfs-cluster-ctl add --shard ../test_data/small_file | tail -1` &&
+    cid=${output:7:47} &&
```
hsanjuan: this looks weird.

ZenGround0: with the power of the new `--only-hashes` flag this will not longer need to look weird!

PR #387, opened 2018-04-22
Add disable_repinning cluster option

s1na:
This PR adds the option `disable_repinning` along with its default value, and test values. `repinFromPeer` returns if this option is true.

I'm having problems with running the tests to ensure everything is passing. I get `ipfscluster_test.go:69: bootstrap unsuccessful`, does anyone know what might be the cause?

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -397,6 +397,10 @@ func (c *Cluster) watchPeers() {
 
 // find all Cids pinned to a given peer and triggers re-pins on them.
 func (c *Cluster) repinFromPeer(p peer.ID) {
+	if c.config.DisableRepinning {
+		return
+	}
+
 	cState, err := c.consensus.State()
 	if err != nil {
 		logger.Warning(err)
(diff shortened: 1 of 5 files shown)


GitCop:

 * Commit: 19b6208c63ae0119f04530054f43830366e0d019
    **License: MIT  
Signed-off-by:**

---

ZenGround0:
Not sure what's causing your tests to fail locally but everything is passing on CI.  If you follow GitCop's instructions for a valid signoff then this LGTM.

s1na:
@ZenGround0 Okay thanks, I have fixed the commit's signature.

hsanjuan:
@s1na this looks good to me, do you think you can add a test? 

i.e. this test https://github.com/ipfs/ipfs-cluster/blob/master/ipfscluster_test.go#L1259 checks that re-allocation happens. Something similar (probably simpler) should check that re-allocation doesn't happen (setting DisableRepinning) at the beginning.

cluster.go
```diff
@@ -397,6 +397,10 @@ func (c *Cluster) watchPeers() {
 
 // find all Cids pinned to a given peer and triggers re-pins on them.
 func (c *Cluster) repinFromPeer(p peer.ID) {
+	if c.config.DisableRepinning {
+		return
```
hsanjuan: let's add a logger.Warn("repinning is disabled. Will not re-allocate cids from %s")

hsanjuan:
THANKS!

PR #393, opened 2018-04-26
Docs/move to website

### LINKED ISSUE #347: docs: clarify how to start a multi-node cluster
Just going to list out a few of the issues I hit and where I felt the docs fell down in assisting with troubleshooting them.

The first and main issue that I encountered was getting the ipfs-cluster-service nodes to be able to communicate with each other. 

This was partly due to me having the incorrect expectations about how the cluster would go about forming, basically, if they had the same `CLUSTER_SECRET` they would auto-discover each other similar to ipfs nodes, when within the same network, which they were in this case. 

Things that I found confusing about the cluster creation process:
- `ipfs-cluster-service init` cmd creates config directory and service.json file. There appears to be no way of providing a service.json file to `init`, which when using a docker container means you have to run the container so it runs `init` and then stopping the container and editing `service.json` and then starting the container again. (Note: I was able to do this by mounting named volumes).
- The chicken-and-egg aspect of starting a cluster isn't very clear in the documentation. I found this really confusing due to the aforementioned false expectation about member auto-discovery based on the shared `CLUSTER_SECRET`. To clarify what I mean by 'chicken-and-egg', the requirement to start one node then take that nodes multiaddress and add it to service.json of the other nodes as the bootstrap address (if doing dynamic cluster membership) 

The other thing is how the documentation is split over several documents and it doesn't always make sense why some are split and some are not.

hsanjuan:
This removes/updates all documentation in the repository to point to the website. I have moved everything there.

Fixes #347


lanzafame:
@hsanjuan I feel like the `CHANGELOG.md` and `CAPTAIN.LOG.md` should have some place still within the code base. Also, I am one who often works on stuff without an internet connection, so not having the development documentation (i.e. architecture docs) in the code base is a pain. I understand the desire for one canonical source but I feel in this case it might be more of a hindrance. 

One solution to this may be to adapt the `Makefile` in `ipfs-cluster-website` to extract the required Markdown files from the `ipfs-cluster` repo, this would mean that to publish the website you would need to download both repos but I think that is a small cost to pay to keep relevant information in the most appropriate places without having to deal with duplicate documentation sources. 

I am not entirely sold on the idea myself but feel it might be a good compromise.

lanzafame:
I should mention that I definitely think that the `News` and `Roadmap` sections should be moved to the website.

ZenGround0:
Piling on @lanzafame's point, I really like the architecture.md within the repo so that people who have downloaded the repo and want to learn about it without looking elsewhere have a good builtin jumping off point.  I am ok with more subject-to-change docs like CAPTAIN.LOG and more intensive tutorials like the cluster-guide living only on the website.

hsanjuan:
In terms of developer documentation, our `godocs` are the main source and those are still here. I partially agree that more architecture information would be useful but I'd like this to be part of the godocs for the main package (I'll submit another PR).

Regarding the Changelog, fine, we can keep it here.

Would this be ok as a compromise?

hsanjuan:
I took thumbs up as approval

PR #394, opened 2018-04-26
Docker: Run with daemon --upgrade by default.

hsanjuan:
Plus add some warnings for users running the container randomly.

License: MIT


docker/entrypoint.sh
```diff
@@ -20,10 +20,15 @@ ipfs-cluster-service --version
 if [ -e "${IPFS_CLUSTER_PATH}/service.json" ]; then
     echo "Found IPFS cluster configuration at ${IPFS_CLUSTER_PATH}"
 else
+    echo "This container only runs ipfs-cluster-service. ipfs needs to be run separately!"
```
ZenGround0: awesome

docker/entrypoint.sh
```diff
@@ -20,10 +20,15 @@ ipfs-cluster-service --version
...
+    echo "Initializing default configuration"
     ipfs-cluster-service init
     if [ -n "$IPFS_API" ]; then
+        echo "Using \$IPFS_API as value for node_multiaddress in configuration"
```
ZenGround0: Already great, but to make this crystal clear say "... as value for the *ipfs_connector* node_multiaddress".  Its hard for newcomers to keep track of all the different nodes and uses of the word node so calling out that this is the IPFS node's address will help those who don't want to dig into the config file.

PR #398, opened 2018-04-29
Emancipate the consensus component

### LINKED ISSUE #235: Re-introduce, configuration option shadowing
Starting cluster with `--bootstrap` or `--leave` will switch those options permanently in the configuration. `ipfs-cluster-service/main.go` should make sure they are restored before exiting.

The case of bootsrap is tricky, because cluster might have modified it too. We can leave that one. But it should happen at least for `leave_on_shutdown`.

### LINKED ISSUE #371: "peers" and "bootstrap" should be Raft configuration options
The consensus layer is in charge of maintaining the peerset and these options are pretty much specifying the peerset during the first boot.

My impression is that they should be part of the `raft` section, because if we were to change the consensus implementation for something else, we may want to provide the peerset in different ways (or not provide it at all).

That said, I need to look closely as what it would main to make this move, as maybe there are hard-dependencies on why peers need to stay at the top level.

### LINKED ISSUE #378: WaitForUpdates may return before state is fully synced
Per https://github.com/ipfs/ipfs-cluster/issues/160#issuecomment-380180327, I see that:

* WaitForUpdates has returned with index 321/321
* There are likely way more log entries in this Raft state
* WatchPeers auto-removes the bootstrapping peer after 5 seconds
* During those 5 seconds, the logs show calls to Track/Untrack, showing that possibly we are applying log entries during that time, even though the sync was finished.
* If not all log entries have been applied, this peer may not realize it has become part of the peerset because that log entry was not applied.
* User claims the peer state is clean on bootstrap (seems so)

### LINKED ISSUE #381: Peers with dynamic IPs
Cluster should support peers with dynamic IPs (docker and other deployment methods are likely to assign new IPs to things on reboot).

One way is to make sure that /dns/ multiaddresses work (that is, we remember them, and we don't record any IP addresses for those peers that we have dns values for). I am not sure, but I think we now resolve everything prior to adding it to the peerstore.

Ideally, we should support a bootstrapping of some form which just fetches peers addresses from a known node.

hsanjuan:
This commit promotes the Consensus component (and Raft) to become a fully
independent thing like other components, passed to NewCluster during
initialization. Cluster (main component) no longer creates the consensus
layer internally. This has triggered a number of breaking changes
that I will explain below.

Motivation: Future work will require the possibility of running Cluster
with a consensus layer that is not Raft. The "consensus" layer is in charge
of maintaining two things:
  * The current cluster peerset, as required by the implementation
  * The current cluster pinset (shared state)

While the pinset maintenance has always been in the consensus layer, the
peerset maintenance was handled by the main component (starting by the "peers"
key in the configuration) AND the Raft component (internally)
and this generated lots of confusion: if the user edited the peers in the
configuration they would be greeted with an error.

The bootstrap process (adding a peer to an existing cluster) and configuration
key also complicated many things, since the main component did it, but only
when the consensus was initialized and in single peer mode.

In all this we also mixed the peerstore (list of peer addresses in the libp2p
host) with the peerset, when they need not to be linked.

By initializing the consensus layer before calling NewCluster, all the
difficulties in maintaining the current implementation in the same way
have come to light. Thus, the following changes have been introduced:

* Remove "peers" and "bootstrap" keys from the configuration: we no longer
edit or save the configuration files. This was a very bad practice, requiring
write permissions by the process to the file containing the private key and
additionally made things like Puppet deployments of cluster difficult as
configuration would mutate from its initial version. Needless to say all the
maintenance associated to making sure peers and bootstrap had correct values
when peers are bootstrapped or removed. A loud and detailed error message has
been added when staring cluster with an old config, along with instructions on
how to move forward.

* Introduce a PeerstoreFile ("peerstore") which stores peer addresses: in
ipfs, the peerstore is not persisted because it can be re-built from the
network bootstrappers and the DHT. Cluster should probably also allow
discoverability of peers addresses (when not bootstrapping, as in that case
we have it), but in the meantime, we will read and persist the peerstore
addresses for cluster peers in this file, different from the configuration.
Note that dns multiaddresses are now fully supported and no IPs are saved
when we have DNS multiaddresses for a peer.

* The former "peer_manager" code is now a pstoremgr module, providing utilities
to parse, add, list and generally maintain the libp2p host peerstore, including
operations on the PeerstoreFile. This "pstoremgr" can now also be extended to
perform address autodiscovery and other things indepedently from Cluster.

* Create and initialize Raft outside of the main Cluster component: since we
can now launch Raft independently from Cluster, we have more degrees of
freedom. A new "staging" option when creating the object allows a raft peer to
be launched in Staging mode, waiting to be added to a running consensus, and
thus, not electing itself as leader or doing anything like we were doing
before. This additionally allows us to track when the peer has become a
Voter, which only happens when it's caught up with the state, something that
was wonky previously.

* The raft configuration now includes an InitPeerset key, which allows to
provide a peerset for new peers and which is ignored when staging==true. The
whole Raft initialization code is way cleaner and stronger now.

* Cluster peer bootsrapping is now an ipfs-cluster-service feature. The
--bootstrap flag works as before (additionally allowing comma-separated-list
of entries). What bootstrap does, is to initialize Raft with staging == true,
and then call Join in the main cluster component. Only when the Raft peer
transitions to Voter, consensus becomes ready, and cluster becomes Ready.
This is cleaner, works better and is less complex than before (supporting
both flags and config values). We also backup and clean the state whenever
we are boostrapping, automatically

* ipfs-cluster-service no longer runs the daemon. Starting cluster needs
now "ipfs-cluster-service daemon". The daemon specific flags (bootstrap,
alloc) are now flags for the daemon subcommand. Here we mimic ipfs ("ipfs"
does not start the daemon but print help) and pave the path for merging both
service and ctl in the future.

While this brings some breaking changes, it significantly reduces the
complexity of the configuration, the code and most importantly, the
documentation. It should be easier now to explain the user what is the
right way to launch a cluster peer, and more difficult to make mistakes.

As a side effect, the PR also:

* Fixes #381 - peers with dynamic addresses
* Fixes #371 - peers should be Raft configuration option
* Fixes #378 - waitForUpdates may return before state fully synced
* Fixes #235 - config option shadowing (no cfg saves, no need to shadow)


Makefile
```diff
@@ -96,15 +96,15 @@ publish: rwundo
...
-	@docker run --name tmp-make-cluster -d cluster-image && sleep 8
-	@docker exec tmp-make-cluster sh -c "ipfs-cluster-ctl version"
-	@docker exec tmp-make-cluster sh -c "ipfs-cluster-service -v"
-	@docker stop tmp-make-cluster && docker rm tmp-make-cluster
-	@docker build -t cluster-image -f Dockerfile-test .
-	@docker run --name tmp-make-cluster -d cluster-image && sleep 8
-	@docker exec tmp-make-cluster sh -c "ipfs-cluster-ctl version"
-	@docker exec tmp-make-cluster sh -c "ipfs-cluster-service -v"
-	@docker stop tmp-make-cluster && docker rm tmp-make-cluster
+	docker build -t cluster-image -f Dockerfile .
```
ZenGround0: Is this just for debugging or do you want this output displayed?

cluster.go
```diff
@@ -17,10 +18,11 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 )
 
-// Common errors
-var (
-	ErrBootstrap = errors.New("bootstrap unsuccessful")
-)
+// ReadyTimeout specifies the  before giving up
```
ZenGround0: "specifies the  before" -> "specifies the time before"

cluster.go
```diff
@@ -17,10 +18,11 @@ import (
...
+// during startup (waiting for consensus to be ready)
+// It may need adjustment according to timeouts in the
+// consensus layer.
+var ReadyTimeout = 30 * time.Second
```
ZenGround0: Would it make sense to have this configurable?  Depending on latency between nodes certain clusters might always want a higher readytimeout.

cluster.go
```diff
@@ -64,7 +66,7 @@ type Cluster struct {
 func NewCluster(
 	host host.Host,
 	cfg *Config,
-	consensusCfg *raft.Config,
+	consensus Consensus,
```
ZenGround0: This is really happening \o/ 🎉

cluster.go
```diff
@@ -439,18 +400,23 @@ func (c *Cluster) ready(timeout time.Duration) {
...
-  - Check that the peer you are trying to connect to is running the
+  - Check that the peer(s) you are trying to connect to is running the
     same version of IPFS-cluster.
 **************************************************
 `)
 		c.Shutdown()
 		return
 	case <-c.consensus.Ready():
+		// Consensus ready means the state is up to date so we can sync
+		// it to the tracker.We ignore errors (normal when state doesn't
```
ZenGround0: typo: space after the "."

consensus/raft/config.go
```diff
@@ -48,6 +51,11 @@ type Config struct {
 	CommitRetries int
 	// How long to wait between retries
 	CommitRetryDelay time.Duration
+	// InitPeerset provides the list of initial cluster peers for new
+	// peers (with no prior state). It is when Raft was already initialized
+	// or when when starting in staging mode, before adding the peer
```
ZenGround0: typo "when when".  Also I don't understand the last sentence.  What 'is when Raft was already initialized?'

ZenGround0:
This is really exciting.  I'll try to do a proper review before next week.

cluster.go
hsanjuan: Note it is configurable, in a sense. cluster-service sets it in relation to wait_for_leader_timeout. Exposing it in config complicates the config and allows misconfiguring it, so I left it.

Before it was exacty the same, but now the main component doesn't have access to the consensusCfg.

pstoremgr/pstoremgr.go
```diff
@@ -0,0 +1,232 @@
...
+		host:          h,
+		peerstorePath: peerstorePath,
+	}
+}
+
+// ImportPeer adds a new peer address to the host's peerstore, optionally
+// dialing to it. It will resolve any DNS multiaddresses before adding them.
+// The address is expected to include the /ipfs/<peerID> protocol part.
+func (pm *Manager) ImportPeer(addr ma.Multiaddr, connect bool) error {
+	if pm.host == nil {
```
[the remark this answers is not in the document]
hsanjuan: you can initialize with a nil host if you just want peerstore parsing capabilities.

ipfs-cluster-service/state.go
```diff
@@ -159,3 +163,8 @@ func exportState(state *mapstate.MapState, w io.Writer) error {
 	enc.SetIndent("", "    ")
 	return enc.Encode(pinSerials)
 }
+
+// CleanupState cleans the state
+func cleanupState(cCfg *raft.Config) error {
```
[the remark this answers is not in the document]
hsanjuan: The `state.go` file (state upgrades in general), are very raft specific. Something we'll have to improve, but not on the scope of this PR. I'm just putting that method there as a reminder that everything in that file will need to be generalized.

consensus/raft/raft.go
```diff
@@ -29,7 +30,7 @@ var errWaitingForSelf = errors.New("waiting for ourselves to depart")
 // folder.
 // TODO: Maybe include this in Config. Not sure how useful it is to touch
```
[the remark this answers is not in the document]
hsanjuan: Ops, I didn't want to change that one actually, I meant `RaftDataBackupKeep`. Good catch.

consensus/raft/raft.go
```diff
@@ -47,156 +48,196 @@ var waitForUpdatesInterval = 100 * time.Millisecond
...
 	boltdb        *raftboltdb.BoltStore
+	staging       bool
 }
 
-// newRaft launches a go-libp2p-raft consensus peer.
-func newRaftWrapper(peers []peer.ID, host host.Host, cfg *Config, fsm hraft.FSM) (*raftWrapper, error) {
+// newRaftWrapper creates a Raft instance and initializes
+// everything leaving it ready to use. Note, that Bootstrap() should be called
+// to make sure the raft instance is usable.
+func newRaftWrapper(
```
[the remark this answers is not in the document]
hsanjuan: At somewhere between 80 and my current editor window length. Things start bothering me over 80 chars. Things bother me a lot when they are over 80 chars and reach the window border (which might be at 81, or at 120 or more, depending on the moment). When something bothers me a lot if probably gets multilined. Does that answer ? :)

consensus/raft/data_helper.go
```diff
@@ -6,24 +6,22 @@ import (
...
 
-// RaftDataBackupKeep indicates the number of data folders we keep around
-// after consensus.Clean() has been called.
-var RaftDataBackupKeep = 5
-
 // dataBackupHelper helps making and rotating backups from a folder.
 // it will name them <folderName>.old.0, .old.1... and so on.
 // when a new backup is made, the old.0 is renamed to old.1 and so on.
-// when the RaftDataBackupKeep number is reached, the last is always
+// when the "keep" number is reached, the last is always
```
lanzafame: s/last/oldest

hsanjuan:
Fixed and squashed @lanzafame

Issue #287, opened 2018-01-09
Use OpenTracing to gather valuable execution information and metrics


hsanjuan:
This would be really useful for cluster, to get a feel of which code paths are the slowest and when.

@frrist is working on adding OpenTracing to go-ipfs. We can probably take advantage of this because it's a really good thing to have.

See some work around it:

https://github.com/ipfs/go-ipfs/issues/4487
https://github.com/ipfs/go-ipfs-cmds/pull/39
https://github.com/ipfs/go-ipfs/pull/4506

First work could concentrate in gathering traces for a "pin" request for example.

Once that works, we can start adding tracing for other requests and operations (particularly the recurrent ones i.e. syncs, metric gathering etc).

frrist:
@hsanjuan I have created an [issue here](https://github.com/ipfs/notes/issues/277) to provide a general overview of the OpenTracing API (and a few discussion points) - any feedback would be appreciated!

s1na:
@hsanjuan I would be interested in working on this, however the tag suggests this to be a hard task. Do you think I can take a stab? and if so, is it okay if I do it in smaller steps?

hsanjuan:
@s1na It would be amazing if you worked on this. I suggest you start by adding tracing for a small thing (like the `ID` method, and once we have a handle on that we can think of doing more interesting traces like the whole `Pin` operations.

hsanjuan:
@s1na I saw you asked on IRC about how to import some things. You should not need to import anything new in terms of dependencies. According to @frrist, the last version of go-log (as of current master), supports open-tracing out of the box

https://godoc.org/github.com/ipfs/go-log#EventLogger

This is very new and @ffrist was going to post here with some details.

You can also check the PR in question: https://github.com/ipfs/go-log/pull/20/files . Note that godoc is no help in terms of documentation, because the methods which are documented are part of a non-exported type, but the comments in the source code seem useful.

s1na:
@hsanjuan Oh, sorry for my misunderstanding. I thought you meant we should also develop something similar.
[...]

PR #401, opened 2018-05-02
[WIP] Add basic tracing

s1na:
This PR tries to add tracing by using `go-log`'s `EventLogger`. As of yet, it only adds a few sample spans, and mostly traces are isolated to their components.

Notes:
* `EventLogger` modifications required by this PR have been merged into master, but not yet published to gx, therefore please make sure you use the master branch of `go-log` for testing.
* Tracing is enabled by default, and the spans are printed to stdout as soon as they're produced. `--trace-out` flag for both `ctl` and `service` allow one to redirect tracing output to a file.
* This patch passes the context from `ctl` to the rpc client as config which gets stored on the struct. This is bad practice, in future, ctx should be rather passed to every function.
* For some reason `ctl` doesn't print the root span, which I traced to `go-log/writer.logRoutine`, probably the channels get closed somewhere. To be fixed.

Implements #287

### DIFF
--- a/api/rest/client/client.go
+++ b/api/rest/client/client.go
@@ -73,6 +73,8 @@ type Config struct {
 
 	// LogLevel defines the verbosity of the logging facility
 	LogLevel string
+
+	Context context.Context
 }
 
 // Client provides methods to interact with the ipfs-cluster API. Use
@@ -91,6 +93,9 @@ type Client struct {
 // NewClient initializes a client given a Config.
 func NewClient(cfg *Config) (*Client, error) {
 	ctx := context.Background()
+	if cfg.Context != nil {
+		ctx = cfg.Context
+	}
 	client := &Client{
 		ctx:    ctx,
 		config: cfg,
--- a/ipfs-cluster-ctl/main.go
+++ b/ipfs-cluster-ctl/main.go
@@ -15,6 +15,7 @@ import (
 
 	cid "github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log"
+	writer "github.com/ipfs/go-log/writer"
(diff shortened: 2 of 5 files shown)


ipfs-cluster-ctl/main.go
```diff
@@ -126,11 +127,38 @@ requires authorization. implies --https, which you can disable with --force-http
...
+			Usage: "output path for tracing result. If not provided, results will be written to std.",
+		},
 	}
 
+	rootCtx := context.Background()
+
 	app.Before = func(c *cli.Context) error {
 		cfg := &client.Config{}
 
+		rootSpanName := "ctl"
```
hsanjuan: Probably put all tracing setup in its own function, but please disable by default.

ipfs-cluster-service/main.go
```diff
@@ -445,6 +451,14 @@ func run(c *cli.Context) error {
 func daemon(c *cli.Context) error {
 	logger.Info("Initializing. For verbose output run with \"-l debug\". Please wait...")
 
+	writerCloser := os.Stdout
+	if c.String("trace-out") != "" {
```
hsanjuan: Also give its own function, disable by default

api/rest/client/client.go
```diff
@@ -91,6 +93,9 @@ type Client struct {
 // NewClient initializes a client given a Config.
 func NewClient(cfg *Config) (*Client, error) {
```
lanzafame: Instead of adding a `context.Context` to the `Config` struct, I suggest providing it to the `Client` via a parameter here. Putting it in `Config` is a round-about way of setting a private field and `Context` is out of place with the other fields of `Config`.

hsanjuan:
Superseeded by #486

PR #406, opened 2018-05-08
[WIP] Allow to use the global flags before or after a subcommand name

MichaelMure:
When testing `ipfs-cluster` I had trouble using the command line. After reading the help text, I tried to put the global flags at the end of the command and got a unhelpful message:

```
$ ipfs-cluster-ctl id --host "/ip4/127.0.0.1/tcp/9094"
Incorrect Usage: flag provided but not defined: -host
...
```

The problem is that the `cli` library expect the user to run `ipfs-cluster-ctl --host "/ip4/127.0.0.1/tcp/9094" id` instead (https://github.com/urfave/cli/issues/585).

This PR fix the issue with the workaround suggested in the linked bug, that is reusing the global flags for each subcommand and using a `Before` to parse them.

Is that something you would accept ?

What is left to do:

- `ipfs-cluster-service` is covered yet
- there is a conflict with `health graph -f` and the global `-f` flag. In this case, the `cli` library panic.
- there is a similar conflict in `ipfs-cluster-service` with `state export -f` and the global `-f`.

### DIFF
--- a/ipfs-cluster-ctl/main.go
+++ b/ipfs-cluster-ctl/main.go
@@ -78,12 +78,7 @@ func checkErr(doing string, err error) {
 }
 
 func main() {
-	app := cli.NewApp()
-	app.Name = programName
-	app.Usage = "CLI for IPFS Cluster"
-	app.Description = Description
-	app.Version = Version
-	app.Flags = []cli.Flag{
+	globalFlags := []cli.Flag{
 		cli.StringFlag{
 			Name:  "host, l",
 			Value: defaultHost,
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 63410ed3e3c9f7090d576e8483f395a70315723e
    **License: MIT  
Signed-off-by:**

---

hsanjuan:
Hey! Thanks but I lean towards not supporting this. Usage says:

` ipfs-cluster-ctl [global options] command [command options] [arguments...]`

and `--host` is clearly under the global options list. 

This makes all global flags local and that has implications. ie. While this now allows passing all flags at the end, it is not parsing global flags anymore (any flags coming before the commands). That is a no-go for me.

Suddenly it also becomes difficult to make distinctions between local and global (`c.Global`) and the subcommand help's option-list gets polluted with global flags. In the end this is a UI choice made by `urfave/cli` and I'm not sure I want to spend time and code working around it. I feel it should be a feature in that library to allow parsing global flags anywhere in the command, not a workaround in our code.

@lanzafame views?

lanzafame:
I am in agreeance with @hsanjuan. I view the ordering in terms of setting up the context of the following commands. --host isn't a parameter of the id command but sets the contract in which it runs. Not really a solution but a helpful way of thinking about the global flags

MichaelMure:
Fair enough :) 

That said, as this construct is not that common, it's easy to overlook this detail and wonder why the option didn't work, if we missed a quote or something. The message isn't really helpful.

I think the help text is good enough, but I would suggest to use the `Before` action of the subcommands to detect when a global flag is used at the wrong place and output an explanation. Obviously that wouldn't work when there is a conflict but the long option name can be covered.

Just a suggestion though, feel free to discard ;)

hsanjuan:
If you want to improve the error message, you provide a custom function https://godoc.org/github.com/urfave/cli#OnUsageErrorFunc . If the error is about `flag provided but not defined`, you can explain the user in better terms.

Context: https://github.com/urfave/cli/blob/a1c7408de3f632d86eee604a3bb755f1ffb68226/command.go#L123
And: https://github.com/urfave/cli/blob/119bb6564841921ce6f1401e0f5d75317bdd9f4d/app.go#L211

I'd be happy to take improvements on this front.

lanzafame:
@MichaelMure are you happy for this PR to be closed and for this to be tracked by the urfave/cli#744 issue?

MichaelMure:
@lanzafame Sure, I was just waiting for you to ack the update.

PR #415, opened 2018-05-11
Fix #408: Race condition when StateSync runs

### LINKED ISSUE #408: TestClustersPin fails too much
We have a problem with this test and it's not just matter of increasing delays for a few seconds.

It fails a lot in Jenkins, it fails sometimes in Travis, and I don't manage to reproduce it locally.

https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fipfs-cluster/detail/PR-400/21/tests

I think it may come from the pinTracker merge, but I'll need to spend some time figuring out where things get struck.

cc @lanzafame

hsanjuan:
When StateSync() runs and triggers Untrack() on items
that have just been removed from the state but on which
Untrack() is underway, the operation tracker would be
reset to phase queued and in some cases stay so.

Also happened for Track()

This caused failures of TestClustersPin as SyncStatus()
is triggered regularly while Tracks() and Untracks() happen.

License: MIT

### DIFF
--- a/ipfscluster_test.go
+++ b/ipfscluster_test.go
@@ -403,16 +403,11 @@ func TestClustersPin(t *testing.T) {
 		}
 	}
 	delay()
-	delay()
-	delay()
-	delay()
 	fpinned := func(t *testing.T, c *Cluster) {
 		status := c.tracker.StatusAll()
 		for _, v := range status {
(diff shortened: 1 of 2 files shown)


hsanjuan:
Disregard Windows failures in jenkins

PR #422, opened 2018-05-15
WIP: Stateless pintracker implementation

lanzafame:
@hsanjuan Please review the current state of this PR.


optracker/operationtracker.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// GetAll gets all inflight operations.
+func (opt *OperationTracker) GetAll() []Operation {
+	var ops []Operation
+	opt.mu.RLock()
+	for _, op := range opt.operations {
+		ops = append(ops, op)
+	}
+	opt.mu.RUnlock()
```
hsanjuan: Can we set this as `defer opt.mu.Runlock()` after locking?

optracker/operationtracker.go
```diff
@@ -0,0 +1,221 @@
...
+				ops = append(ops, op)
+			}
+		case Phase:
+			if op.Phase == filter {
+				ops = append(ops, op)
+			}
+		default:
+		}
+	}
+	opt.mu.RUnlock()
```
hsanjuan: same

pintracker/maptracker/maptracker.go
```diff
@@ -271,50 +273,50 @@ func (mpt *MapPinTracker) unpin(c api.Pin) error {
...
+				optracker.OperationUnpin,
 			)
 			mpt.unpin(c)
 		}
 		mpt.set(c.Cid, api.TrackerStatusRemote)
 		return nil
 	}
 
-	mpt.optracker.trackNewOperation(mpt.ctx, c.Cid, operationPin)
+	if opc, ok := mpt.optracker.Get(c.Cid); ok {
```
hsanjuan: Note sure I agree with the re-ordering. i.e. to parallel calls to Track might run NewOperation (both) and trigger two unpins where there should only be one.

pintracker/ptutil/pin.go
```diff
@@ -0,0 +1,28 @@
+package ptutil
```
hsanjuan: we can call it util, since util submodules are `util` in other places.

pintracker/ptutil/pin.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"github.com/ipfs/ipfs-cluster/api"
+
+	logging "github.com/ipfs/go-log"
+	peer "github.com/libp2p/go-libp2p-peer"
+)
+
+var logger = logging.Logger("pintrackerutil")
```
hsanjuan: I think go-log wont truncates long name. Might wanna go with `pintrckutil` or something shorter.

Actually, It doesn't make me super happy to have more logging facilities around. I would just not log anything here. Perhaps we can simply not log that message or log it somewhere else. Otherwise, this need to be added to the list of facilities in `logging.go`

pintracker/statelesstracker/config.go
```diff
@@ -0,0 +1,86 @@
+package statelesstracker
+
+import (
+	"encoding/json"
+	"errors"
+
+	"github.com/ipfs/ipfs-cluster/config"
+)
+
+const configKey = "statelesstracker"
```
hsanjuan: `stateless` works, I think. Note that it already goes inside a `pin_tracker` config section.

pintracker/statelesstracker/statelesstracker.go
```diff
@@ -0,0 +1,380 @@
...
+	cid "github.com/ipfs/go-cid"
+	logging "github.com/ipfs/go-log"
+	peer "github.com/libp2p/go-libp2p-peer"
+)
+
+var logger = logging.Logger("pintracker")
+
+// StatelessPinTracker uses the optracker.OperationTracker to manage
+// transitioning shared ipfs-cluster state (Pins) to the local IPFS node.
+type StatelessPinTracker struct {
```
hsanjuan: If we rename the module to `stateless` and this to `Tracker` things sound good: `stateless.Tracker`.

pintracker/statelesstracker/statelesstracker.go
```diff
@@ -0,0 +1,380 @@
...
+	pinCh   chan api.Pin
+	unpinCh chan api.Pin
+
+	shutdownMu sync.Mutex
+	shutdown   bool
+	wg         sync.WaitGroup
+}
+
+// NewStatelessPinTracker creates a new StatelessPinTracker.
+func NewStatelessPinTracker(cfg *Config, pid peer.ID) *StatelessPinTracker {
```
hsanjuan: `func New()`

pintracker/statelesstracker/statelesstracker.go
```diff
@@ -0,0 +1,380 @@
...
+	spt.wg.Wait()
+	spt.shutdown = true
+	return nil
+}
+
+// Track tells the StatelessPinTracker to start managing a Cid,
+// possibly triggering Pin operations on the IPFS daemon.
+func (spt *StatelessPinTracker) Track(c api.Pin) error {
+	logger.Debugf("tracking %s", c.Cid)
+	if ptutil.IsRemotePin(c, spt.peerID) {
```
hsanjuan: again, unsure about checking remote before checking for ongoing operations

pintracker/statelesstracker/statelesstracker.go
```diff
@@ -0,0 +1,380 @@
...
+		logger.Error(err.Error())
+		return err
+	}
+	return nil
+}
+
+// StatusAll returns information for all Cids pinned to the local IPFS node.
+func (spt *StatelessPinTracker) StatusAll() []api.PinInfo {
+	// get statuses from ipfs node first
+	localpis, _ := spt.ipfsStatusAll()
```
hsanjuan: ipfsStatusAll receives a map, which is converted to a slice, which is here converted again to a map, which is then converted to a slice. There should be just one map->slice operation.

pintracker/statelesstracker/statelesstracker.go
```diff
@@ -0,0 +1,380 @@
...
+	)
+	if err != nil {
+		return err
+	}
+
+	spt.optracker.Finish(c.Cid)
+	return nil
+}
+
+func (spt *StatelessPinTracker) syncStatus(c *cid.Cid, ips api.IPFSPinStatus) api.PinInfo {
```
hsanjuan: // TODO remove?

pintracker/ptutil/pin.go
lanzafame: Or just return `(bool, error)`?

lanzafame: scratch that idea...

pintracker/maptracker/maptracker.go
lanzafame: Hmm yeah not sure what happened there. I fix that up.

lanzafame:
@hsanjuan so as it stands I can't seem to figure why the `cluster_test.go:TestClusterRecoverAllLocal` test is not passing for the `StatelessPinTracker`. I have added the `pintracker/pintracker_test.go` tests which only test the public interface of the `PinTracker`, I want to move the `TrackUntrack` et al tests over from the `maptracker` tests as well. Basically, the tests in mappintracker and stateless pkgs will only tests unexported functionality, that we want to test. Adding the `pintracker/pintracker_test.go` was good for understanding the differences between the two implementations. I think this might also be a good way to get around running the `basicmon` vs `pubsubmon` tests issue and having to run each implementation separately. Thoughts?

pintracker/statelesstracker/statelesstracker.go
```diff
@@ -0,0 +1,510 @@
...
+type Tracker struct {
+	config *Config
+
+	optracker *optracker.OperationTracker
+
+	opErrsMu sync.RWMutex
+	opErrs   map[string]api.PinInfo
+
+	remotesMu sync.RWMutex
+	remotes   map[string]api.PinInfo
```
hsanjuan: We shouldn't store `remotes` as it can potentially grow very big (in the same order as the original state that we are trying to remove)

This has consequences (cannot distinguish between something `unpinned` or `remote`). Perhaps this is fine.

pintracker/statelesstracker/statelesstracker.go
```diff
@@ -0,0 +1,510 @@
...
+	select {
+	case spt.unpinCh <- api.PinCid(c):
+	default:
+		err := errors.New("unpin queue is full")
+		spt.setError(c, err)
+		logger.Error(err.Error())
+		return err
+	}
+	return nil
+}
```
hsanjuan: Note: `Track` and `Untrack` are almost identical. First, if it's code doing exactly the same they should be identical. Second, probably worth refactoring. Perhaps it's worth having a `base` tracker that implements partial functionality for both since they share a lot of code.

pintracker/statelesstracker/statelesstracker.go
```diff
@@ -0,0 +1,510 @@
...
+// SyncAll verifies that the statuses of all tracked Cids (from the shared state)
+// match the one reported by the IPFS daemon. If not, they will be transitioned
+// to PinError or UnpinError.
+//
+// SyncAll returns the list of local status for all tracked Cids which
+// were updated or have errors. Cids in error states can be recovered
+// with Recover().
+// An error is returned if we are unable to contact the IPFS daemon.
+func (spt *Tracker) SyncAll() ([]api.PinInfo, error) {
+	pis := spt.getErrorsAll()
```
hsanjuan: You should actually make sure that 

1) There are no ongoing Pin operations for things that are already `pinned` in ipfs
2) No ongoing Unpin operations for things already `unpinned` in ipfs
3) No errors for things which have no error (they became pinned or unpinned)

pintracker/statelesstracker/statelesstracker.go
```diff
@@ -0,0 +1,510 @@
...
+// the IPFS daemon. If not, it will be transitioned
+// to PinError or UnpinError.
+//
+// Sync returns the updated local status for the given Cid.
+// Pins in error states can be recovered with Recover().
+// An error is returned if we are unable to contact
+// the IPFS daemon.
+func (spt *Tracker) Sync(c *cid.Cid) (api.PinInfo, error) {
+	// no-op in stateless tracker implementation
+	return spt.Status(c), nil
```
hsanjuan: Same as above

util.go
```diff
@@ -69,7 +69,9 @@ func pinInfoSliceToSerial(pi []api.PinInfo) []api.PinInfoSerial {
 	return pis
 }
 
-func globalPinInfoSliceToSerial(gpi []api.GlobalPinInfo) []api.GlobalPinInfoSerial {
+// GlobalPinInfoSliceToSerial is a helper function for serializing a slice of
+// api.GlobalPinInfos.
+func GlobalPinInfoSliceToSerial(gpi []api.GlobalPinInfo) []api.GlobalPinInfoSerial {
```
hsanjuan: It does not need exporting, does it ? 

If exported, it would fit better as part of `api`.

lanzafame:
@hsanjuan question regarding `api.PinInfo`, do we use the `TS` field anywhere for anything in particular?

hsanjuan:
@lanzafame just to show the user when the PinInfo was last updated. Nothing else.

lanzafame:
Closed as #460 replaces this PR.

PR #427, opened 2018-05-24
ipfshttp: Use custom http clients for pin/unpin

lanzafame:
so that the IPFSRequestTimeout is triggered on all other requests
other than those two. Timing out is handled by the context timeout
in the pin and unpin functions.

License: MIT


lanzafame:
Sounds good 👍

PR #429, opened 2018-05-24
StateSync(): some improvements

hsanjuan:
This commit:

* Does not collect and return changed items when doing StateSync (they are
not used)
* Removes the StateSync RPC method (no longer used)
* Uses tracker.StatusAll() rather than requesting Status on each Cid (should
be faster with upcoming pintracker)
* Does not launch a go-routine to track every item. Track is an async
operation. This likely causes 1000s goroutines to be started with no good
reason.

License: MIT

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -726,31 +726,36 @@ func (c *Cluster) Join(addr ma.Multiaddr) error {
 }
 
 // StateSync syncs the consensus state to the Pin Tracker, ensuring
-// that every Cid that should be tracked is tracked. It returns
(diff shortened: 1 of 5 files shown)


cluster.go
```diff
@@ -728,29 +728,34 @@ func (c *Cluster) Join(addr ma.Multiaddr) error {
 // StateSync syncs the consensus state to the Pin Tracker, ensuring
 // that every Cid that should be tracked is tracked. It returns
```
lanzafame: Docstring needs updating.

PR #436, opened 2018-05-26
fix #428: Use a regexp to match for indirect status in IPFSPinStatusF…

### LINKED ISSUE #428: IPFSPinStatusFromString() will not work with indirect pins
https://github.com/ipfs/ipfs-cluster/blob/master/api/types.go#L105

The type in indirect pins is not `indirect` but `indirect through <cid>`.  So this needs a small modification in order to work for indirect pins (we currently don't use those for anything so no big deal).

Laevos:
…romString()

Since indirect pins are of the form `indirect through <cid>`, let's use
a regexp to match them instead of an equality operator.

License: MIT
Signed-off-by: Lilith McMullen <iggnsthe@live.com>

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -10,10 +10,10 @@ package api
 
 import (
 	"fmt"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
-	"regexp"
 
 	cid "github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log"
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 2429c861f3770cb5d0fa636bfd8c96c7abd84e10
    **License: MIT  
Signed-off-by:**

---

PR #441, opened 2018-05-27
Fix #346: Enable NAT hole punching for libp2p host

### LINKED ISSUE #346: ipfs-cluster libp2p host does not enable NAT support
We're going to leave this there until we update libp2p again so that this can be used: https://github.com/libp2p/go-libp2p/pull/293

hsanjuan:
Fixes #346


hsanjuan:
To be rebased on top of #440

PR #455, opened 2018-06-01
Fix: API constants should explicit their type

hsanjuan:
Otherwise, they are int. This forces API users to do manual variable
type declaration when initializing something to one of these constants

License: MIT


hsanjuan:
Same way I've missed it for 2 years...

PR #458, opened 2018-06-04
Issue #340: fix some data races

### LINKED ISSUE #340: Run tests with -race
We have some races for sure, even if not critical. We should fix them so that tests pass with `-race`.

hsanjuan:
~~As pointed out in #340, still some failures coming from go-logging.~~

This fixes some of the data races identified when running with `-race`. Unfortunately, there are still data races in libp2p that we cannot take care of for the moment.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -406,7 +406,9 @@ This might be due to one or several causes:
 	}
 
 	close(c.readyCh)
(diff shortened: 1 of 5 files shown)


hsanjuan:
@lanzafame should merge these fixes, even without enabling -race. better than nothing.

PR #460, opened 2018-06-08
implements a stateless pintracker

lanzafame:
Also updates to the optracker to make retrieving information easier.

This PR replaces #422.

---
This change is [Reviewable](https://reviewable.io/reviews/ipfs/ipfs-cluster/460)


pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,489 @@
...
+	var gpin api.PinSerial
+	err := spt.rpcClient.Call(
+		"",
+		"Cluster",
+		"PinGet",
+		api.PinCid(c).ToSerial(),
+		&gpin,
+	)
+	if err != nil {
+		if _, ok := err.(*api.CidNotInGlobalStateError); ok {
```
hsanjuan: As a note, the fact that this works is because of the local-rpc-shortcut part of gorpc. I think the right way to fix this is on `gorpc`, declaring custom error types there are having a function like `gorpc.IsRPCError(err)` to identify them.

Mind going that way instead?

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,489 @@
...
+func (spt *Tracker) getError(c *cid.Cid) (api.PinInfo, bool) {
+	return spt.optracker.FilterGet(c, optracker.PhaseError)
+}
+
+func (spt *Tracker) getErrorsAll() []api.PinInfo {
+	return spt.optracker.Filter(optracker.PhaseError)
+}
+
+func (spt *Tracker) removeError(c *cid.Cid) {
+	spt.optracker.Clean(spt.optracker.GetOp(c))
```
hsanjuan: Can you add a `cleanError(cid)` to the optracker which gets an op and cleans it only if it's in error?

And then remove `GetOp(c)` from it. I don't like that you can get operations from it that you did not have beforehand (or are new).

pintracker/stateless/stateless.go
lanzafame: Yeah, that sounds like a really good solution. 👍

lanzafame:
@hsanjuan I have updated this PR to use the new `gorpc` method.

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,490 @@
...
+		return nil, err
+	}
+	pins := make(map[string]api.PinInfo, 0)
+	for cidstr, ips := range ipsMap {
+		c, err := cid.Decode(cidstr)
+		if err != nil {
+			logger.Error(err)
+			continue
+		}
+		p := api.PinInfo{
```
hsanjuan: Set `.TS` to `time.Now()`, as it reflects the last time the record was updated.

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,490 @@
...
+}
+
+// localStatus returns a joint set of consensusState and ipfsStatus
+// marking pins which should be remote and leaving any ipfs pins that
+// aren't in the consensusState out.
+func (spt *Tracker) localStatus(incRemote bool) (map[string]api.PinInfo, error) {
+	pininfos := make(map[string]api.PinInfo)
+
+	// get shared state
+	var csps []api.PinSerial
```
hsanjuan: I'd appreciate longer, more explicit, variable names. We have `cpsp`, `cspis`.. -> `statePinsSerial`, `statePins` etc. for example. or make sure all variables have comments about what they are storing.

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,490 @@
...
+		struct{}{},
+		&csps,
+	)
+	if err != nil {
+		logger.Error(err)
+		return nil, err
+	}
+	var cspis []api.Pin
+	for _, p := range csps {
+		cspis = append(cspis, p.ToPin())
```
hsanjuan: hmm, the fact that the whole state needs to be serialized, only to re-convert it to the original form supports my idea that we could pass the state component directly to this pintracker during initialization and work directly with it.

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,490 @@
...
+		return nil, err
+	}
+
+	for _, p := range cspis {
+		if p.IsRemotePin(spt.peerID) && incRemote {
+			// add pin to pininfos with a status of remote
+			pininfos[p.Cid.String()] = api.PinInfo{
+				Cid:    p.Cid,
+				Peer:   spt.peerID,
+				Status: api.TrackerStatusRemote,
```
hsanjuan: +TS

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,490 @@
...
+	localpis, err := spt.ipfsStatusAll()
+	if err != nil {
+		logger.Error(err)
+		return nil, err
+	}
+
+	for _, p := range cspis {
+		if p.IsRemotePin(spt.peerID) && incRemote {
+			// add pin to pininfos with a status of remote
+			pininfos[p.Cid.String()] = api.PinInfo{
```
hsanjuan: .String() is a bit expensive operation for Cids. Might as well store it in a variable and re-use.

pintracker/optracker/operationtracker.go
```diff
@@ -138,7 +138,51 @@ func (opt *OperationTracker) Get(c *cid.Cid) api.PinInfo {
 	return pInfo
 }
 
-// GetAll returns PinInfo objets for all known operations
+// FilterGet returns a PinInfo for Cid, only if the Phase or OperationType
+// match what the associated Operation has. Note, only supports
+// filters of type OperationType or Phase, any other type
+// will result in a nil slice being returned.
+func (opt *OperationTracker) FilterGet(c *cid.Cid, filter interface{}) (api.PinInfo, bool) {
```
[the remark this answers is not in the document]
hsanjuan: If `getError` is not needed (per comment below), `FilterGet` can go away too.

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,490 @@
...
+
+func (spt *Tracker) getError(c *cid.Cid) (api.PinInfo, bool) {
+	return spt.optracker.FilterGet(c, optracker.PhaseError)
+}
+
+func (spt *Tracker) getErrorsAll() []api.PinInfo {
+	return spt.optracker.Filter(optracker.PhaseError)
+}
+
+func (spt *Tracker) removeError(c *cid.Cid) {
```
hsanjuan: this is not used. Also, `CleanError` should probably be used (atomic removal of error), but it can be called directly rather than having a wrapping method for the single call.

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,490 @@
...
+	}
+
+	for _, p := range spt.optracker.FilterOps(optracker.OperationPin, optracker.PhaseError) {
+		if _, ok := localpis[p.Cid().String()]; ok {
+			spt.optracker.Clean(p)
+		}
+	}
+
+	for _, p := range spt.optracker.FilterOps(optracker.OperationUnpin, optracker.PhaseError) {
+		if _, ok := localpis[p.Cid().String()]; ok {
```
hsanjuan: Shouldn't it be `!ok`? If it's reported by IPFS, it's not been unpinned, thus it should not be Cleaned.

pintracker/optracker/operationtracker.go
```diff
@@ -159,3 +221,64 @@ func (opt *OperationTracker) GetOpContext(c *cid.Cid) context.Context {
...
+	var fltops []*Operation
+	opt.mu.RLock()
+	defer opt.mu.RUnlock()
+	for _, op := range filterOpsMap(opt.operations, filters) {
+		fltops = append(fltops, op)
+	}
+	return fltops
+}
+
+func filterOpsMap(ops map[string]*Operation, filters []interface{}) map[string]*Operation {
```
hsanjuan: This seems unefficient. It loops the whole map for every filter. Instead of looping it once and check, for every element, if it matches one of the filters. I'm assuming that looping the map is generally more expensive than looping the filters, even if the final number of comparisons is the same.

I also don't see the advantage in using recursion here when a simple `for` over the filters solves the issue. It should be significantly cheaper than creating a whole new context to call the function, but mostly is about readability (there are not so many filter types).

pintracker/optracker/operationtracker.go
```diff
@@ -159,3 +221,64 @@ func (opt *OperationTracker) GetOpContext(c *cid.Cid) context.Context {
...
+		pinfos = append(pinfos, opt.unsafePinInfo(op))
+	}
+	return pinfos
+}
+
+// FilterOps returns a slice that only contains operations
+// with the matching filter. Note, only supports
+// filters of type OperationType or Phase, any other type
+// will result in a nil slice being returned.
+func (opt *OperationTracker) FilterOps(filters ...interface{}) []*Operation {
```
hsanjuan: Per comments above/below, FilterOps should not be necessary.

hsanjuan: i.e. should be replaced by `CleanErrors` which atomically filters and cleans errors.

I will insist forever, there should be no way to get an operation out of the tracker (except when creating one).

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,490 @@
...
+			spt.optracker.Clean(p)
+		}
+	}
+
+	return spt.getErrorsAll(), nil
+}
+
+// Sync returns the updated local status for the given Cid.
+func (spt *Tracker) Sync(c *cid.Cid) (api.PinInfo, error) {
+	return spt.Status(c), nil
```
hsanjuan: `Sync` should do things similar to SyncAll right? Check if it's in error, and check if it should no longer be and clean up the operation then.

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,490 @@
...
+// An error is returned if we are unable to contact the IPFS daemon.
+func (spt *Tracker) SyncAll() ([]api.PinInfo, error) {
+	// get ipfs status for all
+	localpis, err := spt.localStatus(false)
+	if err != nil {
+		logger.Error(err)
+		return nil, err
+	}
+
+	for _, p := range spt.optracker.FilterOps(optracker.OperationPin, optracker.PhaseError) {
```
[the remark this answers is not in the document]
hsanjuan: Actually, the other way to do this (and which saves this atomic method) is to `TrackNewOperation(cid, OperationPin, PhaseDone)`, followed by a call to Clean(op) if it returned a new op (see `maptracker.syncStatus()`)).

api/types.go
```diff
@@ -669,3 +684,14 @@ type Error struct {
 func (e *Error) Error() string {
 	return fmt.Sprintf("%s (%d)", e.Message, e.Code)
 }
+
+// CidNotInGlobalStateError allows for the differentiation of network
+// and rpc errors from an error indicating that the Pin isn't in the
+// global state.
+type CidNotInGlobalStateError struct {
```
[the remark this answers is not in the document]
lanzafame: ✅

cluster.go
```diff
@@ -888,7 +888,7 @@ func (c *Cluster) Pins() []api.Pin {
 func (c *Cluster) PinGet(h *cid.Cid) (api.Pin, error) {
 	pin, ok := c.getCurrentPin(h)
 	if !ok {
-		return pin, errors.New("cid is not part of the global state")
+		return pin, &api.CidNotInGlobalStateError{Cid: h}
```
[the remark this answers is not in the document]
lanzafame: ✅

cluster_test.go
```diff
@@ -323,7 +327,7 @@ func TestClusterRecoverAllLocal(t *testing.T) {
 	defer cl.Shutdown()
 
 	c, _ := cid.Decode(test.TestCid1)
-	err := cl.Pin(api.PinCid(c))
+	err := cl.Pin(api.Pin{Cid: c, ReplicationFactorMax: -1})
```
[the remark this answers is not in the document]
lanzafame: Hmm, yeah I think I was making sure whilst debugging something. Removed.

pintracker/maptracker/maptracker.go
```diff
@@ -375,3 +375,8 @@ func (mpt *MapPinTracker) SetClient(c *rpc.Client) {
 	mpt.rpcClient = c
 	mpt.rpcReady <- struct{}{}
 }
+
+// OpContext exports the internal optracker's OpContext method.
```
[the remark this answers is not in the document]
lanzafame: I had the comment on the stateless tracker one, 🤦‍♂️

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,490 @@
...
+		&struct{}{},
+	)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Enqueue puts a new operation on the queue, unless ongoing exists.
+func (spt *Tracker) enqueue(c api.Pin, typ optracker.OperationType) error {
```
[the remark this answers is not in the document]
lanzafame: Because they are fields of the tracker itself, it didn't make sense to pass `enqueue` something it already has.

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,490 @@
...
+	err := spt.rpcClient.Call(
+		"",
+		"Cluster",
+		"PinGet",
+		api.PinCid(c).ToSerial(),
+		&gpin,
+	)
+	if err != nil {
+		if !rpc.IsRPCError(err) {
+			// if _, ok := err.(*api.CidNotInGlobalStateError); ok {
```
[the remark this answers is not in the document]
lanzafame: ✅

pintracker/stateless/stateless.go
lanzafame: Good catch, thanks.

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,490 @@
...
+	}
+	go spt.opWorker(spt.unpin, spt.unpinCh)
+	return spt
+}
+
+// receives a pin Function (pin or unpin) and a channel.
+// Used for both pinning and unpinning
+func (spt *Tracker) opWorker(pinF func(*optracker.Operation) error, opChan chan *optracker.Operation) {
+	logger.Debug("entering opworker")
+	ticker := time.NewTicker(1 * time.Second) //TODO(ajl): make config var
```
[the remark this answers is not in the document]
lanzafame: ✅

pintracker/stateless/stateless.go
lanzafame: ✅

pintracker/stateless/stateless.go
lanzafame: ✅

pintracker/stateless/stateless.go
lanzafame: ✅

pintracker/stateless/stateless.go
lanzafame: maybe 🙃

pintracker/stateless/stateless.go
lanzafame: But this is just retrieving the record? It isn't actually updating the record.

pintracker/stateless/stateless.go
lanzafame: See `ipfsStatus` comment re: retrieval vs update

pintracker/optracker/operationtracker.go
lanzafame: > This seems unefficient. It loops the whole map for every filter. 

It may be inefficient but it isn't looping the whole map for every filter. It only loops the whole map once, and then only loops the filtered map for consecutive filters. Reading my godoc of the filter-related functions after not looking at them for a while, I realise they don't convey this at all so I shall change those. 

> Instead of looping it once and check, for every element, if it matches one of the filters. 

This wouldn't allow for the usage seen in `stateless.SyncAll` where the map is filtered by an operation, i.e. Pin/Unpin, and then by phase, i.e. Error, essentially this function results in an [intersection](https://en.wikipedia.org/wiki/Intersection_(set_theory)) of the provided filters on the operation map. 

> I also don't see the advantage in using recursion here when a simple for over the filters solves the issue. 

The reason why I went with recursion as it resulted in much more readable code and code climate complained about my initial non-recursive implementation.

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,490 @@
...
+			for _, doneOp := range doneOps {
+				spt.optracker.Clean(doneOp)
+			}
+		case op := <-opChan:
+			if cont := applyPinF(pinF, op); cont {
+				continue
+			}
+
+			switch op.Type() {
+			case optracker.OperationUnpin, optracker.OperationRemote:
```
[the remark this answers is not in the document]
lanzafame: ¯\_(ツ)_/¯ have fixed

pintracker/optracker/operationtracker.go
hsanjuan: Ah, I was reading this as an OR filter, not an AND. If the problem is filtering by operations and then by phase, we might as well use TrackerStatus as filter value directly

pintracker/stateless/stateless.go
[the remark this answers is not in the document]
hsanjuan: saves a few lines of code (the `switch`). So it was passing information that enqueue did not have (which channel to use). Also, making it hard to arrive to the default case where it doesn't know which channel to use.

pintracker/stateless/stateless.go
hsanjuan: Read it as: "As of <timestamp>, the state of this Cid was <status>".

pintracker/stateless/stateless.go
hsanjuan: Means we need a test (probably missing in maptracker too?)

pintracker/optracker/operationtracker.go
lanzafame: The issue with using a TrackerStatus is that then you can't get all `errors` or all `pinOps` regardless of phase.

lanzafame: I got halfway through implementing it to use TrackerStatus when I realised this.

hsanjuan: Use `-1` for all ?

lanzafame: But that would not be filtering anything?

hsanjuan: Sorry, misread. Maybe making filter be a slice of slices. [][]TrackerStatus were `{{status1, status2}, {status3}}` means (Status1 OR status 2) AND (status3) ? Or simply two methods (filterOR, filterAND)

lanzafame: thoughts on this as a way of doing the filtering via trackerstatuses: https://play.golang.org/p/ykGXNsnderW

lanzafame: So I implemented this and it adds a lot of complexity that really outweighs the benefits of getting rid of the `interface{}` parameter. And it is actually more error prone because the conversion from `TrackerStatus` to `OperationType` and `Phase` isn't perfect. On top of that, the behaviour is less intuitive. 

I am of the opinion that we leave it as is and if we have issues with this later, we tackle it as a separate thing to the stateless tracker. I am happy to push up the changes as separate PR?

pintracker/stateless/stateless.go
hsanjuan: This is still missing

pintracker/stateless/stateless_test.go
```diff
@@ -0,0 +1,532 @@
...
+	return nil
+}
+
+func init() {
+	flag.Var(&customLogLvlFacilities, "logfacs", "use -logLevel for only the following log facilities; comma-separated")
+	flag.StringVar(&logLevel, "loglevel", logLevel, "default log level for tests")
+	flag.Parse()
+
+	if len(customLogLvlFacilities) <= 0 {
+		for f := range ipfscluster.LoggingFacilities {
```
hsanjuan: This introduces a dependency back to cluster from this package. 

As a policy, components should not need to use the `ipfscluster` package for anything as they are fully independent and this can only complicate things in the future (create cycles, difficult extractions etc).


This also introduces code which is essentially a dup from the main package tests. My impression is that, since these tests are limited to this component you should not need any fine-tuning of the facilities (there is basically 1 only).

pintracker/stateless/stateless.go
lanzafame: ✅

pintracker/stateless/stateless_test.go
lanzafame: I want to split this test logging flags into a separate package, in a different PR, I will just remove it from here for now. Was only for debugging purposes anyway.

pintracker/stateless/stateless.go
```diff
@@ -0,0 +1,526 @@
...
+	pinCh   chan *optracker.Operation
+	unpinCh chan *optracker.Operation
+
+	shutdownMu sync.Mutex
+	shutdown   bool
+	wg         sync.WaitGroup
+}
+
+// New creates a new StatelessPinTracker.
+func New(cfg *Config, pid peer.ID) *Tracker {
```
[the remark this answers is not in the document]
lanzafame: I added a benchmark to the `localStatus` function, where the `Cluster.Pins` rpc call is made and the deserialization occurs, that way I can measure an improvement via passing in the State. I will do that in a separate PR.

lanzafame:
@hsanjuan I think I have address all the outstanding comments.

hsanjuan:
@lanzafame I had it in mind that some tests here were failing and I should have a look. Was this the case and the issues were fixed or it was simply not the case?

lanzafame:
I think there was but that was a while ago and is now fixed. Though now I have code climate issue 😑 🙃

hsanjuan:
It's fine with me if not fixed, unless the fix improves things and is simple

hsanjuan:
@lanzafame: I can't request your review because it's your PR, but mind looking at the latest commits before we merge?

PR #472, opened 2018-06-20
Fix #465: Remove unneeded repotoken from goveralls

### LINKED ISSUE #465: Do not attempt to push coverage in repository forks
#### Basic information

  * [x] Master
  * [ ] Latest stable version
  * [x] Bug
  * [ ] Feature request
  * [ ] Enhancement
  * [x] Linux
  * [ ] macOS
  * [ ] Windows
  * [ ] Other: `which?`

#### Description

When someone forks this repository and sends a PR, our travis script tries to push coverage to coveralls, but since people forking the project has no access to the TOKEN, this fails and the whole travis job with it. We should be able to detect that a build is not happening in the original repository and avoid failing the whole build because of that. If this is not easy/possible. in the worst case, we should just ignore `goveralls` failures.

Laevos:
License: MIT
Signed-off-by: Lilith McMullen <iggnsthe@live.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ jobs:
         - go get -u golang.org/x/tools/cmd/cover
         - make deps
         - go test -v -coverprofile=coverage.out -covermode=count ./...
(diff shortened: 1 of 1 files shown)


Laevos:
@hsanjuan Comments addressed.

PR #478, opened 2018-06-24
Fix #445: Implemented status filter for ipfs-cluster-ctl

### LINKED ISSUE #445: Status filters for `ipfs-cluster-ctl status`
#### Pre-check

#### Basic information

  * [ ] Master
  * [ ] Latest stable version
  * [ ] Bug
  * [ ] Feature request
  * [ ] Enhancement
  * [ ] Linux
  * [ ] macOS
  * [ ] Windows
  * [ ] Other: `which?`
  * [ ] Built from sources
  * [ ] Docker
  * [ ] Snap
  * [ ] Other: `which?`

#### Description

`ipfs-cluster-ctl` status command should support a filter like `--filter pinned`, `--filter error`, `--filter queued`.

These should match the types of [`TrackerStatus`](https://godoc.org/github.com/ipfs/ipfs-cluster/api#pkg-constants) but also have aliases. i.e. `--filter error` should give items in `pin_error` and `unpin_error`. Perhaps, also a way to combining several filters `--filter error,pinning`.

The filter function should probably be implemented in the `rest/api` component and be a server-side function controlled by a url query parameter in the `/pins` endpoint. Support for it needs to be bubbled to the rest/client and ipfs-cluster-ctl.

In particular, the filter should remove elements from []GlobalPinInfo when none of the peers in GlobalPinInfo match the filter. If one peer in the GlobalPinInfo matches the filter, the whole object is returned, including the information for the other peers which may or not match it.

sona1111:
Familiarizing myself with the automated merge checks, waiting for confirmation from @lanzafame on further proposed changes for this Fix

License: MIT
Signed-off-by: Paul Jewell sona1111@zoho.com

### DIFF
--- a/api/rest/restapi.go
+++ b/api/rest/restapi.go
@@ -585,6 +585,69 @@ func (api *API) allocationHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// defined groups / aliases for filtering
+var filterGroups = map[string][]string{
+	"error":  {"error", "pin_error", "unpin_error", "cluster_error"},
+	"queued": {"queued", "pin_queued", "unpin_queued"},
+}
+
+// check if any of the strings in matches are in the list
+func anyStringInSlice(matches, list []string) bool {
+	for _, b := range list {
+		for _, d := range matches {
+			if len(b) > 0 && d == b {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func checkFilterMatches(filterList []string, value string) bool {
+	// in case we have not defined any alises for the passed string, just use the string itself
+	if alias, exists := filterGroups[value]; exists {
+		return anyStringInSlice(alias, filterList)
+	}
+	return anyStringInSlice([]string{value}, filterList)
+}
+
+func filterStatusAll(filter string, pinInfos []types.GlobalPinInfoSerial) []types.GlobalPinInfoSerial {
+	if filter == "" {
+		return pinInfos
+	}
+	filterList := strings.Split(strings.ToLower(filter), ",")
+	for i := len(pinInfos) - 1; i >= 0; i-- {
+		removeThisEntry := true
+		for _, entry := range pinInfos[i].PeerMap {
+			if checkFilterMatches(filterList, entry.Status) {
+				removeThisEntry = false
+				break
+			}
+		}
+		// if we get to this point, no peers matched, so remove the group
+		if removeThisEntry == true {
+			pinInfos = append(pinInfos[:i], pinInfos[i+1:]...)
+		}
+	}
+	return pinInfos
+}
+
+func filterStatus(filter string, pinInfos []types.PinInfoSerial) []types.PinInfoSerial {
+	if filter == "" {
+		return pinInfos
+	}
+	filterList := strings.Split(strings.ToLower(filter), ",")
+	for i := len(pinInfos) - 1; i >= 0; i-- {
+		if checkFilterMatches(filterList, pinInfos[i].Status) {
+			// one peer matched, so we can just exit
+			break
+		}
+		pinInfos = append(pinInfos[:i], pinInfos[i+1:]...)
+
+	}
+	return pinInfos
+}
+
 func (api *API) statusAllHandler(w http.ResponseWriter, r *http.Request) {
 	queryValues := r.URL.Query()
 	local := queryValues.Get("local")
@@ -596,6 +659,7 @@ func (api *API) statusAllHandler(w http.ResponseWriter, r *http.Request) {
 			"StatusAllLocal",
 			struct{}{},
 			&pinInfos)
+		pinInfos = filterStatus(queryValues.Get("filter"), pinInfos)
 		sendResponse(w, err, pinInfosToGlobal(pinInfos))
 	} else {
 		var pinInfos []types.GlobalPinInfoSerial
@@ -604,6 +668,7 @@ func (api *API) statusAllHandler(w http.ResponseWriter, r *http.Request) {
 			"StatusAll",
 			struct{}{},
 			&pinInfos)
+		pinInfos = filterStatusAll(queryValues.Get("filter"), pinInfos)
 		sendResponse(w, err, pinInfos)
 	}
 }
--- a/api/rest/restapi.go
+++ b/api/rest/restapi.go
@@ -591,19 +591,19 @@ var filterGroups = map[string][]string{
 	"queued": {"queued", "pin_queued", "unpin_queued"},
 }
 
-func parseAllFilters(raw_filters string) map[string]bool {
+func parseAllFilters(rawFilters string) map[string]bool {
 	// given the initial string of filters, return a slice of all individual strings
 	// to match on. This includes replacing alises
-	rawFilterList := strings.Split(strings.ToLower(raw_filters), ",")
+	rawFilterList := strings.Split(strings.ToLower(rawFilters), ",")
 	var filterList = map[string]bool{}
 
-	for _, raw_filter := range rawFilterList {
-		if alias_slice, exists := filterGroups[raw_filter]; exists {
-			for _, alias := range alias_slice {
+	for _, rawFilter := range rawFilterList {
+		if aliasSlice, exists := filterGroups[rawFilter]; exists {
+			for _, alias := range aliasSlice {
 				filterList[alias] = true
 			}
 		} else {
-			filterList[raw_filter] = true
+			filterList[rawFilter] = true
 		}
 	}
 	return filterList
--- a/api/rest/restapi.go
+++ b/api/rest/restapi.go
@@ -591,59 +591,52 @@ var filterGroups = map[string][]string{
 	"queued": {"queued", "pin_queued", "unpin_queued"},
 }
 
-// check if any of the strings in matches are in the list
-func anyStringInSlice(matches, list []string) bool {
-	for _, b := range list {
-		for _, d := range matches {
-			if len(b) > 0 && d == b {
-				return true
+func parseAllFilters(raw_filters string) map[string]bool {
+	// given the initial string of filters, return a slice of all individual strings
+	// to match on. This includes replacing alises
+	rawFilterList := strings.Split(strings.ToLower(raw_filters), ",")
+	var filterList = map[string]bool{}
+
+	for _, raw_filter := range rawFilterList {
+		if alias_slice, exists := filterGroups[raw_filter]; exists {
+			for _, alias := range alias_slice {
+				filterList[alias] = true
 			}
+		} else {
+			filterList[raw_filter] = true
 		}
 	}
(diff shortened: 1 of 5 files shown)


api/rest/restapi.go
```diff
@@ -585,6 +585,69 @@ func (api *API) allocationHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// defined groups / aliases for filtering
+var filterGroups = map[string][]string{
```
lanzafame: @hsanjuan Ideal candidate for separate `ipfs-cluster-ctl` configuration file?

lanzafame: I don't mean for this to be included in this PR. Just making a note.

api/rest/restapi.go
```diff
@@ -585,6 +585,69 @@ func (api *API) allocationHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+	return anyStringInSlice([]string{value}, filterList)
+}
+
+func filterStatusAll(filter string, pinInfos []types.GlobalPinInfoSerial) []types.GlobalPinInfoSerial {
+	if filter == "" {
+		return pinInfos
+	}
+	filterList := strings.Split(strings.ToLower(filter), ",")
+	for i := len(pinInfos) - 1; i >= 0; i-- {
```
lanzafame: This would be one of those loops where a [`label`](https://golang.org/ref/spec#Label_scopes) really simplifies the code:
```go
	var filtered []types.GlobalPinInfoSerial

NEXTPI:
	for _, pi := range pinInfos {
		for _, entry := range pi.PeerMap {
			if anyStringInSlice(filterList, entry.Status) {
				filtered = append(filtered, pi)
				continue NEXTPI
			}
		}
	}
```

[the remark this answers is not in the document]
sona1111: @lanzafame Iteration is performed over the `pinInfos` slice, which has the possibility of being modified inside the loop at line 629, so reverse iterating is used to ensure that no indices are skipped in the case that an entry is removed. Please let me know if this is a common strategy.

api/rest/restapi.go
```diff
@@ -585,6 +585,69 @@ func (api *API) allocationHandler(w http.ResponseWriter, r *http.Request) {
...
 }
 
+// defined groups / aliases for filtering
+var filterGroups = map[string][]string{
+	"error":  {"error", "pin_error", "unpin_error", "cluster_error"},
+	"queued": {"queued", "pin_queued", "unpin_queued"},
+}
+
+// check if any of the strings in matches are in the list
+func anyStringInSlice(matches, list []string) bool {
```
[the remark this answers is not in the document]
sona1111: Is using a very short definition like 'find' an acceptable practice here? Should this function be moved into a general utilities file instead of restapi.go ? Your stolen parameter names, though, are spot on.

api/rest/restapi.go
```diff
@@ -585,6 +585,69 @@ func (api *API) allocationHandler(w http.ResponseWriter, r *http.Request) {
...
+		return anyStringInSlice(alias, filterList)
+	}
+	return anyStringInSlice([]string{value}, filterList)
+}
+
+func filterStatusAll(filter string, pinInfos []types.GlobalPinInfoSerial) []types.GlobalPinInfoSerial {
+	if filter == "" {
+		return pinInfos
+	}
+	filterList := strings.Split(strings.ToLower(filter), ",")
```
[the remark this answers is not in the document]
sona1111: This works, but it seems to assume that only a single aliased filter can be used at a time? For example, we have both 'error' and 'queued' aliases, it seems like with this suggestion we would not support `--filter="error,queued"`, is this acceptable?

api/rest/restapi.go
sona1111: So in this case, it is preferred in general, instead of removing from the existing slice, to build up a new slice instead? I will apply this.

sona1111: @lanzafame If you have a chance, can you link to a doc or relay how to start the mock RPC and run the tests in restapi_test.go ? Thank you!

hsanjuan: @sona1111 to run the api tests just do `go test -v ./api/rest` (from the repo root)

hsanjuan: (after `make deps`, that is)

sona1111:
@lanzafame I have made new commits which address the concerns previously brought up, but I am not sure if as you said many of the changes will be rendered redundant due to the ongoing discussion around the issue. Please let me know if I should make any more changes for this PR at this time.

api/rest/restapi.go
```diff
@@ -585,6 +585,62 @@ func (api *API) allocationHandler(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func filterStatus(filter string, pinInfos []types.PinInfoSerial) []types.PinInfoSerial {
+	if filter == "" {
+		return pinInfos
+	}
+	filterList := parseAllFilters(filter)
+	for i := len(pinInfos) - 1; i >= 0; i-- {
+		if _, exists := filterList[pinInfos[i].Status]; !exists {
+			pinInfos = append(pinInfos[:i], pinInfos[i+1:]...)
```
lanzafame: I was adding another test for just the `filterStatus` function and it kept failing even though `TestAPIStatusAllFilter` was passing. It finally clicked that my input slice, which was being passed to the two test cases, was being directly edited. So the first test case was passing, `error alias`, but the input slice then only contained the results of the `error alias` and so the second use case then failed as there were no `pinnin` statuses present. 

Generally in Go, if you want to edit a slice/array directly, that is fine but you just don't return a result of the same type, for example [`json.Unmarshal`](https://golang.org/pkg/encoding/json/#example_Unmarshal). But if you want to return the filtered result, you shouldn't edit the input slice, either by making a copy or the way it is done in `filterStatusAll`. See this [post](https://blog.golang.org/go-slices-usage-and-internals) for more information about how slices work.

api/rest/restapi.go
```diff
@@ -585,6 +585,62 @@ func (api *API) allocationHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// defined groups / aliases for filtering
+var filterGroups = map[string][]string{
+	"error":  {"error", "pin_error", "unpin_error", "cluster_error"},
```
lanzafame: `error` and `queued` shouldn't be included in the lists of actual statuses as there is no associated `api.TrackerStatus` with those values.

api/rest/restapi.go
```diff
@@ -585,6 +585,62 @@ func (api *API) allocationHandler(w http.ResponseWriter, r *http.Request) {
...
+func filterStatusAll(filter string, pinInfos []types.GlobalPinInfoSerial) []types.GlobalPinInfoSerial {
+	if filter == "" {
+		return pinInfos
+	}
+	filterList := parseAllFilters(filter)
+	var filtered []types.GlobalPinInfoSerial
+NEXTPI:
+	for _, pi := range pinInfos {
+		for _, entry := range pi.PeerMap {
+			if _, exists := filterList[entry.Status]; exists {
```
lanzafame: `filterList` is a `map[string]bool` where the `bool` value is only true if it does exist, so this if statement can be shortened to the following:
```go
if filterList[entry.Status] {
```

lanzafame: This is the same in `filterStatus`.

api/rest/restapi.go
sona1111: @lanzafame While I was applying this fix I ran into a little confusion about the way in which the filter should work with a CID specified (specifically in statusHandler() as opposed to statusAllHandler() ). In the previous commits I excluded filtering in this case for two reasons: 1) I did not know how to properly return a null result from the handler function when there were no matches. (I could likely figure this out myself with some work) and 2) I am not sure this is the desired behavior for a filter. To elaborate on (2), with one CID specified, the user would specify a filter for example 'pin_queued'. They would see one of two results; either the single CID with all peers if even one peer matched, or nothing at all. This is consistent with the behavior of the filter for statusAllHandler() but does not seem to be very useful to the user? Let me know if I should proceed with this logic or only apply filter to statusAllHandler() for now.

lanzafame:
@sona1111 Sorry for the radio silence. A two trip to Europe and jetlag recovery. I will be getting to this PR this week 👍

lanzafame:
@sona1111 as it stands please continue with this PR. This is this simplest approach to begin with anyway. Cheers.

sona1111:
@lanzafame Thanks for replying, I know you have had a lot to catch up with. The actual question was a little lower in the conversation, though. I will quote it here:

> While I was applying this fix I ran into a little confusion about the way in which the filter should work with a CID specified (specifically in statusHandler() as opposed to statusAllHandler() ). In the previous commits I excluded filtering in this case for two reasons: 1) I did not know how to properly return a null result from the handler function when there were no matches. (I could likely figure this out myself with some work) and 2) I am not sure this is the desired behavior for a filter. To elaborate on (2), with one CID specified, the user would specify a filter for example 'pin_queued'. They would see one of two results; either the single CID with all peers if even one peer matched, or nothing at all. This is consistent with the behavior of the filter for statusAllHandler() but does not seem to be very useful to the user? Let me know if I should proceed with this logic or only apply filter to statusAllHandler() for now.

Let me know what you decide to do for the statusHandler() case when you have a chance.

api/rest/restapi.go
lanzafame: From the user's perspective, if calling `filterStatus` it should be under the assumption that the query is being run against the peer that they are connected to, i.e.
```sh
$ ssh user@cluster-peer-1
$ ipfs-cluster-ctl --filter 'pin_queued' status QmdGd4C72HGpeGPLxUtDJMSxmgu7PxKYUbEzkfCvyJxdRX
QmdGd4C72HGpeGPLxUtDJMSxmgu7PxKYUbEzkfCvyJxdRX :
    > Peer QmSCZtfARHfCG6Ai8CYvfpNCTNudGa9UkajFuJNVRyDZaq : PIN_QUEUED | 2018-08-10T01:43:15Z
$ ssh user@cluster-peer-9
$ ipfs-cluster-ctl --filter 'pin_queued' status QmdGd4C72HGpeGPLxUtDJMSxmgu7PxKYUbEzkfCvyJxdRX
<no output because it isn't being pinned on this peer>
```

Does that clarify the issue?

lanzafame:
@sona1111 Sorry I swear I replied to this late last week but it appears GitHub ate my reply. 

For the `statusHandler` still apply filtering. The filtering will be in the context of statuses returned from the cluster peer that you are connected to, required if the `--local` flag is passed to the status command. I.e:

```
$ ssh cluster-peer-1
$ ipfs-cluster-ctl status --local --filter='pin_queued'
<filtered output from cluster-peer-1>
```

kishansagathiya:
@sona1111 
I have raise another PR to finish the work done by you. Thanks a lot for you contribution.

PR #489, opened 2018-07-17
Feat: Enable DHT-based peer discovery and routing for cluster peers

hsanjuan:
This uses go-libp2p-kad-dht as routing provider for the Cluster Peers.

This means that:

* A cluster peer can discover other Cluster peers even if they are
not in their peerstore file.
* We remove a bunch of code sending and receiving peers multiaddresses
when a new peer was added to the Cluster.
* PeerAdd now takes an ID and not a multiaddress. We do not need to
ask the new peer which is our external multiaddress nor broadcast
the new multiaddress to everyone. This will fix problems when bootstrapping
a new peer to the Cluster while not all the other peers are online.
* Adding a new peer does not mean to open connections to all peers
anymore. The number of connections will be made according to the DHT
parameters (this is good to have for future work)

The that detecting a peer addition in the watchPeers() function does
no longer mean that we have connected to it or that we know its
multiaddresses. Therefore it's no point to save the peerstore in these
events anymore.

Here a question opens, should we save the peerstore at all, and should we
save multiaddresses only for cluster peers, or for everyone known?
Currently, the peerstore is only updated on clean shutdown,
and it is updated with all the multiaddresses known, and not limited to
peer IDs in the cluster, (because, why not).

License: MIT

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -15,7 +15,9 @@ import (
 	rpc "github.com/hsanjuan/go-libp2p-gorpc"
 	cid "github.com/ipfs/go-cid"
 	host "github.com/libp2p/go-libp2p-host"
+	dht "github.com/libp2p/go-libp2p-kad-dht"
 	peer "github.com/libp2p/go-libp2p-peer"
+	routedhost "github.com/libp2p/go-libp2p/p2p/host/routed"
 	ma "github.com/multiformats/go-multiaddr"
 )
 
@@ -34,6 +36,7 @@ type Cluster struct {
 	id          peer.ID
 	config      *Config
 	host        host.Host
+	dht         *dht.IpfsDHT
 	rpcServer   *rpc.Server
 	rpcClient   *rpc.Client
 	peerManager *pstoremgr.Manager
@@ -86,6 +89,8 @@ func NewCluster(
 		return nil, errors.New("cluster host is nil")
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
+
 	listenAddrs := ""
 	for _, addr := range host.Addrs() {
 		listenAddrs += fmt.Sprintf("        %s/ipfs/%s\n", addr, host.ID().Pretty())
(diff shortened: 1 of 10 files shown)


rpc_api.go
```diff
@@ -335,24 +334,6 @@ func (rpcapi *RPCAPI) ConsensusPeers(ctx context.Context, in struct{}, out *[]pe
 	return err
 }
 
-/*
```
lanzafame: Should we really be 'just' removing these RPC calls? Yes, we are pre-v1 but I feel like we should provide at least some form of depreciation period. Especially, considering we don't know to what extent our users have integrated with the API.

cluster.go
```diff
@@ -97,15 +102,30 @@ func NewCluster(
 		logger.Infof("IPFS Cluster v%s listening on:\n%s\n", Version, listenAddrs)
 	}
 
+	// Note, we already loaded peers from peerstore into the host
+	// in daemon.go.
 	peerManager := pstoremgr.New(host, cfg.GetPeerstorePath())
 
-	ctx, cancel := context.WithCancel(context.Background())
+	idht, err := dht.New(ctx, host)
```
lanzafame: Is this a new cluster DHT or a new connection to the IPFS public DHT?

rpc_api.go
hsanjuan: RPC API is internal, so yes, I'm happy 'just' removing.

cluster.go
hsanjuan: cluster DHT

rpc_api.go
lanzafame: 👍

hsanjuan:
@lanzafame is it good from code point of view? (i fixed the tests now)

Is it ok to just save the peerstore on shutdown now? And is it ok to include all known addresses in it? Other possibilities are: not saving the peerstore at all (not touching what the user provides), or saving only addresses for current peers.

lanzafame:
> @lanzafame is it good from code point of view? (i fixed the tests now)

Yep.

> Is it ok to just save the peerstore on shutdown now? And is it ok to include all known addresses in it?

Yes and yes.

hsanjuan:
@lanzafame what's missing here? I'm going to look at that failing test, see if I can make it not fail randomly

lanzafame:
@hsanjuan sorry, nothing, it just fell off my radar.

PR #498, opened 2018-08-10
Feat/rest/client: auto-handling of libp2p endpoints

hsanjuan:
This removes PeerAddr and uses APIAddr directly, figuring out if it is
a Peer multiaddress or not.

PeerAddr is actually kept for compatiblity.

It also fixes a bad panic when resolving returned 0 results

License: MIT

### DIFF
--- a/api/rest/client/client.go
+++ b/api/rest/client/client.go
@@ -2,7 +2,6 @@ package client
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 	"time"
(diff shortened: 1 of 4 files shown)


api/rest/client/client.go
```diff
@@ -39,28 +39,28 @@ type Config struct {
 	Password string
 
 	// The ipfs-cluster REST API endpoint in multiaddress form
-	// (takes precedence over host:port). Only valid without PeerAddr.
+	// (takes precedence over host:port). It this address contains
+	// an /ipfs/, /p2p/ or /dnsaddr, the API will be contacted
+	// through a libp2p tunnel, thus getting encryption for
+	// free. The latter overrides any SSL configurations.
```
lanzafame: > The latter overrides any SSL configurations.

I am not sure what is going to override any SSL configurations? On first read, I thought `latter` was referring to `/dnsaddr` but that doesn't make much sense.

hsanjuan:
@lanzafame should be good now

PR #505, opened 2018-08-16
Fix #495: Hijack proxy /add correctly. Disable keep-alives.

### LINKED ISSUE #495: Proxy-add does not work for larger files
#### Pre-check

#### Basic information

  * [x] Master
  * [x] Latest stable version
  * [x] Bug
  * [ ] Feature request
  * [ ] Enhancement
  * [x] Linux
  * [ ] macOS
  * [ ] Windows
  * [ ] Other: `which?`
  * [ ] Built from sources
  * [ ] Docker
  * [ ] Snap
  * [ ] Other: `which?`

#### Description

Adding a file through the proxy fails for someone larger files (~MB). I think we are not correctly reading the full multipart body when adding, and likely just sending the first part in the proxied request.

This issue should be fixable (very easily) after the sharding branch is merged.

hsanjuan:
No more hacks around /add. This uses the local adder when hijacking /add.
It supports the parameters and works pretty well with the ipfs CLI, showing
progress and everything.

#495


hsanjuan:
Related to why we need to disable keep alives for adding to work correctly: https://github.com/ipfs/go-ipfs/issues/5168

adder/adderutils/adderutils.go
```diff
@@ -0,0 +1,68 @@
+// Package adderutils provides some utilities for adding content to cluster.
```
lanzafame: I not sure this really requires being in its own utils package... Unless I am missing an import cycle, I believe this could simply be a function in the `adder` package.

hsanjuan: yeah, import cycle :(. adder[utils] -> `local` -> `adder`

PR #506, opened 2018-08-17
Feat: Flush added output

hsanjuan:
This flushes AddedOutput updates to the client as soon as they happen,
thus avoiding the weird buffering that was happening before.

License: MIT


hsanjuan:
Overseed by #505

PR #544, opened 2018-09-22
update go-cid and go-libp2p

lanzafame:
License: MIT

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,15 +15,15 @@
     },
     {
       "author": "whyrusleeping",
-      "hash": "QmUDzeFgYrRmHL2hUB6NZmqcBVQtUzETwmFRUc9onfSSHr",
+      "hash": "QmUEqyXr97aUbNmQADHYNknjwjjdVpJXEt1UZXmSG81EV4",
       "name": "go-libp2p",
-      "version": "6.0.8"
+      "version": "6.0.12"
     },
     {
       "author": "hsanjuan",
-      "hash": "Qma1GvD8sVdAaawmhh6rCM9fhqHuxoXZqLDPJNmD1YK8mZ",
+      "hash": "QmaR1KVXHKDyysrMUCT1WKbu9fkFs12U7nLDPhfkswonzj",
       "name": "go-libp2p-raft",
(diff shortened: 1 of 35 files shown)


package.json
```diff
@@ -69,27 +69,27 @@
...
       "name": "go-ipfs-api",
-      "version": "1.3.5"
+      "version": "1.3.6"
     },
     {
       "author": "whyrusleeping",
-      "hash": "QmPqk5grMduFU1Cj5onoB4GyV9XRsSG82KtX4wQs8Gd3XA",
+      "hash": "QmY1L5krVk8dv8d74uESmJTXGpoigVYqBVxXXz1aS8aFSb",
       "name": "go-libp2p-floodsub",
-      "version": "100.9.24"
```
hsanjuan: I use a version of pubsub with a custom `package.json` which only has `libp2p` as dep, and not dependencies of the libp2p stack separately, because they are usually out of sync with the actual dependency tree in libp2p and it's a pain to align (and the pubsub people think this is ok).

So process to upgrade consists in taking this branch https://github.com/libp2p/go-floodsub/tree/gx/libp2p-dep,  aligning with master, simplyfing the `package.json` like https://github.com/libp2p/go-floodsub/blob/gx/libp2p-dep/package.json and publishing with majorVersion+100. So `100.9.24` corresponds to the repackaging of `0.9.24`. Sounds horrible? Way better than to deal with the each of the subdependencies of pubsub when they don't match those in libp2p.

hsanjuan:
(tests fail)

hsanjuan:
Pretty good job bubbling though. No `deps dupes` of relevance. :+1:

package.json
lanzafame: I took the easier over more correct route in updating the `gx/libp2p-dep` branch, I rebased master branch then reset back to the last gx update (v0.9.28) and then performed the simplifying of the dependencies. I can go about another way if want @hsanjuan: https://github.com/libp2p/go-floodsub/pull/108

hsanjuan: yeah that's ok. Then upgrading libp2p becomes easy.

PR #550, opened 2018-09-26
Fix #520 snapcraft stages

### LINKED ISSUE #520: Snapcraft push happens for all the 6 sub-jobs in Travis CI
#### Basic information

  * [x] Master
  * [x] Latest stable version
  * [x] Bug
  * [ ] Feature request
  * [ ] Enhancement

#### Description

Since we paralelized jobs, it seems that master builds run the `deploy` step in every of the build jobs. This results in multiple snapcraft pushes and delays.

hsanjuan:
Fixes #520

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ go:
 services:
 - docker
 
+sudo: required
+
 cache:
   directories:
     - $GOPATH/src/gx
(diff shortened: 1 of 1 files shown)


hsanjuan:
@elopio is there a way to automatically publish snaps tagged `stable` when we tag a release?

elopio:
@hsanjuan as far as I know, it's not possible just with snapcraft.io. I think you would have to use travis or launchpad.

For example, I'm doing something like this with ipfs: https://github.com/elopio/ipfs-snap/blob/master/.travis.yml#L22
And like this with solc: https://launchpad.net/solidity/+snaps

But, I'm no longer working on snapcraft. Things might have changed. You can ask on forum.snapcraft.io, or talk to @popey and @flexiondotorg.

PR #551, opened 2018-09-26
Allow running peers with different cluster versions in the same cluster

hsanjuan:
This patch modifies the RPC protocol tag to use Major and Minor parts of the
version and not all of it.

This means all peers on the 0.5.x can run in the same cluster.

As cluster has become more mature and I see less risks in letting peers from
similar versions run together. This is useful when upgrading too.

License: MIT

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -99,11 +99,7 @@ func NewCluster(
 		listenAddrs += fmt.Sprintf("        %s/ipfs/%s\n", addr, host.ID().Pretty())
 	}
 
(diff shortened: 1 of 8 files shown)


hsanjuan:
@lanzafame what do you mean? The Commit information is part of the Version object now (as `.Build`). UI-wise things should be very similar (?).

lanzafame:
@hsanjuan Ahh sorry, I missed that.

PR #557, opened 2018-09-28
Fix #544: Push snaps to the right channel depending on tags

hsanjuan:
Also, make sure they are only built on master builds of the original
repo.

License: MIT


hsanjuan:
I think I broke travis for this.

PR #558, opened 2018-09-28
Update to go-1.11

hsanjuan:
Start testing with go-1.11

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,7 @@
     "url": "https://github.com/ipfs/ipfs-cluster"
   },
   "gx": {
-    "dvcsimport": "github.com/ipfs/ipfs-cluster",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -4,8 +4,7 @@
     "url": "https://github.com/ipfs/ipfs-cluster"
   },
   "gx": {
-    "dvcsimport": "github.com/ipfs/ipfs-cluster",
-    "goversion": "1.10"
```
hsanjuan: Unfortunately I need to remove this. This places a requirement on the Go version Gx was compiled with (and not that of our project). Since we download pre-built binaries for Gx already it doesn't add much other than being confusing (different go versions in different places).

hsanjuan:
good catch

PR #562, opened 2018-10-01
Snap build fixes. Re-enable multiarch builds.

hsanjuan:
License: MIT


hsanjuan:
Unsure how snaps event built before, if TARGET_ARCH seemed by all means unset. :/ magic

PR #569, opened 2018-10-03
Add endpoints output: improvements and compliance

hsanjuan:
This straigthens some mistakes with the outputs of the /add endpoints.

Currently, we had exactly the same output format which:

* was not exactly the ipfs API output format but was sort of similar
* made some weird concessions to be compatible (like having a string-type "size")
* was not aligned with Cluster API conventions (lowercase keys)

This corrects all this:

* The Cluster API /add output format now uses the right types and lowercase keys.
* `Hash` is now `Cid`, because the field carries a Cid.
* We copy error handling with request trailers from IPFS, and avoid carrying the
  errors in the output objects.
* The proxy now returns exactly the types as ipfs would
* We add the X-Chunked-Output: 1 header, which is custom and redundant, but
otherwise breaks js-ipfs-api integrations with the /add endpoint.

Fixes: https://github.com/ipfs-shipyard/ipfs-companion/issues/600

License: MIT

### DIFF
--- a/adder/ipfsadd/add.go
+++ b/adder/ipfsadd/add.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	gopath "path"
-	"strconv"
 
 	"github.com/ipfs/ipfs-cluster/api"
 
@@ -390,9 +389,9 @@ func outputDagnode(out chan *api.AddedOutput, name string, dn ipld.Node) error {
 	}
 
 	out <- &api.AddedOutput{
-		Hash: dn.Cid().String(),
+		Cid:  dn.Cid().String(),
 		Name: name,
-		Size: strconv.FormatUint(s, 10),
+		Size: s,
 	}
 
 	return nil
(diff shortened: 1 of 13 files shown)


adder/ipfsadd/add.go
```diff
@@ -390,9 +389,9 @@ func outputDagnode(out chan *api.AddedOutput, name string, dn ipld.Node) error {
 	}
 
 	out <- &api.AddedOutput{
-		Hash: dn.Cid().String(),
+		Cid:  dn.Cid().String(),
```
lidel: I suspect this change will require update to js-ipfs-api, which still expects `Hash` in [file-result-stream-converter.js#L32](https://github.com/ipfs/js-ipfs-api/blob/4191525f69433e16dd09b862821b7f8b95f5ef06/src/utils/file-result-stream-converter.js#L32) and [L38](https://github.com/ipfs/js-ipfs-api/blob/4191525f69433e16dd09b862821b7f8b95f5ef06/src/utils/file-result-stream-converter.js#L38)
(It should try `Hash` and if missing use `Cid`)

hsanjuan: hey, this only affects the Cluster API /add. The IPFS Proxy /add output will be fully compatible with the ipfs output format.

hsanjuan: swap is here: https://github.com/ipfs/ipfs-cluster/pull/569/files#diff-66308423c5638970cde2dd02574201c3R431

lidel: Ah, I get it now, thanks for explanation!

PR #573, opened 2018-10-10
Pubsubmon: Gossipsub

hsanjuan:
Uh oh


hsanjuan:
:-( @vyzo so 1 test with just 2 peers, one of them does `Publish` but the other one doesn't seem to receive any message.

However, things seem to work fine on the tests with multiple peers (6) and which would fail if things didn't reach all of them.

vyzo:
You probably want to lower the initial heartbeat delay for the tests, because it takes a 100ms before the overlay is ready to use -- see `GossipSubHeartbeatInitialDelay`

PR #576, opened 2018-10-13
Extract the IPFS Proxy from ipfshttp

### LINKED ISSUE #453: Extract the IPFS Proxy from ipfshttp and make it an api module
#### Basic information

  * [ ] Bug
  * [ ] Feature request
  * [x] Enhancement

#### Description

The `ipfshttp` IPFSConnector implementation includes the so called IPFS Proxy. And endpoint which offers an IPFS API, hijacking some interesting requests and forwarding the rest to the ipfs daemon. This is, in essence, an API Component implementation for Cluster, but it is mixed with the IPFS Connector (whose only task should be to talk to ipfs).

So,

* The whole Proxy code should be moved to `api/ipfsproxy` and be just another Cluster API implementation.
* The new component will need the ipfs daemon location as part of its configuration. Note this option will be duplicate in the `ipfshttp` and `ipfsproxy` configuration sections.
* Related and needed along with this #452

kishansagathiya:
Extract the IPFS Proxy from `ipfshttp` and make it an api module

The `ipfshttp` IPFSConnector implementation includes the so called IPFS
Proxy. An endpoint which offers an IPFS API, hijacking some interesting
requests and forwarding the rest to the ipfs daemon.

`ipfshttp` should contain an implementation of IPFSConnector whose only
task should be to talk to IPFS
A new module should be created, `api/ipfsproxy`, an API Component
implementation for Cluster. The whole proxy code should be moved here.

TODO

- [x] Move proxy into `api/ipfsproxy`
- [x] Implement Config interface for ipfsproxy
- [x] Make the cluster use IPFSProxy Component
- [ ] Tests 

Fixes #453


kishansagathiya:
This is yet not a working PR, will add more changes

kishansagathiya:
@lanzafame @hsanjuan 
I have tested the new IPFSProxy component locally and current sharness tests are also not broken. Let me know if the change looks good to you. After that, I will get started on writing tests.

api/ipfsproxy/ipfsproxy.go
```diff
@@ -0,0 +1,467 @@
...
+// From https://github.com/ipfs/go-ipfs/blob/master/core/coreunix/add.go#L49
+type ipfsAddResp struct {
+	Name  string
+	Hash  string `json:",omitempty"`
+	Bytes int64  `json:",omitempty"`
+	Size  string `json:",omitempty"`
+}
+
+// NewIPFSProxy returns and IPFSProxy component
+func NewIPFSProxy(cfg *Config) (*IPFSProxy, error) {
```
hsanjuan: We can rename this to `New(..) (*Server, error)`.

Then the methods for the server object can be like `func(proxy *Server) Name(...)`.

cluster.go
```diff
@@ -49,6 +49,7 @@ type Cluster struct {
 
 	consensus Consensus
 	api       API
+	proxy     API
```
hsanjuan: we actually should have an `[]API` slice where we put all the APIs.

cluster.go
```diff
@@ -174,6 +177,7 @@ func (c *Cluster) setupRPCClients() {
 	c.tracker.SetClient(c.rpcClient)
 	c.ipfs.SetClient(c.rpcClient)
 	c.api.SetClient(c.rpcClient)
+	c.proxy.SetClient(c.rpcClient)
```
hsanjuan: `for _, api := range apis { api.SetClients(c.rpcClient) }` and so on...

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -496,12 +181,8 @@ func (ipfs *Connector) Shutdown() error {
 		return nil
 	}
 
-	logger.Info("stopping IPFS Proxy")
```
hsanjuan: would be good to log `stopping IPFS Connector`.

kishansagathiya: Had missed this one.

api/ipfsproxy/config.go
```diff
@@ -0,0 +1,180 @@
...
+
+// Default values for Config.
+const (
+	DefaultProxyAddr              = "/ip4/127.0.0.1/tcp/9095"
+	DefaultNodeAddr               = "/ip4/127.0.0.1/tcp/5001"
+	DefaultProxyReadTimeout       = 0
+	DefaultProxyReadHeaderTimeout = 5 * time.Second
+	DefaultProxyWriteTimeout      = 0
+	DefaultProxyIdleTimeout       = 60 * time.Second
+	DefaultIPFSRequestTimeout     = 5 * time.Minute
```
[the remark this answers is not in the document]
kishansagathiya: It is used here https://github.com/ipfs/ipfs-cluster/pull/576/files#diff-40d1514e01fa463b9839f0d7c6625184R398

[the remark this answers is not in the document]
hsanjuan: Let's use the `r.Context()` instead of a timeout. You can create the contexts using: https://github.com/ipfs/ipfs-cluster/blob/master/rpcutil/rpcutil.go#L35

api/ipfsproxy/ipfsproxy.go
```diff
@@ -0,0 +1,467 @@
...
+		ctx:      ctx,
+		config:   cfg,
+		cancel:   cancel,
+		nodeAddr: nodeAddr,
+		rpcReady: make(chan struct{}, 1),
+		listener: l,
+		server:   s,
+	}
+	smux.Handle("/", proxyHandler)
+	smux.HandleFunc("/api/v0/pin/add", ipfs.pinHandler)
```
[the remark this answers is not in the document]
kishansagathiya: I thought those routes were missed. I guess not, will remove those extra routes.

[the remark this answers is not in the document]
kishansagathiya: pretty good catch btw

hsanjuan:
By the way @kishansagathiya , regarding tests, it should be matter of just moving them. The proxy is tested in the `ipfshttp` module, as far as I remember.

cluster.go
kishansagathiya: We can use some tool that automates such code reviews such as golintci

lanzafame: Those tools can also be run locally before pushing and you can automate that with things like [pre-commit](https://pre-commit.com/) as well. Plus in this particular case, it wouldn't have caught it as this comment is based on a requested change above.

kishansagathiya: Will set it up locally and will look into pre-commit. thanks @lanzafame

kishansagathiya:
> This looks quite good to me. The only outstanding question is whether we want to maintain configuration compatibility and still recognize the proxy configuration options in the ipfshttp connector configuration section. I think we should do this.

@hsanjuan Do you mean to have multiple proxies(one in ipfsconnector and one seperately) in order to not break the compatibility ?

kishansagathiya:
@lanzafame Will rebase it. This requires a bunch more things, such as tests and  still recognize the proxy configuration options in the ipfshttp connector configuration. Working on that.

api/ipfsproxy/ipfsproxy.go
```diff
@@ -0,0 +1,467 @@
...
+			ipfs.config.NodeAddr,
+		)
+		err := ipfs.server.Serve(ipfs.listener) // hangs here
+		if err != nil && !strings.Contains(err.Error(), "closed network connection") {
+			logger.Error(err)
+		}
+	}()
+}
+
+// Handlers
```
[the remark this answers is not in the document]
kishansagathiya: ` function premature` is a new work for me :laughing: 
handler is to tell that from now on all handers are starting.

config/config.go
```diff
@@ -306,15 +314,19 @@ func (cfg *Manager) LoadJSON(bs []byte) error {
...
-	loadCompJSON(sections[Monitor], jcfg.Monitor)
-	loadCompJSON(sections[Allocator], jcfg.Allocator)
-	loadCompJSON(sections[Informer], jcfg.Informer)
-	loadCompJSON(sections[Sharder], jcfg.Informer)
+	loadSectionJSON(sections[Consensus], jcfg.Consensus)
+	loadSectionJSON(sections[API], jcfg.API)
+	// Should we change hardcoded "ipfsproxy" to something else
+	if _, ok := jcfg.API["ipfsproxy"]; !ok {
+		loadCompJSON("ipfsproxy", sections[API]["ipfsproxy"], jcfg.IPFSConn)
+	}
```
kishansagathiya: This seems a bit hardcoded. Seeing how I can change this.

kishansagathiya:
Right now working on breaking tests. Will go over this PR a few more times given its size.

kishansagathiya:
ideally we should get rid of the proxy suffix from config elements (`proxy_listen_multiaddress`, `proxy_write_timeout` etc), but since we want to support earlier config as well, let's keep this as it is.

config/config.go
kishansagathiya: Also, I have utilized the property that superset JSON object can be unmarshalled into subject subset JSON structs

Here, previous IPFSConnector config is superset of new IPFSConnector config and IPFSProxy config. So, if our service.json has old IPFSConnector config, we can unmarshal configs for new IPFSConnector and IPFSProxy.
I don't know how well known this propery is so [here](https://play.golang.org/p/qdO4AJ-QbfG) is an illustration.

kishansagathiya:
Will go over this once more before making it ready for review.

kishansagathiya:
This is ready for review

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -496,15 +181,14 @@ func (ipfs *Connector) Shutdown() error {
...
-
 	ipfs.cancel()
 	close(ipfs.rpcReady)
-	ipfs.server.SetKeepAlivesEnabled(false)
-	ipfs.listener.Close()
 
 	ipfs.wg.Wait()
 	ipfs.shutdown = true
+
+	logger.Info("stopping IPFS Connector")
```
hsanjuan: Log this before you actually stop it (in the place where it was before)

hsanjuan:
Hey @kishansagathiya , in general, can you write

> ideally we should get rid of the proxy suffix from config elements (`proxy_listen_multiaddress`, `proxy_write_timeout` etc), but since we want to support earlier config as well, let's keep this as it is.

Actually, this is a good point. You can add both `proxy_` and the fields we want to the ipfs proxy JSON config  object. Leave the `proxy_` ones with `omitempty` and empty values.

Then in LoadJSON, right after unmarshaling the section, check if, for example, the `read_timeout` is 0, then set it to the value from `proxy_read_timeout`. And so on for other fields. And then let the function proceed as before.

kishansagathiya:
> Actually, this is a good point. You can add both proxy_ and the fields we want to the ipfs proxy JSON config object. Leave the proxy_ ones with omitempty and empty values.

That's a great idea, but I am inclined to do that a separate PR.
- This would require more changes like changing all variable names, using field with `proxy_` when a service.json file is generated etc
- This PR is already getting huge, removing `proxy_` would just add more confusion.

If you think otherwise I don't really mind changing what you asked.
@hsanjuan

hsanjuan:
@kishansagathiya please create a follow up issue (or PR). Thanks!

PR #588, opened 2018-10-23
Add a docker-compose.yml example with 2 peers and auto-bootstrap.

hsanjuan:
`docker-compose up` runs a cluster with 2 peers and 2 ipfs daemons right away. Stopping and `up` again still works. This should be an example, I think every user will have to adapt this to their needs, particularly if they want to run this with `swarm` in multiple hosts, there should be a way to set `placement` options so that both `ipfs0 and cluster0` run in the same place (for convenience I guess?).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,8 +38,7 @@ services:
     depends_on:
       - ipfs0
     environment:
-      # Change this! Same for all peers!
-      CLUSTER_SECRET: adc72159e0ce63350c13de6a5e4d3b2f500232a770cb8c3e7829fc514d89d109 
+      CLUSTER_SECRET: ${CLUSTER_SECRET} # From shell variable
       IPFS_API: /dns4/ipfs0/tcp/5001
     ports:
           - "127.0.0.1:9094:9094" # API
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,104 @@
...
+      - ./compose/ipfs0:/data/ipfs
+      
+  cluster0:
+    container_name: cluster0
+    image: ipfs/ipfs-cluster:latest
+    depends_on:
+      - ipfs0
+    environment:
+      # Change this! Same for all peers!
+      CLUSTER_SECRET: adc72159e0ce63350c13de6a5e4d3b2f500232a770cb8c3e7829fc514d89d109
```
lanzafame: Wouldn't it be better to set this to an environment variable and put a 'default' cluster secret in the Makefile?

hsanjuan: Yes... it would be better. I just don't want people to forget and this is an example. But I guess a cluster without secret is as bad as a cluster with a default one. At least we can generate one on the fly in the Makefile. I'll fix this.

PR #589, opened 2018-10-23
Make metrics human

hsanjuan:
Issue #572 exposes metrics but they carry the peer ID in binary.

This was ok with our internal codecs but it doesn't seem to work
very well with json, and makes the output format unusable.

This makes the Metric.Peer field a string.

Additionally, fixes calling the command without arguments and displaying
the date in the right format.

License: MIT

### DIFF
--- a/api/rest/client/request.go
+++ b/api/rest/client/request.go
@@ -92,9 +92,10 @@ func (c *defaultClient) handleResponse(resp *http.Response, obj interface{}) err
 			var apiErr api.Error
 			err = json.Unmarshal(body, &apiErr)
 			if err != nil {
+				// not json. 404s etc.
 				return &api.Error{
 					Code:    resp.StatusCode,
-					Message: err.Error(),
+					Message: string(body),
 				}
 			}
 			return &apiErr
--- a/api/types.go
+++ b/api/types.go
@@ -9,6 +9,8 @@
 package api
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"sort"
@@ -847,10 +849,10 @@ func (n *NodeWithMeta) Size() uint64 {
 // the Value, which should be interpreted by the PinAllocator.
 type Metric struct {
 	Name   string
-	Peer   peer.ID // filled-in by Cluster.
+	Peer   peer.ID
 	Value  string
-	Expire int64 // UnixNano
-	Valid  bool  // if the metric is not valid it will be discarded
+	Expire int64
+	Valid  bool
 }
 
 // SetTTL sets Metric to expire after the given time.Duration
@@ -874,6 +876,51 @@ func (m *Metric) Expired() bool {
 // Discard returns if the metric not valid or has expired
 func (m *Metric) Discard() bool {
 	return !m.Valid || m.Expired()
+}
+
+// helper for JSON marshaling. The Metric type is already
+// serializable, but not pretty to humans (API).
+type metricSerial struct {
+	Name   string `json:"name"`
+	Peer   string `json:"peer"`
+	Value  string `json:"value"`
+	Expire int64  `json:"expire"`
+	Valid  bool   `json:"valid"`
+}
+
+// MarshalJSON allows a Metric to produce a JSON representation
+// of itself.
+func (m *Metric) MarshalJSON() ([]byte, error) {
+	return json.Marshal(&metricSerial{
+		Name:   m.Name,
+		Peer:   peer.IDB58Encode(m.Peer),
+		Value:  m.Value,
+		Expire: m.Expire,
+	})
+}
+
+// UnmarshalJSON decodes JSON on top of the Metric.
+func (m *Metric) UnmarshalJSON(j []byte) error {
+	if bytes.Equal(j, []byte("null")) {
+		return nil
+	}
+
+	ms := &metricSerial{}
+	err := json.Unmarshal(j, ms)
+	if err != nil {
+		return err
+	}
+
+	p, err := peer.IDB58Decode(ms.Peer)
+	if err != nil {
+		return err
+	}
+
+	m.Name = ms.Name
+	m.Peer = p
+	m.Value = ms.Value
+	m.Expire = ms.Expire
+	return nil
 }
 
 // Alert carries alerting information about a peer. WIP.
--- a/cluster.go
+++ b/cluster.go
@@ -275,7 +275,10 @@ func (c *Cluster) alertsHandler() {
 			// only the leader handles alerts
 			leader, err := c.consensus.Leader()
 			if err == nil && leader == c.id {
-				logger.Warningf("Peer %s received alert for %s in %s", c.id, alrt.MetricName, alrt.Peer.Pretty())
+				logger.Warningf(
+					"Peer %s received alert for %s in %s",
+					c.id, alrt.MetricName, alrt.Peer,
+				)
 				switch alrt.MetricName {
 				case pingMetricName:
 					c.repinFromPeer(alrt.Peer)
--- a/cmd/ipfs-cluster-ctl/formatters.go
+++ b/cmd/ipfs-cluster-ctl/formatters.go
@@ -6,8 +6,10 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/ipfs/ipfs-cluster/api"
+	peer "github.com/libp2p/go-libp2p-peer"
 )
 
 func jsonFormatObject(resp interface{}) {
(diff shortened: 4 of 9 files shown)


cmd/ipfs-cluster-ctl/main.go
```diff
@@ -768,10 +768,21 @@ graph of the connections.  Output is a dot file encoding the cluster's connectio
 					Description: `
 This commands displays the latest valid metrics of the given type logged
 by this peer for all current cluster peers.
+
+Currently supported metrics depend on the informer component used,
+but usually are:
+
+- freespace
+- ping
```
lanzafame: :+1:

cluster.go
```diff
@@ -275,10 +275,16 @@ func (c *Cluster) alertsHandler() {
 			// only the leader handles alerts
 			leader, err := c.consensus.Leader()
 			if err == nil && leader == c.id {
-				logger.Warningf("Peer %s received alert for %s in %s", c.id, alrt.MetricName, alrt.Peer.Pretty())
+				logger.Warningf(
+					"Peer %s received alert for %s in %s",
+					c.id, alrt.MetricName, alrt.Peer)
```
lanzafame: nitpick: can we bump the paren to the next line please

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -216,7 +217,8 @@ func textFormatPrintAddedOutput(obj *api.AddedOutput) {
 }
 
 func textFormatPrintMetric(obj *api.Metric) {
-	fmt.Printf("%s: %s | Expire : %d\n", obj.Peer.Pretty(), obj.Value, obj.Expire)
+	date := time.Unix(0, obj.Expire).UTC().Format(time.RFC3339)
+	fmt.Printf("%s: %s | Expire: %s\n", obj.Peer, obj.Value, date)
```
lanzafame: @hsanjuan Why not just decode the `peer.ID` here instead of doing it everywhere else in PR?

api/rest/client/request.go
```diff
@@ -92,9 +92,10 @@ func (c *defaultClient) handleResponse(resp *http.Response, obj interface{}) err
 			var apiErr api.Error
 			err = json.Unmarshal(body, &apiErr)
 			if err != nil {
+				// not json. 404s etc.
 				return &api.Error{
 					Code:    resp.StatusCode,
-					Message: err.Error(),
+					Message: string(body),
```
kishansagathiya: or body.String() ?

api/types.go
```diff
@@ -846,11 +846,11 @@ func (n *NodeWithMeta) Size() uint64 {
...
-	Name   string
-	Peer   peer.ID // filled-in by Cluster.
-	Value  string
-	Expire int64 // UnixNano
-	Valid  bool  // if the metric is not valid it will be discarded
+	Name   string `json:"name"`
+	Peer   string `json:"peer"` // filled-in by Cluster.
+	Value  string `json:"value"`
+	Expire int64  `json:"expire"` // UnixNano
+	Valid  bool   `json:"valid"`  // if the metric is not valid it will be discarded
```
kishansagathiya: :+1:

cluster.go
kishansagathiya: why not `alrt.Peer.Pretty`?

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -768,10 +768,21 @@ graph of the connections.  Output is a dot file encoding the cluster's connectio
...
-					ArgsUsage: "Metric name",
+					ArgsUsage: "<metric name>",
 					Action: func(c *cli.Context) error {
-						resp, cerr := globalClient.Metrics(c.Args().First())
+						metric := c.Args().First()
+						if metric == "" {
+							checkErr("", errors.New("provide a metric name"))
+						}
+
+						resp, cerr := globalClient.Metrics(metric)
```
kishansagathiya: :+1:

monitor/basic/peer_monitor_test.go
```diff
@@ -33,7 +33,7 @@ func (mf *metricFactory) newMetric(n string, p peer.ID) api.Metric {
 	defer mf.l.Unlock()
 	m := api.Metric{
 		Name:  n,
-		Peer:  p,
+		Peer:  peer.IDB58Encode(p),
```
kishansagathiya:
```suggestion
		Peer:  p.Pretty(),
```

monitor/basic/peer_monitor_test.go
```diff
@@ -233,7 +237,7 @@ func TestPeerMonitorAlerts(t *testing.T) {
 			if alrt.MetricName != "test" {
 				t.Error("Alert should be for test")
 			}
-			if alrt.Peer != test.TestPeerID1 {
+			if alrt.Peer != peer.IDB58Encode(test.TestPeerID1) {
```
kishansagathiya:
```suggestion
			if alrt.Peer != test.TestPeerID1.Pretty() {
```

monitor/metrics/checker_test.go
```diff
@@ -17,7 +17,7 @@ func TestChecker(t *testing.T) {
 
 	metr := api.Metric{
 		Name:  "test",
-		Peer:  test.TestPeerID1,
+		Peer:  peer.IDB58Encode(test.TestPeerID1),
```
kishansagathiya:
```suggestion
		Peer:  test.TestPeerID1.Pretty(),
```

monitor/metrics/checker_test.go
```diff
@@ -61,7 +61,7 @@ func TestCheckerWatch(t *testing.T) {
...
```
kishansagathiya:
```suggestion
		Peer:  test.TestPeerID1.Pretty(),
```

kishansagathiya: This is at a bunch of other places, avoiding to comment everywhere

cluster.go
lanzafame: @kishansagathiya `Pretty()` is a method on a `peer.ID`, at this point `alrt.Peer` is now a string

kishansagathiya: right, my bad

api/rest/client/request.go
hsanjuan: body is a `[]byte`, it doesn't have `String()`.

monitor/basic/peer_monitor_test.go
hsanjuan: No, IDB58Encode will probably always produce B58 encoding of the peer. Pretty() is subject to people whims on what 'pretty' means. Even the output produced by `.String()` has changed recently, without notice. Same is going to happen for `cid.String()` which will probably change to be base32.

cmd/ipfs-cluster-ctl/formatters.go
lanzafame: Not really or I misunderstood.

The point I was trying to make is that now have to do a decode operation every time you want to store the peer, which is a lossy operation in the type system sense, instead of making peer.ID.Pretty() work properly or making a peer.ID.JSON() function which would be called here and there be the changes all over the metrics code. Does that make sense? To summarise, can we fix the output of the value so we aren't having to do it on the input

hsanjuan: @lanzafame but I'm fixing the output of the `/monitor/metrics` endpoint, it's not just the user output here (which did not work anyway because I don't think json knows how to deserealize byte slices properly)

hsanjuan: i.e. I don't think this is acceptable (and doesn't work):

```
[10:49:37] ~ $ curl -s localhost:9094/monitor/metrics/ping | jq
[
  {
    "Name": "ping",
    "Peer": "\u0012 \u000bȇ�\u001a�4�����J���\r4�_U�E5�)xݭD�\u001c",
    "Value": "",
    "Expire": 1540457399324099800,
    "Valid": true
  }
]
```

lanzafame: Right yep. I still prefer the struct bring an peer id because we pass it around so much. We could just define a json marshaller for the metric struct so it handles it properly

hsanjuan: I'll go the `MetricSerial` way then, for consistency with the rest. Since we convert object into serializable-Go objects, which can then be encoded not only as json, but also with more efficient codecs (like the go-codec thing we use for rpcs etc). I don't know if we should change that, but if we do I'd like it changed everywhere.

hsanjuan: Actually I'll check json marshaler/unmarshaler too, first. Maybe worth making an exception because this type is serializable, it's just json not working well for it.

hsanjuan:
@lanzafame this is in line with your proposal to have Metric know how to Marshal itself into and out of JSON.

hsanjuan:
Related: https://github.com/libp2p/go-libp2p/issues/393 We should implement marhsaling in libp2p-peer.

PR #594, opened 2018-10-25
Fix publishing snaps

hsanjuan:
Snaps won't publish with classic confimement or non-edge.
For the moment this works around that so that, at least,
there is releases on 'edge'  with strict confinement.

License: MIT

fyi @brianmcmichael . Will potentially re-enable `classic` if ever we're whitelisted.


hsanjuan:
merging this. failures are because of other libp2p problems

PR #595, opened 2018-10-25
Fix tests with latest libp2p

hsanjuan:
I merged too quickly what I thought was a harmless libp2p upgrade and this is the result.


hsanjuan:
Context: p2p-circuit was enabled by default in libp2p, that makes `Addrs()` return two addresses. We assumed that the first listen address in our hosts would be actually dialable.

PR #596, opened 2018-10-26
add env config; refactor LoadJSON tests

### LINKED ISSUE #575: Env variables for configuration
This is a feature request.

The hardest part of maintaining `ipfs-cluster` is changing the configuration "on the fly" by changing the env variables. This is quiet common in docker containers which I am extensively using.

Most of the configuration now lives into its own file, and can't be changed after running the first init, and even by adding `-f`. For example if you need to change the `CLUSTER_SECRET` variable, and keep persistency, you need to force a new initialisation or delete the `service.json` file

What would be nice to have is the configuration to get updated based on the environment variables. The `CLUSTER_SECRET` is one quick-win, easy example, but it can be applied to several different cases (ex: bootstrapping vs pre-defined peers).

This can be implemented in different ways: 

* only for docker: forcing a `-f init` each restart
* by implementing some more common practices. For example [Kelsey Hightower's library](https://github.com/kelseyhightower/envconfig) can be quite useful to achieve this.

lanzafame:
Resolves #575 

@hsanjuan as I mentioned in #575 viper was a bit much of a jump in terms of how we use config and how viper provides access to config, i.e. every viper.Get("var") will query the current env/file/where ever else it gets config from, and we don't expect config to change underneath us. Not that it is a bad thing, but I wanted something that got us over the line with changing every config read we do.

Tests were refactored to make it not possible to have any state left over from the previous test case.

Docs need to be added to show that prefixing any config var with `CLUSTER_` in the environment will allow the setting of that config field.


lanzafame:
@hsanjuan I will remove the envconfig aspect of this PR and just leave the test refactors, in light of this [comment](https://github.com/ipfs/ipfs-cluster/issues/575#issuecomment-433351623)

cluster_config_test.go
```diff
@@ -23,118 +24,178 @@ var ccfgTestJSON = []byte(`
...
-	j = &configJSON{}
-	json.Unmarshal(ccfgTestJSON, j)
-	j.ReplicationFactorMin = 0
-	j.ReplicationFactorMax = 0
-	tst, _ = json.Marshal(j)
-	err = cfg.LoadJSON(tst)
-	if cfg.ReplicationFactorMin != -1 || cfg.ReplicationFactorMax != -1 {
-		t.Error("expected default replication factors")
-	}
+	loadJSON := func(t *testing.T) (*Config, error) {
```
hsanjuan: I have mixed feelings about this refactor. It's longer, more difficult to read and understand, more difficult to copy-paste, just looks nicer on the output but that's not really helpful anyway.

It's good to not copy paste though, but I feel the overall result is not improved.

lanzafame: I find it easier to read because I can read it as just the contents of the `t.Run(...)` and not have to worry about whatever came before a particular test.

hsanjuan: well, you have to worry about jumping to see what `loadJSON2` actually does.

hsanjuan:
@lanzafame ok, merging this for next release. We can do other components in different PRs.

hsanjuan:
Actually, @lanzafame , should we make `CLUSTER_LEAVEONSHUTDOWN` -> `CLUSTER_LEAVE_ON_SHUTDOWN` ? I forgot to take care of that before merging, but I'm also not sure if worth the trouble.

lanzafame:
I will go through and fix all of the kind of named vars.

PR #613, opened 2018-12-01
go-ipfs-files 2.0 updates

magik6k:
These are updates for changes in https://github.com/ipfs/go-ipfs-files/pull/2
* `go-ipfs-api`, `go-mfs` and `go-unixfs` need updating to have matching files versions
* All but 1 tests pass for me
  * `TestAdder_ContextCancelled` fails, but probably is just racy
* Feel free to pick this up once https://github.com/ipfs/go-ipfs-files/pull/2 is merged


GitCop:

 * Commit: 2ffb41179cc4c6af404e80c0507de8b37a6c2f72
    **License: MIT  
Signed-off-by:**

---

adder/ipfsadd/add.go
```diff
@@ -297,12 +298,12 @@ func (adder *Adder) addFile(file files.File) error {
 	}
 	adder.liveNodes++
 
-	if file.IsDirectory() {
-		return adder.addDir(file)
+	if dir, ok := nd.(files.Directory); ok {
```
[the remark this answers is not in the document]
magik6k: A separate helper makes sense, probably more like `files.ToDir(f)` so one can use the cast value.

test/sharding.go
```diff
@@ -121,17 +121,22 @@ func (sth *ShardingTestHelper) GetTreeSerialFile(t *testing.T) files.Directory {
 // a testing random file of the given size (in kbs). The random
 // file is different every time.
 func (sth *ShardingTestHelper) GetRandFileMultiReader(t *testing.T, kbs int) (*files.MultiFileReader, io.Closer) {
-	st := sth.makeRandFile(t, kbs)
-	sf, err := files.NewSerialFile(sth.path(shardingTestFile), false, st)
+	slf, sf := sth.GetRandFileReader(t, kbs)
```
hsanjuan: This is completely changing the semantics of these methods. Before they were readers to a plain file, now they reading files from a directory. This should not return a MultiFileReader with an underlying SliceFile. Could you explain this change?

adder/adder_test.go
```diff
@@ -105,8 +106,20 @@ func TestAdder_ContextCancelled(t *testing.T) {
 	sth := test.NewShardingTestHelper()
 	defer sth.Clean(t)
 
-	mr, closer := sth.GetRandFileMultiReader(t, 50000) // 50 MB
+	lg, closer := sth.GetRandFileReader(t, 50000) // 50 MB
+	st := sth.GetTreeSerialFile(t)
 	defer closer.Close()
+	defer st.Close()
+
+	slf := files.NewSliceFile([]files.DirEntry{
```
hsanjuan: This test should not need to adder to be interating over a SliceFile. Can you explain the change?

magik6k: A change in `adder.FromFiles` changed this behaviour slightly:

Before the change in https://github.com/ipfs/ipfs-cluster/pull/613/files?utf8=%E2%9C%93&diff=split#diff-a2c1e35d2137b0f7d1899f06c11d87b6L136 we would:
* Enter into the loop
* Check context  (successfully)
* Call `NextFile` and process it
* Loop again
* Check context (timeout in test), return

After the change:
* Enter into the loop, calling `Next`
* Check context (successfully)
* Process the file
* Exit loop as `Next` returns false (note that we don't check context as we don't re-enter the loop)
* Exit `FromFiles` successfully as nothing else used the context

Strictly speaking old behaviour is more correct, but since we added the file fully in both cases, I'd argue that it makes sense to just return the result instead of timing out since we are already on the return path either way.

I needed to make this test take more than one file in order to force the loop to go around at least twice

test/sharding.go
magik6k: The old version is returning a directory too (`files.File` was used for both files and dirs) - `NewSliceFile` - with one regular file - `NewSerialFile` called `randomfile`.

One change here is that `NewMultiFileReader` can now return an error (which is actually not needed anymore, I'll change this in `go-ipfs-file`)

magik6k:
Addressed the review and made CI happier. Do you mind giving this / https://github.com/ipfs/go-ipfs-files/pull/2 another pass?

hsanjuan:
And really, thanks a lot @magik6k !

PR #615, opened 2018-12-07
Sharding: Add some tests for sharding codepaths

hsanjuan:
I had these changes laying around on a different branch (along some other code not included here).  They are the ones that surfaced the problems with the batching DAGService. Since that was fixed, we can merge these so they are not lost (I think).

### DIFF
--- a/add_test.go
+++ b/add_test.go
@@ -6,6 +6,7 @@ import (
 	"mime/multipart"
 	"testing"
 
+	"github.com/ipfs/ipfs-cluster/adder/sharding"
 	"github.com/ipfs/ipfs-cluster/api"
 	"github.com/ipfs/ipfs-cluster/test"
 )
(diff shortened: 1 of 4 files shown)


add_test.go
```diff
@@ -44,9 +45,163 @@ func TestAdd(t *testing.T) {
...
+
+		// for next test, this needs to be unpinned.
+		err = clusters[0].Unpin(ci)
+		if err != nil {
+			t.Fatal(err)
+		}
+		pinDelay()
+	})
+
+	t.Run("shard", func(t *testing.T) {
```
lanzafame: It would be great if each of these tests could have a bit more explanation as to what they are testing as the testcase is hard to grok from the code.

RubenKelevra:
@hsanjuan mind cleaning this up. Was reading it and then noticed that date :D

PR #626, opened 2018-12-19
Fix ipfs-cluster-ctl with HTTPs multiaddresses

hsanjuan:
Before we resolved all /dns*/ multiaddresses before we used them.

When using HTTPs, the Go HTTP Client only sees the resolved IP address
and it is unable to negotiate TLS with a cerficate because the request
is not going to the hostname the certificate is signed for, but to
the IP. This leverages a recent feature in go-multiaddr-net
and uses directly the user-provided hostname.

### DIFF
--- a/api/rest/client/client.go
+++ b/api/rest/client/client.go
@@ -6,6 +6,7 @@ package client
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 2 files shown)


lanzafame:
@hsanjuan so it works but it should be noted that TLS termination prior to the cluster plus basic-auth requires that you pass the `--force-http` flag because cluster expects to handle TLS termination and doesn't expect to have the connection be downgraded from https to http between the load balancer and the cluster peer.

hsanjuan:
@lanzafame that should not be like that. TLS termination should be fully transparent to a client. How is the connection downgraded? I don't think a HTTPs connection can be downgraded to HTTP.

hsanjuan:
@lanzafame that means the server you're testing with is not using https

PR #637, opened 2019-01-09
Fix #632: Add --no-stream to ipfs-cluster-ctl

### LINKED ISSUE #632: Proper JSON output from ipfs-cluster-ctl
  * [x] Master
  * [x] Release candidate for next version
  * [x] Latest stable version
  * [ ] Bug
  * [x] Feature request
  * [ ] Enhancement
  * [x] Linux
  * [x] macOS
  * [x] Windows
  * [ ] Other: `which?`
  * [x] Built from sources
  * [x] Docker
  * [x] Snap
  * [ ] Other: `which?`

#### Description
I am programmatically adding the [benchmark results](https://benchmarks.ipfs.team) with [`ipfs-cluster-ctl`](https://github.com/ipfs/benchmarks/blob/master/runner/ipfs.js#L15) and need to return the SHA for presentation purposes. The current output with `--enc json` needs [additional work](https://github.com/ipfs/benchmarks/blob/master/runner/ipfs.js#L27) before it can be parsed as JSON. 

```
{"code":0,"message":"","Name":"test1.jpg","Hash":"QmPxBet4MHo93b9nCE2FYRtDTowVA3bMRdDBZcYAiBkZKN","Size":"4696045"}
{"code":0,"message":"","Name":"test2.jpg","Hash":"QmfJ8fzLwCmuvxNG8RH41RmmjGzMrXw38pjP85EJSiXKJ4","Size":"4696045"}
```
_Note the missing surrounding `[]` and commas between the objects._

As per [this issue](https://github.com/ipfs/ipfs-cluster/issues/526#issuecomment-418120772) I'd like to request a flag to return `non-streaming` json output which can be parsed by `JSON.parse` as is.

> Hi, this is not a bug (it's a feature). /add response is a streaming response and mimics what ipfs /add does (for the moment). Perhaps you'd like to turn this into a feature request asking to support a stream=false parameter and return a single complete json response, or something else, whatever your usecase is, and we'll consider it if there is good reasoning behind it.

hsanjuan:
tl;dr: this solves the user's immediate need and, even if not tne strictest
solution, it is the simplest one.

I think we should not have the server buffer output when we can do it rather
easily client side and have the clients use their own memory for the task even
if `stream-channels=false` would do this.

We can always change the approach, but this is the minimal solution to
json array with all the AddedOutput things.

We might not buffer at all and hack a `[`, `,` separating elements and `]`
at the end, when json encoding is enabled, etc. But that would not be clean,
specially if we mean to support more output formats at some point.

Enabling supporting stream-channels=false in the api/rest/client means adding
new methods, with tests, modifying the interface etc etc. for what is
essentially a presentation issue in "ctl" in the end. Similarly we could
buffer inside the client, but it is so trivial that I don't see advatange.
We should also start thinking about moving to streaming API endpoints and
when that moment arrives we shall revisit this discussion.

I have removed the hacky manual output parts by declaring a custom
addedOutputQuiet type which wraps added output and is understood by
the formatters.go helpers.

License: MIT

### DIFF
--- a/cmd/ipfs-cluster-ctl/formatters.go
+++ b/cmd/ipfs-cluster-ctl/formatters.go
@@ -13,6 +13,11 @@ import (
 	"github.com/ipfs/ipfs-cluster/api"
 )
 
+type addedOutputQuiet struct {
+	added *api.AddedOutput
+	quiet bool
+}
+
 func jsonFormatObject(resp interface{}) {
 	switch resp.(type) {
 	case nil:
@@ -25,6 +30,10 @@ func jsonFormatObject(resp interface{}) {
 		jsonFormatPrint(resp.(api.Pin).ToSerial())
 	case api.AddedOutput:
 		jsonFormatPrint(resp.(api.AddedOutput))
+	case addedOutputQuiet:
+		// print original object as in json it does
+		// not make sense quiet output.
+		jsonFormatPrint(resp.(addedOutputQuiet).added)
 	case api.Version:
 		jsonFormatPrint(resp.(api.Version))
 	case api.Metric:
@@ -57,6 +66,13 @@ func jsonFormatObject(resp interface{}) {
 	case []api.AddedOutput:
 		serials := resp.([]api.AddedOutput)
 		jsonFormatPrint(serials)
+	case []addedOutputQuiet: // print original output
+		serials := resp.([]addedOutputQuiet)
+		var actual []*api.AddedOutput
+		for _, s := range serials {
+			actual = append(actual, s.added)
+		}
+		jsonFormatPrint(actual)
 	case []api.Metric:
 		serials := resp.([]api.Metric)
 		jsonFormatPrint(serials)
@@ -87,6 +103,9 @@ func textFormatObject(resp interface{}) {
 	case api.AddedOutput:
 		serial := resp.(api.AddedOutput)
 		textFormatPrintAddedOutput(&serial)
+	case addedOutputQuiet:
+		serial := resp.(addedOutputQuiet)
+		textFormatPrintAddedOutputQuiet(&serial)
 	case api.Version:
 		serial := resp.(api.Version)
 		textFormatPrintVersion(&serial)
@@ -110,6 +129,10 @@ func textFormatObject(resp interface{}) {
 		}
 	case []api.AddedOutput:
 		for _, item := range resp.([]api.AddedOutput) {
+			textFormatObject(item)
+		}
+	case []addedOutputQuiet:
+		for _, item := range resp.([]addedOutputQuiet) {
 			textFormatObject(item)
 		}
 	case []api.Metric:
@@ -216,6 +239,14 @@ func textFormatPrintPin(obj *api.PinSerial) {
 
 func textFormatPrintAddedOutput(obj *api.AddedOutput) {
 	fmt.Printf("added %s %s\n", obj.Cid, obj.Name)
+}
+
+func textFormatPrintAddedOutputQuiet(obj *addedOutputQuiet) {
+	if obj.quiet {
+		fmt.Printf("%s\n", obj.added.Cid)
+	} else {
+		textFormatPrintAddedOutput(obj.added)
+	}
 }
 
 func textFormatPrintMetric(obj *api.Metric) {
--- a/cmd/ipfs-cluster-ctl/formatters.go
+++ b/cmd/ipfs-cluster-ctl/formatters.go
@@ -31,8 +31,8 @@ func jsonFormatObject(resp interface{}) {
 	case api.AddedOutput:
 		jsonFormatPrint(resp.(api.AddedOutput))
 	case addedOutputQuiet:
-		// print original object as in json it does
-		// not make sense quiet output.
+		// print original object as in JSON it does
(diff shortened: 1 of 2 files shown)


cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -25,6 +30,10 @@ func jsonFormatObject(resp interface{}) {
 		jsonFormatPrint(resp.(api.Pin).ToSerial())
 	case api.AddedOutput:
 		jsonFormatPrint(resp.(api.AddedOutput))
+	case addedOutputQuiet:
+		// print original object as in json it does
```
lanzafame: nitpick: comment needs rewording

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -277,6 +277,10 @@ If you prefer faster adding, add directly to the local IPFS and trigger a
 					Name:  "quieter, Q",
 					Usage: "Write only final hash to output",
 				},
+				cli.BoolFlag{
+					Name:  "no-stream",
+					Usage: "Buffer output locally. Produces a valid slice with --enc=json.",
```
lanzafame: nitpick: s/valid slice/valid JSON slice

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -391,29 +395,36 @@ If you prefer faster adding, add directly to the local IPFS and trigger a
 				wg.Add(1)
 				go func() {
 					defer wg.Done()
-					var last *api.AddedOutput
+
+					var buffered []addedOutputQuiet
+					var lastBuf = make([]addedOutputQuiet, 1, 1)
+					var qq = c.Bool("quieter")
+					var q = c.Bool("quiet") || qq
```
kishansagathiya: On running with `quiet` and `quieter`, it is also printing name and size.

cmd/ipfs-cluster-ctl/main.go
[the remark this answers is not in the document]
hsanjuan: hmm good catch. I changed the name of the flag and introduced a bug. The `--no-stream` output should produce a `[ .. ]` result printed at the end. The streaming output prints AddedOuput object as they arrive.

I want to name it `no-stream` because it relates to how results are streamed to the user. And might eventually mean that this will use no-streaming endpoint from the server in the future.

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -391,29 +395,36 @@ If you prefer faster adding, add directly to the local IPFS and trigger a
...
+					var bufferResults = !c.Bool("stream")
```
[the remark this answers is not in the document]
hsanjuan: yes

cmd/ipfs-cluster-ctl/main.go
hsanjuan: ok

cmd/ipfs-cluster-ctl/formatters.go
hsanjuan: done

cmd/ipfs-cluster-ctl/main.go
hsanjuan: should be fixed to after `--no-stream`

cmd/ipfs-cluster-ctl/main.go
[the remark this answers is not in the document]
hsanjuan: > why not always buffer locally?

In the general case, you are just using memory on stuff that you can flush out. Mid-term plan will be to move any endpoints listing things to streaming, as otherwise it's unpractical to deal with huge pinsets.

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -277,6 +277,10 @@ If you prefer faster adding, add directly to the local IPFS and trigger a
 					Name:  "quieter, Q",
 					Usage: "Write only final hash to output",
```
kishansagathiya: Nothing to do with this PR, but this description is wrong. `quieter` doesn't print just hash. It also prints name and size(that's the behaviour I see in master). It's a little that I think we can push along with this PR.

hsanjuan: It does just print the hash? Unless you mean in the json output, to which this does not really apply because we print the json objects.

```
[18:30:23] ~/ipfstest $ ipfs-cluster-ctl  add -Q -r testsh/
QmekoCaPuQ8wcWhuws4KRZnNhzaTNrbM1e6dqgWRVLyZnw
[18:30:28] ~/ipfstest $ ipfs-cluster-ctl  add --no-stream -Q -r testsh/
QmekoCaPuQ8wcWhuws4KRZnNhzaTNrbM1e6dqgWRVLyZnw
```

kishansagathiya: yup, I was talking about the json output

hsanjuan: yeah, json output should print valid json in any case. -q and -Q have text-formatting implications which do not apply to json (here or in `ipfs add`)

hsanjuan:
Not releasing today so this can wait for your OK @lanzafame

PR #638, opened 2019-01-10
Fix #382 (again): A better strategy for handling proxy headers

### LINKED ISSUE #382: Proxy endpoint doesn't respect IPFS header policies for hijacked paths
IPFS enforces a bunch of stuff regarding CORs, CRLF attacks protection etc. Hijacked requests don't do any of that.

We need to let the user provide IPFS-like API configuration for the proxy, with whatever headers there are etc, and work just like ipfs would work, with the difference that we would still be hijacking the requests.

hsanjuan:
This changes the current strategy to extract headers from the IPFS daemon to
use them for hijacked endpoints in the proxy. The ipfs daemon is a bit of a
mess and what we were doing is not really reliable, specially when it comes to
setting CORS headers right (which we were not doing).

The new approach is:

* For every hijacked request, make an OPTIONS request to the same path, with
the given Origin, to the IPFS daemon and extract some CORS headers from
that. Use those in the hijacked response 
* Avoid hijacking OPTIONS request,
they should always go through so the IPFS daemon controls all the
CORS-preflight things as it wants.  
* Similar to before, have a
only-once-triggered request to extract other interesting or custom headers
from a fixed IPFS endpoint.  This allows us to have the proxy forward other
custom headers and to catch `Access-Control-Expose-Methods`. The difference is
that the endpoint use for this and the additional headers are configurable by
the user (but with hidden configuration options because this is quite exotic
from regular usage).

Now the implementation:

* Replaced the standard Muxer with gorilla/mux (I have also taken the change
to update the gxed version to the latest tag). This gives us much better
matching control over routes and allows us to not handle OPTIONS requests.

* This allows also to remove the extractArgument code and have proper handlers
for the endpoints passing command arguments as the last segment of the URL. A
very simple handler that wraps the default ones can be used to extract the
argument from the url and put it in the query.  Overall much cleaner this way.

* No longer capture interesting headers from any random proxied request.  This
made things complicated with a wrapping handler. We will just trigger the one
request to do it when we need it.

* When preparing the headers for the hijacked responses:
  * Trigger the OPTIONS request and figure out which CORS things we should set
  * Set the additional headers (perhaps triggering a POST request to fetch them)
  * Set our own headers.

* Moved all the headers stuff to a new headers.go file.

* Added configuration options (hidden by default) to:
  * Customize the extract headers endpoint
  * Customize what additional headers are extracted
  * Use HTTPs when talking to the IPFS API
    * I haven't tested this, but I did not want to have hardcoded 'http://' urls
      around, as before.

* Added extra testing for this, and tested manually a lot comparing the
daemon original output with our hijacked endpoint outputs while looking
at the API traffic with ngrep and making sure the requets happen as expected.
Also tested with IPFS companion in FF and Chrome.

License: MIT


api/ipfsproxy/config.go
```diff
@@ -50,16 +55,34 @@ type Config struct {
...
+	// IPFS daemon responses and re-used in responses from hijacked paths.
+	// This is only useful if the user has configured custom headers
+	// in the IFPS daemon. CORS-related headers are already
+	// taken care of by the proxy.
+	ExtractHeadersExtra []string
+
+	// If the user wants to extract some extra custom headers configured
+	// on the IPFS daemon so that they are used in hijacked responses,
+	// this request path will be used. Defaults to /version. This will
+	// trigger a single request to extract those headers and remember them
```
lanzafame: This functionality means that upgrades to the ipfs daemon (in deployment situations) _**may**_ require restarting the associated ipfs-cluster peer.

api/ipfsproxy/headers.go
```diff
@@ -0,0 +1,157 @@
...
+	req.Header["Origin"] = srcRequest.Header["Origin"]
+	req.Header.Set("Access-Control-Request-Method", srcRequest.Method)
+	// error is logged. We proceed if request failed.
+	proxy.copyHeadersFromIPFSWithRequest(corsHeaders, dest, req)
+}
+
+// see setHeaders
+func (proxy *Server) setAdditionalIpfsHeaders(dest http.Header, srcRequest *http.Request) {
+	// Avoid re-requesting these if we have them
+	if ok := proxy.setIPFSHeaders(dest); ok {
```
lanzafame: address the comment I made above having to restart cluster, could we add a TTL to the IPFS headers we are caching?

api/ipfsproxy/headers.go
```diff
@@ -0,0 +1,157 @@
...
+	// The request URL only has a valid Path(). See http.Request docs.
+	srcURL := fmt.Sprintf("%s%s", proxy.nodeAddr, srcRequest.URL.Path)
+	req, err := http.NewRequest(http.MethodOptions, srcURL, nil)
+	if err != nil { // this should really not happen.
+		logger.Error(err)
+		return
+	}
+
+	req.Header["Origin"] = srcRequest.Header["Origin"]
+	req.Header.Set("Access-Control-Request-Method", srcRequest.Method)
```
kishansagathiya: It appears that we have to set `Access-Control-Request-Headers` also
I am referring https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request for this

api/ipfsproxy/headers.go
```diff
@@ -0,0 +1,157 @@
...
+
+// rememberIPFSHeaders extracts headers and stores them for re-use with
+// setIPFSHeaders.
+func (proxy *Server) rememberIPFSHeaders(hdrs http.Header) {
+	for _, h := range proxy.ipfsHeaders() {
+		proxy.ipfsHeadersStore.Store(h, hdrs[h])
+	}
+}
+
+// rememberIPFSHeaders adds the known IPFS Headers to the destination
```
kishansagathiya: typo: setIPFSHeaders

api/ipfsproxy/headers.go
hsanjuan: I consider this an acceptable trade off, but given it's not difficult and it's useful and makes things better, I've just added it.

hsanjuan: Thanks for the suggestion!

api/ipfsproxy/config.go
```diff
@@ -50,16 +55,34 @@ type Config struct {
 	// Server-side amount of time a Keep-Alive connection will be
 	// kept idle before being reused
 	IdleTimeout time.Duration
+
+	// A list of custom headers that should be extracted from
+	// IPFS daemon responses and re-used in responses from hijacked paths.
+	// This is only useful if the user has configured custom headers
+	// in the IFPS daemon. CORS-related headers are already
```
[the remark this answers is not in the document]
hsanjuan: Thanks

api/ipfsproxy/headers.go
hsanjuan: Good question/catch, but things are ok as they are now.

`Request-Headers` is useful in OPTIONS to get `Allow-Headers` in the response. However `Allow-Headers` is not a header that normal (non OPTIONS) requests should return. The headers we want to extract from OPTIONS are those we can re-use in the subsequent response to a normal request.

Note that a CORS-enabled client may have triggered its own OPTIONS request (which is forwarded to ipfs), so the preflight will work and they will get `Allow-Headers` correctly.  Then they trigger the normal request (which we hijack). We just need to make sure that `Allow-Origin`, `Allow-Credentials` and `Expose-Header` are set, because those are the ones that make sense for normal requests.

api/ipfsproxy/headers.go
hsanjuan: thanks

api/ipfsproxy/ipfsproxy.go
```diff
@@ -524,23 +491,40 @@ func (proxy *Server) repoStatHandler(w http.ResponseWriter, r *http.Request) {
...
-		return arg, true
-	}
-
-	p := strings.TrimPrefix(u.Path, "/api/v0/")
-	segs := strings.Split(p, "/")
-
-	if len(segs) > 2 {
-		warnMsg := "You are using an undocumented form of the IPFS API."
+// slashHandler returns a handler which converts a /a/b/c/<argument> request
+// into an /a/b/c/<argument>/?arg=<argument> one. And uses the given origHandler
```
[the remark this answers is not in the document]
hsanjuan: ok

api/ipfsproxy/headers.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	return nil
+}
+
+// setHeaders sets some headers for all hijacked endpoints:
+// - First we fix CORs headers by making an OPTIONS request to IPFS with the
+//   same Origin. Our objective is to get headers for non-preflight requests
+//   only (the ones we hijack).
+// - Second we add any of the one-time-extracted headers that we deem necessary
+//   or the user needs from IPFS (in case of custom headers), or i by ours.
```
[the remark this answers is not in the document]
hsanjuan: ok

kishansagathiya:
Just one more thing. I was trying to test this. I think the idea is not necessarily to have all headers which are return by ipfs, but have at least the CORS related headers

I see something like this
```
kishan@kishansagathiya:~/go/ipfs-cluster/src/github.com/ipfs/ipfs-cluster$ curl -F file=/home/kishan/Downloads/temp -v "http://localhost:9095/api/v0/add?recursive=false"
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 9095 (#0)
> POST /api/v0/add?recursive=false HTTP/1.1
> Host: localhost:9095
> User-Agent: curl/7.58.0
> Accept: */*
> Content-Length: 166
> Content-Type: multipart/form-data; boundary=------------------------828c9c75273536a1
> 
< HTTP/1.1 200 OK
< Access-Control-Expose-Headers: X-Stream-Output, X-Chunked-Output, X-Content-Length
< Cache-Control: no-cache
< Connection: close
< Content-Type: application/json
< Server: ipfs-cluster/ipfsproxy/0.7.0+gita0185fac2a0a3ee621797d624ec2f9663878ba64
< Trailer: X-Stream-Error
< Vary: Origin
< Vary: Access-Control-Request-Method
< Vary: Access-Control-Request-Headers
< X-Chunked-Output: 1
< Date: Fri, 11 Jan 2019 12:37:37 GMT
< Transfer-Encoding: chunked

````
I see the CORS headers in `Vary`, but not separately. Is this expected?

hsanjuan:
> I see the CORS headers in Vary, but not separately. Is this expected?

Yes, because you are not setting an `Origin` header for your request so no cors headers are returned. Only the Vary header and the Expose-Headers because ipfs returns those Vary headers for all preflight requests and the Expose-Header because we extracted it from a normal request.

hsanjuan:
Since not releasing today either... @lanzafame how does it look?

PR #641, opened 2019-01-11
Sharness: update configuration files used in sharness

hsanjuan:
Maintenance.

Depends on #640 . Last commit only (trivial review).


hsanjuan:
@kishansagathiya  you need to click approve to make things green. But I'll merge anyway.

PR #642, opened 2019-01-11
config: Fix confusing errors

hsanjuan:
The JSON parsing of the config could error, but we skipped error checking and
use Validate() at the end. This caused that maybe some JSON parsing errors
where logged but the final error when validating the configuration came from
somewhere different, creating very confusing error messages for the user.

This changes this, along with removing hardcoded section lists. This also
removes a Sharder component section because, AFAIK, it was a left over
from the sharding branch and in the end there is no separate sharding
component that needs a config.

License: MIT

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -46,7 +46,7 @@ type ComponentConfig interface {
 // These are the component configuration types
 // supported by the Manager.
 const (
-	Cluster = iota
+	Cluster SectionType = iota
 	Consensus
 	API
 	IPFSConn
@@ -55,11 +55,20 @@ const (
 	Monitor
 	Allocator
 	Informer
-	Sharder
+	endTypes // keep this at the end
 )
 
 // SectionType specifies to which section a component configuration belongs.
 type SectionType int
+
+// SectionTypes returns the list of supported SectionTypes
+func SectionTypes() []SectionType {
+	var l []SectionType
+	for i := Cluster; i < endTypes; i++ {
+		l = append(l, i)
+	}
+	return l
+}
 
 // Section is a section of which stores
 // component-specific configurations.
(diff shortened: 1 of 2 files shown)


hsanjuan:
Hmm I need to fix a small quirk with this

config/config.go
```diff
@@ -165,7 +174,29 @@ type jsonConfig struct {
 	Monitor    jsonSection      `json:"monitor,omitempty"`
 	Allocator  jsonSection      `json:"allocator,omitempty"`
 	Informer   jsonSection      `json:"informer,omitempty"`
-	Sharder    jsonSection      `json:"sharder,omitempty"`
+}
+
+func (jcfg *jsonConfig) getSection(i SectionType) jsonSection {
+	switch i {
+	case Consensus:
```
kishansagathiya: no `Cluster` section here?

config/util.go
```diff
@@ -96,6 +96,11 @@ type DurationOpt struct {
 // ParseDurations takes a time.Duration src and saves it to the given dst.
 func ParseDurations(component string, args ...*DurationOpt) error {
 	for _, arg := range args {
+		if arg.Duration == "" {
```
kishansagathiya: Any duration field if not passed will take the default value, right?

hsanjuan: yes, since all configs are defaulted before being parsed.

config/config.go
hsanjuan: Fixed (it's not used and I forgot).

hsanjuan: No, actually there is not a section for Clusterbecause it's a *json.RawMessage and it's handled differently. So this was correct.

hsanjuan:
I'll merge this since I'm going to release.

PR #647, opened 2019-01-16
API: Allow user to set peer allocations for pinning

### LINKED ISSUE #646: Ability to Add/Pin file only to certain nodes among one cluster.
Hi There!

Some use cases, especially for enterprise solutions requires control to which certain nodes we want to add/pin file among one cluster.

So it will be good to have this feature in IPFS-Cluster.

This topic have been discussed with Héctor S.
"We don't support manual peer allocations per pin right now, but it can be done rather easily (open an issue)."
Here the link: https://discuss.ipfs.io/t/scalability-and-highload-big-private-network-and-ipfs-cluster/4509/2

Thanks

kishansagathiya:
Added support for `allocations` query value in ` POST /pins/<cid>`
endpoint
`allocations` will be a comma-separated list of peer ids to which we want
to pin.
`allocations` will be prioritized over other query values of similar
effect such as replication factor, replication factor min and
replication factor max

- [ ] support this in command as well
- [ ] tests 

Fixes #646


cmd/ipfs-cluster-ctl/main.go
```diff
@@ -453,9 +453,18 @@ in the cluster and should be part of the list offered by "pin ls".
...
+An optional allocations argument can be provided, allocations should be a
+comma-separated list of peer ids of node on which we want to pin.
+Allocations will be prioritized over other similar options such as replication,
+replication-min and replication-max
 `,
 					ArgsUsage: "<CID>",
 					Flags: []cli.Flag{
+						cli.StringFlag{
+							Name:  "allocations",
+							Usage: "Comma-separated list of peer ids of node to which we want to pin",
```
hsanjuan: `peer IDs`

hsanjuan: `Optional comma-separated list of peer IDs.`

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -453,9 +453,18 @@ in the cluster and should be part of the list offered by "pin ls".
 An optional replication factor can be provided: -1 means "pin everywhere"
 and 0 means use cluster's default setting. Positive values indicate how many
 peers should pin this content.
+
+An optional allocations argument can be provided, allocations should be a
```
hsanjuan: This needs to be updated to explain that the given allocations will have priority over automatically determined ones, but replication factors are respected.

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -495,15 +504,31 @@ peers should pin this content.
 						ci, err := cid.Decode(cidStr)
 						checkErr("parsing cid", err)
 
-						rpl := c.Int("replication")
-						rplMin := c.Int("replication-min")
-						rplMax := c.Int("replication-max")
-						if rpl != 0 {
+						var rpl, rplMin, rplMax int
```
hsanjuan: Oneliner `api.StringsToPeers`.

api/rest/client/methods.go
```diff
@@ -63,17 +63,36 @@ func (c *defaultClient) PeerRm(id peer.ID) error {
...
 // human-friendliness.
-func (c *defaultClient) Pin(ci cid.Cid, replicationFactorMin, replicationFactorMax int, name string) error {
+func (c *defaultClient) Pin(ci cid.Cid, allocations []peer.ID, replicationFactorMin, replicationFactorMax int, name string) error {
+	var urlString string
 	escName := url.QueryEscape(name)
-	err := c.do(
-		"POST",
-		fmt.Sprintf(
+
+	if len(allocations) > 0 {
```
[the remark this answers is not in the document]
hsanjuan: let's scape the query though.

hsanjuan:
Per the PinPath PR, I'm thinking how to best integrate Allocations as part of PinOptions. 

One way would be to move `Allocations` to `PinOptions`. Since it's an array of peer IDs, we will want to have `PinOptionsSerial` in which it becomes an array of strings. As with other types, we will need methods to convert between them. This will cascade to `api.PinSerial` (which will need to carry `PinOptionsSerial` instead of `PinOptions`), and eventually will need a `PinPathSerial` in the same way.

Another way would be to add a `UserAllocations []peer.ID` field. `peer.ID` are actually a string (a byte slice casted into a string actually). This is perfectly serializable, however we cannot spit it out of the API without calling something like `PeersToStrings()` first. We could however, have `PinOptions` implement `json.Marshaler` and `json.Unmarshaler` interface and this will make sure it spitted out correctly out of the API. This is similar to what we started doing with the `Metric` type.

Therefore, for the moment, I'm leaning towards the latter option. Also, keeping separate user-selected allocations and system-given ones seems useful as it allows us to compare if the requested ones were fulfilled.

hsanjuan:
I just checked and `peer.ID` implements JSON marshable interfaces, so it may work "as is", leaving `UserAllocations` as `[]peer.ID`, but needs testing.

kishansagathiya:
@hsanjuan 
In situations when replication factor and allocations don't match, say replication factor is 2, but allocations have more than 2 peers, which one should take priority? Allocations?

hsanjuan:
> @hsanjuan
> In situations when replication factor and allocations don't match, say replication factor is 2, but allocations have more than 2 peers, which one should take priority? Allocations? or error out?

The allocator code will choose enough allocations to satisfy the replication factor. Replication factor has priority.

hsanjuan:
Let's wait for #681 to be merged.

PR #649, opened 2019-01-18
Fix #593: Remove snaps

### LINKED ISSUE #593: Proposal to remove snapcraft builds from ipfs-cluster
#### Basic information

  * [ ] Bug
  * [ ] Feature request
  * [x] Enhancement

#### Description

Publishing snaps is a time sink for me due to their constant annoyances. Either builds break, or the credentials expire, or the store build requires manual approval, or, as now, they make us go and explain why we need `classic` confinement and after all the explanations they just don't seem to take action anyway (https://forum.snapcraft.io/t/classic-confinement-request-for-the-ipfs-cluster-snap/8056/8).

In the meantime, Travis is red. Actually we cannot even publish snaps on the `stable` channel from Travis because they only take `edge` when pushing from Travis, and they don't document any workaround. I am trying to setup their build pipeline but, it must be a joke, their DNS server is not there to resolve `ipfs.io`:

https://build.snapcraft.io/user/ipfs/ipfs-cluster/361367

We also have not much documentation about how to configure and nothing to orchestrate a snap cluster setup. Stats report 30 active devices for cluster: https://snapcraft.io/ipfs-cluster/metrics?period=6m

Ideally, we would hand this to a community member interested in maintaining this packaging format in a way that the Cluster team doesn't have to be fully commited to its maintenance, documentation and support.

@brianmcmichael you're the only person I can identify using snaps, please feel free to drop in your point of view here.

hsanjuan:
Snap builds have broken again. It seems the credentials have expired without
warning, even though they were not so old anyways. As promised,
next time snaps would break, they would be removed.

License: MIT


hsanjuan:
This should fix master builds, but the only way to test it is after merging.

Issue #654, opened 2019-01-24
Moving away from *Serial types


hsanjuan:
The `*Serial` types have been used to:

a) Provide JSON representation for types in the API output
b) Have a serializable-by-default type that can also be used by Msgpack so that RPC works.

It was an early thing and I probably didn't know better. This has cascaded into converting between serializable slices and original ones etc. A mess. There's also a rather big performance impact (converting between types), and the fact that it is quite difficult to understand why things are like they are and not mess them up.

Our main troubles with serialization are:

* `cid.Cid` not serializable by default -> needs to be converted to string
* `peer.ID` are a string, but need to be humanized for JSON output.
* `time.Time` in metrics (? unsure)

If we remove all `*Serial` types, we need to make sure:

* To get efficient and working go-codec/msgpack serialization of the main types.
* To get exactly the same JSON representation

I'd propose to:

* Make `cid.Cid` implement the `https://golang.org/pkg/encoding/` interface (horrible gx bubbling) so that go-codec knows how to handle it.
* Ensure `peer.ID` `MarshalJSON` works out of the box.

There is an extra problem with Cids and JSON, as by default they encode as `{ "/" : <cid> }` which imho is the most unfriendly json encoding you could choose. The easy path is to break the API (and get aligned with how Go IPFS broke their API). The hard path is to have MarhsalJSON methods for all our types.

hsanjuan:
Turns out protobuf supports Json https://developers.google.com/protocol-buffers/docs/proto3#json

hsanjuan:
tl;dr: made necessary changes to go-libp2p-peer and go-cid and bubbled. #688 removes *Serial types and make the standard ones serializable by default. Protobuf is only used for permanent storage of Pins in the datastore.

PR #655, opened 2019-01-25
State: Use go-datastore to implement the state interface

hsanjuan:
Since the beginning, we have used a Go map to store the shared state (pinset)
in memory. The mapstate knew how to serialize itself so that libp2p-raft would
know how to write to disk when it:

* Saved snapshots of the state on shutdown
* Sent the state to a newcomer peer

hashicorp.Raft assumes an in-memory state which is snapshotted from time to
time and read from disk on boot.

This commit adds a `dsstate` implementation of the state interface using
`go-datastore`. This allows to effortlessly switch to a disk-backed state in
the future (as we will need), and also have at our disposal the different
implementations and utilities of Datastore for fine-tuning (caching, batching
etc.).

`mapstate` has been reworked to use dsstate. Ideally, we would not even need
`mapstate`, as it would suffice to initialize `dsstate` with a
`MapDatastore`. BUT, we still need it separate to be able to auto-migrate to
the new format.

This will be the last migration with the current system. Once this has been
released and users have been able to upgrade we will just remove `mapstate` as
it is now.

License: MIT

Depends on https://github.com/libp2p/go-libp2p-raft/pull/50


hsanjuan:
Also, kind of depends on discussion #654 . If we slightly changed how things are  serialized soon, I'd rather not have to migrate right after this.

lanzafame:
LGTM. Only thing would be a better explanation of what the migration process is and how the migration process will be different in the future, i.e. post v6.

lanzafame:
@hsanjuan It seems that in some cases, this PR has slowed down cluster as there have now been 2 test timeout failures. I am happy to bump the timeout up again and then chip away at getting it down again.

hsanjuan:
> Also, kind of depends on discussion #654 . If we slightly changed how things are serialized soon, I'd rather not have to migrate right after this.

I don't fully know. But since checking state version will mean opening the datastore and reading one key, we know that will be different. The migration will be looping the keys and overwriting them with their updated form.

The key idea is that Marshal/Unmarshal will either not be used, or can happen normally regardless of the state version (unlike now).

hsanjuan:
> @hsanjuan It seems that in some cases, this PR has slowed down cluster as there have now been 2 test timeout failures. I am happy to bump the timeout up again and then chip away at getting it down again.

I only see jenkins, which I retriggered now. It make take a while before I merge this anyway.

Does this approval include changes in go-libp2p-raft ?

lanzafame:
@hsanjuan I just added some comments on the go-libp2p-raft PR.

hsanjuan:
@lanzafame do me a favour and have a look to the last commits (added tracing) and lmk if we can leave this ready for merge, as I want it to go before #670 .

cluster.go
```diff
@@ -739,20 +739,7 @@ func (c *Cluster) Join(ctx context.Context, addr ma.Multiaddr) error {
...
-		if err != nil {
-			logger.Error(err)
-		}
-		ch <- time.Now() // boostrap
-		defer close(ch)
-		select {
-		case <-dhtBstCtx.Done(): // shut down the process
-			proc.Close()
-		}
+		c.dht.BootstrapOnce(c.ctx, dht.DefaultBootstrapConfig)
```
lanzafame:
```suggestion
		c.dht.BootstrapOnce(ctx, dht.DefaultBootstrapConfig)
```
`ctx` has already been derived from the `c.ctx` at the top of the function.

api/types.go
```diff
@@ -704,10 +707,11 @@ func (pT PinType) String() string {
 
 // PinOptions wraps user-defined options for Pins
 type PinOptions struct {
-	ReplicationFactorMin int    `json:"replication_factor_min"`
-	ReplicationFactorMax int    `json:"replication_factor_max"`
-	Name                 string `json:"name"`
-	ShardSize            uint64 `json:"shard_size"`
+	ReplicationFactorMin int               `json:"replication_factor_min" codec:"rn,omitempty"`
```
lanzafame: @hsanjuan Is there a compile-time or test-time check that we can add so that we don't conflict with the codec variable?

GitCop:

 * Commit: 1491a01ec8028fd4e593bcff6a33cf66cce6d060
    **License: MIT  
Signed-off-by:**

---

GitCop:

 * Commit: 1491a01ec8028fd4e593bcff6a33cf66cce6d060
    **License: MIT  
Signed-off-by:**

---

api/pb/types.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+package api.pb;
+
+message Pin {
+  bytes Cid = 1;
+  enum PinType {
```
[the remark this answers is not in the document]
hsanjuan: I can't have an enum called `Type` and a field called `Type`

api/types.go
hsanjuan: As discussed, tests break in most cases when screwing up the codec field names.

api/pb/types.proto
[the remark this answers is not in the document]
hsanjuan: let's leave this as is, I don't mind so much and we can change it later.

api/types.go
hsanjuan: Additionally I have added a test that verifies that there are no dup fields.

hsanjuan:
@lanzafame I squashed, rebased and cleaned up. Then I tested with larger states and had to fix some issues (latest commit). Can you check again? I think we can merge then.

lanzafame:
@hsanjuan LGTM

PR #659, opened 2019-02-06
Fix/update json config

### LINKED ISSUE #658: Docker image issue
when start cluster via docker image it shows
16:52:09.448 WARNI     config: raft section is empty, generating default config.go:333
16:52:09.448 WARNI     config: restapi section is empty, generating default config.go:333
16:52:09.448 WARNI     config: ipfsproxy section is empty, generating default config.go:333
16:52:09.448 WARNI     config: ipfshttp section is empty, generating default config.go:333
16:52:09.448 WARNI     config: maptracker section is empty, generating default config.go:333
16:52:09.448 WARNI     config: stateless section is empty, generating default config.go:333
16:52:09.448 WARNI     config: pubsubmon section is empty, generating default config.go:333
16:52:09.448 WARNI     config: monbasic section is empty, generating default config.go:333
16:52:09.448 WARNI     config: disk section is empty, generating default config.go:333
16:52:09.448 WARNI     config: numpin section is empty, generating default config.go:333
16:52:09.448 WARNI     config: tracing section is empty, generating default config.go:333
16:52:09.448 WARNI     config: metrics section is empty, generating default config.go:333
16:52:09.448 WARNI     config: ipfshttp section is empty, generating default config.go:333

seems config is broken and IPFS_API is not applied
seems issue related to latest docker build https://hub.docker.com/r/ipfs/ipfs-cluster/tags

Errors
17:18:30.941  INFO  consensus: existing Raft state found! raft.InitPeerset will be ignored raft.go:204
17:18:30.941  INFO    restapi: REST API (HTTP): /ip4/127.0.0.1/tcp/9094 restapi.go:443
17:18:31.022 ERROR   ipfshttp: error posting to IPFS: Post http://127.0.0.1:5001/api/v0/id: dial tcp 127.0.0.1:5001: connect: connection refused ipfshttp.go:426

lanzafame:
Resolves #658.

jsonConfigs.getSection() returned a value when it needed to
return a pointer to the jsonSection fields inside the struct.

Even though the jsonSection type is a map, therefore on the heap,
returning it as a value (non-pointer) resulted in it being
disassociated with the jsonConfigs overarching struct.

License: MIT

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -38,18 +38,9 @@ func setupConfigManager() *config.Manager {
 	cfg := config.NewManager()
 	mockCfg := &mockCfg{}
 	cfg.RegisterComponent(config.Cluster, mockCfg)
-	cfg.RegisterComponent(config.API, mockCfg)
-	cfg.RegisterComponent(config.API, mockCfg)
-	cfg.RegisterComponent(config.IPFSConn, mockCfg)
-	cfg.RegisterComponent(config.Consensus, mockCfg)
-	cfg.RegisterComponent(config.PinTracker, mockCfg)
-	cfg.RegisterComponent(config.PinTracker, mockCfg)
-	cfg.RegisterComponent(config.Monitor, mockCfg)
-	cfg.RegisterComponent(config.Monitor, mockCfg)
-	cfg.RegisterComponent(config.Informer, mockCfg)
-	cfg.RegisterComponent(config.Informer, mockCfg)
-	cfg.RegisterComponent(config.Observations, mockCfg)
-	cfg.RegisterComponent(config.Observations, mockCfg)
+	for _, sect := range config.SectionTypes() {
+		cfg.RegisterComponent(sect, mockCfg)
+	}
 	return cfg
 }
 
@@ -73,12 +64,22 @@ func TestManager_ToJSON(t *testing.T) {
       "a": "b"
     }
   },
+  "state": {
+    "mock": {
+      "a": "b"
+    }
+  },
   "pin_tracker": {
     "mock": {
       "a": "b"
     }
   },
   "monitor": {
+    "mock": {
+      "a": "b"
+    }
+  },
+  "allocator": {
     "mock": {
       "a": "b"
     }
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -5,137 +5,106 @@ package config_test
 
 import (
 	"bytes"
-	"encoding/json"
-	"reflect"
 	"testing"
 
-	ipfscluster "github.com/ipfs/ipfs-cluster"
-	"github.com/ipfs/ipfs-cluster/api/ipfsproxy"
-	"github.com/ipfs/ipfs-cluster/api/rest"
 	"github.com/ipfs/ipfs-cluster/config"
-	"github.com/ipfs/ipfs-cluster/pintracker/maptracker"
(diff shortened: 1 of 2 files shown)


config/config_test.go
```diff
@@ -0,0 +1,141 @@
...
+	cfg.RegisterComponent(config.Cluster, clusterCfg)
+	cfg.RegisterComponent(config.API, apiCfg)
+	cfg.RegisterComponent(config.API, ipfsproxyCfg)
+	cfg.RegisterComponent(config.PinTracker, maptrackerCfg)
+	return cfg
+}
+
+func TestManager_ToJSON(t *testing.T) {
+	var want1 bytes.Buffer
+	json.Compact(&want1, []byte(`{
```
hsanjuan: Overkill? This just needs to test that sections are not empty. Otherwise it's stepping into component configurations (components are responsible their ToJSON() work properly). This module should be fully agnostic to component configurations (if ever we split into multiple repos we'll have a nice circular dep here).

Preferably it should be testing with a mock configuration implementation.

And I really don't want to have to change this every time the configuration changes somewhere.

I also disagree with using reflect (I don't fully understand why, having https://golang.org/pkg/bytes/#Equal).

Sorry for the work here, but overly complicated imho.

Ideally:

* Create mock config which ToJSONs() to something short like `{"a":"b"}`
* Register that to all SectionTypes() on a Config.
* Call ToJSON()
* Do the bytes equal (I guess) on the whole of it (unless ToJson comes unordered, in which case you need to extract section as now).

Not sure we need iterative testcases here (there is no different ways of behaving for ToJSON()), so I only see 1 testcase.

lanzafame: I use https://github.com/cweill/gotests to generate the test boilerplate, the default for the tool is to do table-driven and use reflect. I have no issue with changing from DeepEqual to bytes.Equal. I was mainly just trying to get some tests to confirm the fix.

config/config_test.go
```diff
@@ -5,137 +5,106 @@ package config_test
...
-				t.Errorf("api section not the same: got: %s want: %s", gotpintracker, wantpintracker)
-			}
-		})
-	}
+func setupConfigManager() *config.Manager {
+	cfg := config.NewManager()
+	mockCfg := &mockCfg{}
+	cfg.RegisterComponent(config.Cluster, mockCfg)
+	cfg.RegisterComponent(config.API, mockCfg)
+	cfg.RegisterComponent(config.API, mockCfg)
```
hsanjuan: Is there a reason there are dups lines?

You can do a `for _, sec := range SectionTypes() { cfg.RegisterComponent(sec, mockCfg) }`

hsanjuan: For all section except the `config.Cluster` one, that is.

hsanjuan:
I don't know why the docker compose thing is failing but I'll check

hsanjuan:
ah @lanzafame , docker compose is probably failing because the container `ipfs/ipfs-cluster:latest` is broken. Let's ignore this error as it should fix itself once we've merged.

config/config_test.go
hsanjuan: fixed myself

lanzafame: @hsanjuan thanks for the fix, I had just copied it across from `cmd/ipfs-cluster-service/config.go`.

Issue #650, opened 2019-01-18
Release 0.9.0


hsanjuan:
* #486 tracing
* #593 #649 remove snaps

lanzafame:
@hsanjuan #486 is now merged. From here I am guessing the steps are:

1. Merge any `RFM` labelled PRs, in this case #649.
2. Tag a release candidate with the following format `v0.9.0-rc1`
3. Do a cross platform build of the service and ctl binaries
4. Push those binaries to `dist.ipfs.io` somehow
5. Have the RC live for a period of time and then repeat from step 2-4 for full release

lanzafame:
- #658 #659 fix config ToJSON

hsanjuan:
* #676 fix bug introduced by tracing

hsanjuan:
* #671 rename config keys in observations/metrics

hsanjuan:
* #674 fix pprofs urls

lanzafame:
Closed by #677.

PR #676, opened 2019-02-15
Fix context cancelled too early.

hsanjuan:
Every 10th pin the ipfs connector calls repo stat.

Because the call is async, the context was cancelled immediately and failed.

@lanzafame for 0.9.0


lanzafame:
@hsanjuan can you please do the context wrapping as done in cluster.go
Shutdown()

hsanjuan:
Need to reference this in #650 and changelog

PR #681, opened 2019-02-20
New options in PinOptions

hsanjuan:
This starts handling Metadata and UserAllocations in the PinOptions
object. The Pin protobuf has been modified to embed a PinOptions message which
is defined separately.

Query arguments for the Metadata map are declared by their "meta-" prefix:
"?meta-something=something&meta-something-else-b".

Additional tests have been added, along with an Equals() method for
PinOptions.

License: MIT


api/pb/types.proto
```diff
@@ -14,10 +14,14 @@ message Pin {
...
-  map<string, string> Metadata = 12;
+  PinOptions Options = 6;
+}
+
+message PinOptions {
+  sint32 ReplicationFactorMin = 1;
+  sint32 ReplicationFactorMax = 2;
+  string Name = 3;
+  uint64 ShardSize = 4;
+  repeated string UserAllocations = 5;
```
lanzafame: @hsanjuan What is the current strategy for when the allocations that a user has specified are no longer available? I am not sure that this is worth keeping around long term?

api/types.go
```diff
@@ -707,21 +707,72 @@ func (pT PinType) String() string {
...
 type PinOptions struct {
 	ReplicationFactorMin int               `json:"replication_factor_min" codec:"rn,omitempty"`
 	ReplicationFactorMax int               `json:"replication_factor_max" codec:"rx,omitempty"`
 	Name                 string            `json:"name" codec:"n,omitempty"`
 	ShardSize            uint64            `json:"shard_size" codec:"s,omitempty"`
+	UserAllocations      []string          `json:"user_allocations" codec:"ua,omitempty"`
 	Metadata             map[string]string `json:"metadata" codec:"m,omitempty"`
 }
 
+// Equals returns true of two PinOption objects are equivalent.
```
lanzafame:
```suggestion
// Equals returns true if two PinOption objects are equivalent.
```

api/types.go
```diff
@@ -750,6 +801,26 @@ func (po *PinOptions) FromQuery(q url.Values) {
...
+		po.ShardSize = shsize
+	}
+
+	if allocs := q.Get("user-allocations"); allocs != "" {
+		po.UserAllocations = strings.Split(allocs, ",")
+	}
+
+	po.Metadata = make(map[string]string)
+	for k := range q {
+		if !strings.HasPrefix(k, pinOptionsMetaPrefix) {
```
lanzafame: What else are we potentially storing in the `metadata` map?

api/pb/types.proto
```diff
@@ -14,10 +14,14 @@ message Pin {
   repeated bytes Allocations = 3;
   sint32 MaxDepth = 4;
   bytes Reference = 5;
-  reserved 6,7;
```
lanzafame: Should we be removing the reserved fields?

api/pb/types.pb.go
```diff
@@ -55,19 +55,15 @@ func (Pin_PinType) EnumDescriptor() ([]byte, []int) {
 }
```
kishansagathiya: This is an auto-generated file which we are not suppose to edit.
Instead of adding it into our source code, shouldn't we add the logic to generate the file into `make build` and let it auto-generate everytime?

api/types.go
```diff
@@ -707,21 +707,72 @@ func (pT PinType) String() string {
...
+		return false
+	}
+
+	lenAllocs1 := len(po.UserAllocations)
+	lenAllocs2 := len(po2.UserAllocations)
+	if lenAllocs1 != lenAllocs2 {
+		return false
+	}
+
+	// avoid side effecs in the original objects
```
kishansagathiya:
```suggestion
	// avoid side effects in the original objects
```

May be a naive question, but what kind of side effects?

lanzafame:
It hasn't been edited, the protobuf definition has been updated.

api/pb/types.proto
hsanjuan: Nothing happens. Things will get preferentially assigned to those allocations and we keep them as an indicator of where things were meant to be pinned in the first place. Will this data ever be re-used for something, I'm not sure at this point...

I think we should keep the field in the protobuf, but we might want to simply not use it (never setting it). But in most cases it won't be used anyway.

api/types.go
hsanjuan: nothing. But the query arguments that go to the `metadata` map start with `meta-`. So if there is a query argument that isn't like that I don't want to put it as metadata.

api/pb/types.proto
hsanjuan: why not.. they were reserved as a sort of separator from the fields that corresponded to PinOptions, but now I don't see reason to reserve them.

api/pb/types.pb.go
hsanjuan: You have a point, but there are some benefits in committing the generated file,

* Generation process won't error or produce different results if something changed in the generation tool during builds
* Users don't need to have the protobuf toolchain installed to build cluster
* Users can bind against cluster libraries without having to "build" cluster first.

Committing the generated protobuf is the rule in our ecosystem afaik.

api/types.go
hsanjuan: The sorting of the Allocations slice. After calling Equal(), the slice would have changed inadvertently.

api/pb/types.proto
lanzafame: I still think it would be a good idea to reserve some fields in case we add anything to the `Pin` struct itself, otherwise it will go after all the `PinOption` fields

api/types.go
```diff
@@ -707,21 +707,72 @@ func (pT PinType) String() string {
...
 
+var pinOptionsMetaPrefix = "meta-"
+
 // PinOptions wraps user-defined options for Pins
 type PinOptions struct {
 	ReplicationFactorMin int               `json:"replication_factor_min" codec:"rn,omitempty"`
 	ReplicationFactorMax int               `json:"replication_factor_max" codec:"rx,omitempty"`
 	Name                 string            `json:"name" codec:"n,omitempty"`
 	ShardSize            uint64            `json:"shard_size" codec:"s,omitempty"`
+	UserAllocations      []string          `json:"user_allocations" codec:"ua,omitempty"`
```
kishansagathiya: `[]peer.ID` instead of `[]string`?

api/pb/types.proto
hsanjuan: Thinking again, if we're not going to use it, we may remove it from the protobuf and add it when we use it.

api/pb/types.pb.go
```diff
@@ -55,19 +55,15 @@ func (Pin_PinType) EnumDescriptor() ([]byte, []int) {
...
-	ReplicationFactorMin int32             `protobuf:"zigzag32,8,opt,name=ReplicationFactorMin,proto3" json:"ReplicationFactorMin,omitempty"`
-	ReplicationFactorMax int32             `protobuf:"zigzag32,9,opt,name=ReplicationFactorMax,proto3" json:"ReplicationFactorMax,omitempty"`
-	Name                 string            `protobuf:"bytes,10,opt,name=Name,proto3" json:"Name,omitempty"`
-	ShardSize            uint64            `protobuf:"varint,11,opt,name=ShardSize,proto3" json:"ShardSize,omitempty"`
-	Metadata             map[string]string `protobuf:"bytes,12,rep,name=Metadata,proto3" json:"Metadata,omitempty" protobuf_key:"bytes,1,opt,name=key,proto3" protobuf_val:"bytes,2,opt,name=value,proto3"`
-	XXX_NoUnkeyedLiteral struct{}          `json:"-"`
-	XXX_unrecognized     []byte            `json:"-"`
-	XXX_sizecache        int32             `json:"-"`
+	Cid                  []byte      `protobuf:"bytes,1,opt,name=Cid,proto3" json:"Cid,omitempty"`
+	Type                 Pin_PinType `protobuf:"varint,2,opt,name=Type,proto3,enum=api.pb.Pin_PinType" json:"Type,omitempty"`
```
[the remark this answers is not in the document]
hsanjuan: No, protobuf doesn't take a field name called Type and a type called Type.

api/types.go
```diff
@@ -858,18 +929,23 @@ func (pin *Pin) ProtoMarshal() ([]byte, error) {
...
 		Metadata:             pin.Metadata,
 	}
+
+	pbPin := &pb.Pin{
+		Cid:         pin.Cid.Bytes(),
+		Type:        convertPinType(pin.Type),
+		Allocations: allocs,
+		MaxDepth:    int32(pin.MaxDepth),
+		Reference:   pin.Reference.Bytes(),
+		Options:     opts,
```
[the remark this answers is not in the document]
hsanjuan: It doesn't store allocations twice. It stores User-requested-allocations on one side, and actual allocations on the other.

> Should we make PinOptions not a part of api.Pin and have two different structs,
one where we store expected state of a Pin (used only as input): PinOptions
one where we get actual state of a Pin (used only as output) : api.Pin ?

Things like ReplicationFactorMin are part of the options (input) but also part of the Pin (output). This is mostly why it is like it is. But we should not take `api.Pin` as input though (or use it as a wrapper to just carry a Cid). But not on this PR.

api/types.go
hsanjuan: My idea was that UserAllocations may also in the future provide peer names instead of peer IDs etc. We should be able to declare them as peer.ID since this exists: https://godoc.org/github.com/libp2p/go-libp2p-peer#ID.MarshalJSON but I want to change this in multiple places (so different PR).

hsanjuan:
> Need to modify PinOptions here
> 
> [ipfs-cluster/api/types.go](https://github.com/ipfs/ipfs-cluster/blob/add08da79c60ec19594e074c35d757b15fea86c8/api/types.go#L977-L1003)
> 
> Lines 977 to 1003 in [add08da](/ipfs/ipfs-cluster/commit/add08da79c60ec19594e074c35d757b15fea86c8)
>  func (pins PinSerial) ToPin() Pin { 
>  	c, err := cid.Decode(pins.Cid) 
>  	if err != nil { 
>  		logger.Debug(pins.Cid, err) 
>  	} 
>  	var ref cid.Cid 
>  	if pins.Reference != "" { 
>  		ref, err = cid.Decode(pins.Reference) 
>  		if err != nil { 
>  			logger.Warning(pins.Reference, err) 
>  		} 
>  	} 
>   
>  	return Pin{ 
>  		Cid:         c, 
>  		Allocations: StringsToPeers(pins.Allocations), 
>  		Type:        PinType(pins.Type), 
>  		MaxDepth:    pins.MaxDepth, 
>  		Reference:   ref, 
>  		PinOptions: PinOptions{ 
>  			Name:                 pins.Name, 
>  			ReplicationFactorMin: pins.ReplicationFactorMin, 
>  			ReplicationFactorMax: pins.ReplicationFactorMax, 
>  			ShardSize:            pins.ShardSize, 
>  		}, 
>  	} 
>  }

Good catch

api/pb/types.proto
hsanjuan: This will make pin options field take an extra byte for encoding. In my view it is not horrible that any new fields will come afterwards, at least not enough not to justify using an extra byte for PinOptions (which will always be used).

hsanjuan:
I have commented out the UserAllocations field from the protobuf. When we have a good reason to save this option field, we can add it. In the meantime we can also decide if it should be stored as []string or as [][]byte.

hsanjuan:
Since go-cid with `MarshalBinary` has been bubbled for us, after this PR I'll make a follow up one removing `Serial` types (I think we can get away now without them, the bad thing will be that cids in json will appear as `{ "/" : <cid> }`.

api/pb/types.proto
```diff
@@ -22,6 +22,6 @@ message PinOptions {
   sint32 ReplicationFactorMax = 2;
   string Name = 3;
   uint64 ShardSize = 4;
-  repeated string UserAllocations = 5;
+  reserved 5;
```
lanzafame: I suggest adding a comment stating that this field is reserved for `UserAllocations` and not just a general reserved field for possible future needs.

PR #685, opened 2019-02-22
Consensus: add new "crdt" consensus component

hsanjuan:
This adds a new "crdt" consensus component using go-ds-crdt.

This implies several refactors to fully make cluster consensus-component
independent:

* Delete mapstate and fully adopt dsstate (after people have migrated).
* Add a new "datastore" modules so that we can configure datastores in the
   main configuration like other components.
* Let the consensus components fully define the "state.State". Thus, they do
not receive the state, they receive the storage where we put the state (a
go-datastore).
* Allow to customize how the monitor component obtains Peers() (the current
  peerset), including avoiding using the current peerset. At the moment the
  crdt consensus uses the monitoring component to define the current peerset.
  Therefore the monitor component cannot rely on the consensus component to
  produce a peerset.
* Re-factor/re-implementation of "ipfs-cluster-service state"
  operations. Includes the dissapearance of the "migrate" one.

The CRDT consensus component defines:

* A broadcaster, using Pubsub
* A DAGSyncer, using IPFSBlockPut/Get from the IPFSConnector

With those and a given datastore, it creates a crdt-datastore. It then uses it
to intiialize a dssate. Thus the crdt-store is elegantly wrapped. Any
modifications to the state get automatically replicated using the Broadcaster
and the DAGSyncer.

This is highly WIP:

* No tracing
* No batching
* No options for badger
* ~~No re-use of pubsub,dht objects~~
* No dht topic peers discovery
* ~~Tests aren't fixed at all~~
* No tests added
* Not even tested with several peers

License: MIT


lanzafame:
@hsanjuan I am testing this locally, and as the PR currently stands I am not getting any convergence of peers (2 locally running peers). They converge when I restart the one the didn't receive the pin request, but they don't converge via broadcast deltas.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -707,6 +707,27 @@ func (ipfs *Connector) BlockGet(ctx context.Context, c cid.Cid) ([]byte, error)
 	return ipfs.postCtx(ctx, url, "", nil)
 }
 
+// FetchRefs asks IPFS to download blocks recursively to the given depth.
+// It discards the response, but waits until it completes.
+func (ipfs *Connector) FetchRefs(ctx context.Context, c cid.Cid, maxDepth int) error {
```
hsanjuan: Maybe remove

monitor/basic/peer_monitor.go
```diff
@@ -122,7 +131,11 @@ func (mon *Monitor) PublishMetric(ctx context.Context, m *api.Metric) error {
 		return nil
 	}
 
-	peers, err := mon.getPeers(ctx)
+	if mon.peers == nil {
```
hsanjuan: Can this happen?

monitor/metrics/store.go
```diff
@@ -69,6 +69,26 @@ func (mtrs *Store) Latest(name string) []*api.Metric {
 	return metrics
 }
 
+// AllMetrics returns the latest metrics for all peers and metrics types.  It
+// may return expired metrics.
+func (mtrs *Store) AllMetrics() []*api.Metric {
```
hsanjuan: Why is this? Remove if not used

monitor/basic/peer_monitor.go
```diff
@@ -175,36 +188,20 @@ func (mon *Monitor) PublishMetric(ctx context.Context, m *api.Metric) error {
...
+
+	if mon.peers == nil {
+		return latest
+	}
 
 	// Make sure we only return metrics in the current peerset
-	peers, err := mon.getPeers(ctx)
+	peers, err := mon.peers(ctx)
 	if err != nil {
 		return []*api.Metric{}
```
hsanjuan: return nil

state/dsstate/datastore.go
```diff
@@ -91,46 +92,49 @@ func (st *State) Rm(ctx context.Context, c cid.Cid) error {
...
 
 	q := query.Query{
 		Prefix: st.namespace.String(),
 	}
 
-	results, err := st.ds.Query(q)
+	results, err := st.dsRead.Query(q)
 	if err != nil {
-		return []*api.Pin{}
+		return []*api.Pin{}, err
```
hsanjuan: return nil

cluster.go
```diff
@@ -512,7 +498,8 @@ func (c *Cluster) Shutdown(ctx context.Context) error {
 		}
 	}
 
-	// We left the cluster or were removed. Destroy the Raft state.
+	// We left the cluster or were removed. Remove any consensus-specific
+	// state.
 	if c.removed && c.readyB {
 		err := c.consensus.Clean(ctx)
```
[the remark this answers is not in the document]
hsanjuan: Peers() will always forcefully include self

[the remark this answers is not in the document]
hsanjuan: From now.

monitor/basic/peer_monitor.go
hsanjuan: basic monitor is removed

hsanjuan: but yeah

monitor/basic/peer_monitor.go
hsanjuan: @lanzafame in the majority we SHOULD return the empty slice and not nil: https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices . nil slices encode to `null` in json, while empty slices encode to `[]`. These responses are used to encode JSON and no metrics should not be `null`, but rather `[]` .

monitor/metrics/store.go
hsanjuan: used in the Checker to alert on all expired metrics.

ipfsconn/ipfshttp/ipfshttp.go
hsanjuan: leaving commented. We will use it one day for something. No sense to have to re-implement.

api/types.go
```diff
@@ -698,7 +698,10 @@ func (pin *Pin) Equals(pin2 *Pin) bool {
 	}
 
 	if pin == pin2 {
-		return true
+		// ask @lanzafame why this is not true
+		// in any case, this is anomalous and we should
+		// not be using this with two nils.
+		return false
```
kishansagathiya: It is `true`
https://play.golang.org/p/yyDohd_hDPk

cmd/ipfs-cluster-service/daemon.go
```diff
@@ -271,3 +273,47 @@ func setupPinTracker(
...
+	checkErr("creating datastore", err)
+	return store
+}
+
+func setupConsensus(
+	name string,
+	h host.Host,
+	pubsub *pubsub.PubSub,
+	cfgs *cfgs,
+	store ds.ThreadSafeDatastore,
```
kishansagathiya: `ThreadSafeDatastore` is removed
https://github.com/ipfs/go-datastore/commit/9f529bc350a59b53118f6dcd37c7fa1be861d923

api/types.go
hsanjuan: we are conveying some meaning in saying that two `nil` pins are not equal when nil because, for a start, you have nothing to compare. At least I think that was @lanzafame's point.

cluster.go
```diff
@@ -635,6 +626,27 @@ func (c *Cluster) PeerAdd(ctx context.Context, pid peer.ID) (*api.ID, error) {
 		return id, err
 	}
 
+	// Send a ping metric so that the new node
+	// knows about this one at least
+	// T
```
hsanjuan: // T what. Remove

cluster.go
```diff
@@ -635,6 +626,27 @@ func (c *Cluster) PeerAdd(ctx context.Context, pid peer.ID) (*api.ID, error) {
...
+	m := &api.Metric{
```
hsanjuan: Potentially move to its own function that can be re-used from two places

cmd/ipfs-cluster-service/main.go
```diff
@@ -358,88 +361,103 @@ prints the state to stdout.
 						}
 						defer w.Close()
 
-						err = export(ctx, w)
-						checkErr("exporting state", err)
+						cfgMgr, cfgs := makeAndLoadConfigs()
+						defer cfgMgr.Shutdown()
+						mgr := newStateManager(c.String("consensus"), cfgs)
+						checkErr("exporting state", mgr.ExportState(w))
+						logger.Info("state successfully exported")
```
hsanjuan: We need to close the state when done with all these operations (badger specially).

consensus/crdt/config.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+var configKey = "crdt"
+var envConfigKey = "cluster_crdt"
+
+// Default configuration values
+var (
+	DefaultClusterName         = "ipfs-cluster"
+	DefaultPeersetMetric       = "ping"
+	DefaultDatastoreNamespace  = "/crdt"
```
hsanjuan: let's shorten this to `c`

consensus/crdt/config.go
```diff
@@ -0,0 +1,153 @@
...
+// ConfigKey returns the section name for this type of configuration.
+func (cfg *Config) ConfigKey() string {
+	return configKey
+}
+
+// Validate returns an error if the configuration has invalid values.
+func (cfg *Config) Validate() error {
+	if cfg.ClusterName == "" {
+		return errors.New("crdt.cluster_name cannot be empty")
+
```
hsanjuan: extra line breaks around here

consensus/crdt/consensus.go
```diff
@@ -0,0 +1,430 @@
...
+
+	if sub := css.subscription; sub != nil {
+		sub.Cancel()
+	}
+
+	if crdt := css.crdt; crdt != nil {
+		crdt.Close()
+	}
+
+	if store := css.store; store != nil {
```
hsanjuan: remove this, as it does break things.

consensus/raft/config.go
```diff
@@ -28,6 +28,7 @@ var (
 	DefaultNetworkTimeout       = 10 * time.Second
 	DefaultCommitRetryDelay     = 200 * time.Millisecond
 	DefaultBackupsRotate        = 6
+	DefaultDatastoreNamespace   = "/raft"
```
hsanjuan: shorten to `r`

monitor/metrics/checker.go
```diff
@@ -32,20 +32,46 @@ func NewChecker(metrics *Store) *Checker {
 	}
 }
 
-// CheckPeers will trigger alerts for expired metrics belonging to the
-// given peerset.
+// CheckPeers will trigger alerts for expired, latest metrics belonging to the
+// given peerset and not alert has been sent before.
```
hsanjuan: no alert

monitor/metrics/checker.go
```diff
@@ -32,20 +32,46 @@ func NewChecker(metrics *Store) *Checker {
...
+				return err
 			}
 		}
 	}
 	return nil
+
+}
+
+// CheckAll will trigger alerts for all latest metrics when expired
+// and not alert has been sent before.
```
hsanjuan: no alert

cluster.go
hsanjuan: actually cannot do that because this is remotely logging a metric.

consensus/crdt/consensus.go
```diff
@@ -0,0 +1,410 @@
...
+		return
+	}
+
+	dagSyncer := newLiteDAGSyncer(css.ctx, ipfs)
+
+	broadcaster, err := crdt.NewPubSubBroadcaster(
+		css.ctx,
+		css.pubsub,
+		css.config.ClusterName,
+	) // subscription name
```
lanzafame: comment should be at the end of the line above

consensus/crdt/consensus.go
```diff
@@ -0,0 +1,410 @@
...
+			"",
+			"Cluster",
+			"Track",
+			pin,
+			&struct{}{},
+		)
+		if err != nil {
+			logger.Error(err)
+		}
+		//css.crdt.PrintDAG()
```
lanzafame:
```suggestion
```

consensus/crdt/consensus.go
```diff
@@ -0,0 +1,410 @@
...
+}
+
+// LogUnpin removes a pin from the shared state.
+func (css *Consensus) LogUnpin(ctx context.Context, pin *api.Pin) error {
+	return css.state.Rm(ctx, pin.Cid)
+}
+
+// Peers returns the current known peerset. It uses
+// the monitor component and considers every peer with
+// valid known metrics a memeber.
```
lanzafame:
```suggestion
// valid known metrics a member.
```

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -710,6 +710,27 @@ func (ipfs *Connector) BlockGet(ctx context.Context, c cid.Cid) ([]byte, error)
 	return ipfs.postCtx(ctx, url, "", nil)
 }
 
+// // FetchRefs asks IPFS to download blocks recursively to the given depth.
```
lanzafame: Why is this commented out?

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -736,7 +757,7 @@ func (ipfs *Connector) updateInformerMetric(ctx context.Context) error {
 	var metric api.Metric
 
 	err := ipfs.rpcClient.GoContext(
-		ctx,
+		ipfs.ctx,
```
lanzafame: `ctx` already wraps `ipfs.ctx` above.

cluster.go
```diff
@@ -435,6 +421,10 @@ This might be due to one or several causes:
 	}
 
 	// Cluster is ready.
+
+	// Bootstrap the DHT now that we possibly have some connections
```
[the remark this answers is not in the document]
hsanjuan: When consensus becomes ready is because Raft has chosen a leader (therefore it has bootstrapped if it needs to bootstrap). 

This may not be the case for crdts but we do trigger another round of bootstrap during `Join` so I think things might work as they are (we'll have to test).

cluster.go
```diff
@@ -627,14 +618,34 @@ func (c *Cluster) PeerAdd(ctx context.Context, pid peer.ID) (*api.ID, error) {
...
 	}
 
+	// Send a ping metric to the new node directly so
+	// it knows about this one at least
+	m := &api.Metric{
+		Name:  pingMetricName,
+		Peer:  c.id,
+		Valid: true,
+	}
+	m.SetTTL(c.config.MonitorPingInterval * 2)
```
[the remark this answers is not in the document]
hsanjuan: We always set the metric TTL to be twice the ping interval (in line with the code broadcasting metrics).

cluster_test.go
```diff
@@ -494,15 +493,15 @@ func TestUnpinShard(t *testing.T) {
 // }
 
 // func TestClusterPinMeta(t *testing.T) {
```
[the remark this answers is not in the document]
hsanjuan: I have another PR where I took care of these...

clusterhost.go
```diff
@@ -5,22 +5,53 @@ import (
...
+	routedhost "github.com/libp2p/go-libp2p/p2p/host/routed"
 	ma "github.com/multiformats/go-multiaddr"
 )
 
-// NewClusterHost creates a libp2p Host with the options from the
-// provided cluster configuration.
-func NewClusterHost(ctx context.Context, cfg *Config) (host.Host, error) {
+// NewClusterHost creates a libp2p Host with the options from the provided
+// cluster configuration. Using that host, it creates pubsub and a DHT
+// instances, for shared used by all cluster components. The returned host uses
```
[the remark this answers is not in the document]
hsanjuan: Fixed in my commit

cmd/ipfs-cluster-service/daemon.go
```diff
@@ -98,25 +90,23 @@ func daemon(c *cli.Context) error {
...
-	tracer, err := observations.SetupTracing(cfgs.tracingCfg)
-	checkErr("setting up Tracing", err)
+) (*ipfscluster.Cluster, ds.Datastore, error) {
 
-	host, err := ipfscluster.NewClusterHost(ctx, cfgs.clusterCfg)
+	host, pubsub, dht, err := ipfscluster.NewClusterHost(ctx, cfgs.clusterCfg)
 	checkErr("creating libP2P Host", err)
 
 	peerstoreMgr := pstoremgr.New(host, cfgs.clusterCfg.GetPeerstorePath())
+	// Import peers but do not connect. We cannot connect to peers until
```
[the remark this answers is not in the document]
hsanjuan: We don't, we let them open connections as needed (or during bootstrap). Clarified in comment.

cmd/ipfs-cluster-service/daemon.go
```diff
@@ -98,25 +90,23 @@ func daemon(c *cli.Context) error {
...
 	c *cli.Context,
 	cfgs *cfgs,
 	raftStaging bool,
-) (*ipfscluster.Cluster, error) {
-	err := observations.SetupMetrics(cfgs.metricsCfg)
-	checkErr("setting up Metrics", err)
-
-	tracer, err := observations.SetupTracing(cfgs.tracingCfg)
-	checkErr("setting up Tracing", err)
+) (*ipfscluster.Cluster, ds.Datastore, error) {
```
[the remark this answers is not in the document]
hsanjuan: Because we need to close the datastore after we shutdown the Cluster. Another possibility is
passing it to the main component and letting it Close it on shutdown (but that's all the main component would do with it). Would this be better? I'm leaning to say yes...

cmd/ipfs-cluster-service/main.go
```diff
@@ -293,60 +306,50 @@ configuration.
...
-						err := locker.lock()
-						checkErr("acquiring execution lock", err)
-						defer locker.tryUnlock()
-
-						err = upgrade(ctx)
-						checkErr("upgrading state", err)
-						return nil
-					},
-				},
+				//				{
```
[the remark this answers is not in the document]
hsanjuan: ok

datastore/badger/config.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// The folder for this datastore. Leave empty for default.
+	Folder string
+}
+
+type jsonConfig struct {
+	Folder string `json:"folder,omitempty"`
+}
+
+// ConfigKey returns a human-friendly identifier for this type of Metric.
```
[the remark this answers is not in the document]
hsanjuan: fixed in my commit

datastore/badger/config.go
```diff
@@ -0,0 +1,102 @@
...
+const (
+	DefaultSubFolder = "badger"
+)
+
+// Config is used to initialize a BadgerDB datastore. It implements the
+// ComponentConfig interface.
+type Config struct {
+	config.Saver
+
+	// The folder for this datastore. Leave empty for default.
```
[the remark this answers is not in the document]
hsanjuan: I changed the approach to be more in line with the empty json defaults that we use in other places. So now the folder defaults to "badger" and not to an empty value.

consensus/crdt/dagsyncer.go
```diff
@@ -0,0 +1,72 @@
...
+	ctx, cancel := context.WithTimeout(ctx, lds.timeout)
+	defer cancel()
+	return lds.getter.GetMany(ctx, cs)
+}
+
+func (lds *liteDAGSyncer) Add(ctx context.Context, n ipld.Node) error {
+	ctx, cancel := context.WithTimeout(ctx, lds.timeout)
+	defer cancel()
+	return lds.dags.Add(ctx, n)
+}
```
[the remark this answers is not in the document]
hsanjuan: fixed in my commit

consensus/crdt/consensus.go
hsanjuan: ok

consensus/crdt/consensus.go
hsanjuan: ok

consensus/crdt/consensus.go
hsanjuan: ok

ipfsconn/ipfshttp/ipfshttp.go
hsanjuan: not used. 60% sure that we will use it one day for something...

ipfsconn/ipfshttp/ipfshttp.go
hsanjuan: ok

cmd/ipfs-cluster-service/daemon.go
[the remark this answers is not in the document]
lanzafame: I would prefer passing it to the main component as well. I think the main component should manage the lifecycle of all the cluster components/dependencies.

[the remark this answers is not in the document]
lanzafame: @hsanjuan this is the only thing I would like to see changed before approving.

monitor/pubsubmon/pubsubmon.go
```diff
@@ -124,7 +127,7 @@ func (mon *Monitor) logFromPubsub() {
 				logger.Error(err)
 				continue
 			}
-			logger.Debugf(
+			logger.Infof(
```
kishansagathiya: Why?
This overwhelms console logs, making it difficult to see useful information

hsanjuan: ops this is a leftover

hsanjuan:
> Test Failures
> 
> ```
> === RUN   TestClustersSyncLocal
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xd48c11]
> ```
> 
> https://travis-ci.com/ipfs/ipfs-cluster/jobs/193608601#L577

I think I have fixed this on go-ds-crdt (has to do with using the datastore after it was closed).

hsanjuan:
> Sanity check works. I see `Pin` and `Unpin` quickly happening on both nodes.
> 
> I see this error always
> 
> ```
> 16:02:28.368 ERROR       crdt: error hashing topic: Function not implemented. Complain to lib maintainer. consensus.go:114
> ```
> 
> Don't see any bad effects so far because of this. But it conveys that something is not working as expected.
> Can we get rid of it or create an issue?

hmm good catch, this is because of half baked support for md5 in multihash... fixing that now...

cluster_test.go
```diff
@@ -138,25 +138,25 @@ type mockTracer struct {
 	mockComponent
 }
 
-func testingCluster(t *testing.T) (*Cluster, *mockAPI, *mockConnector, state.State, PinTracker) {
-	clusterCfg, _, _, _, consensusCfg, maptrackerCfg, statelesstrackerCfg, psmonCfg, _, _ := testingConfigs()
+func testingCluster(t *testing.T) (*Cluster, *mockAPI, *mockConnector, PinTracker) {
+	clusterCfg, _, _, _, _, raftCfg, _, maptrackerCfg, statelesstrackerCfg, psmonCfg, _, _ := testingConfigs()
```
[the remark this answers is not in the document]
hsanjuan: `cluster_test.go` has single peer-tests aimed to test the main component. The actual consensus component used is not really important, but will switch it if crdt becomes the default.

test/ipfs_mock.go
```diff
@@ -197,7 +206,10 @@ func (m *IpfsMock) handler(w http.ResponseWriter, r *http.Request) {
 		if err != nil {
 			goto ERROR
 		}
-		ok = m.pinMap.Has(ctx, c)
+		ok, err = m.pinMap.Has(ctx, c)
+		if err != nil {
+			goto ERROR
```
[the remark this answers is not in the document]
hsanjuan: I think it's fine, I'm happy with this `goto`. Removing it would not mean more readability.

kishansagathiya:
woohoo

eiTanLaVi:
Q: if we use the --upgrade flag, on v0.10.1, will the --consensus crdt flag take affect when restarting the daemon (systemd) ? what is the safe way to upgrade to this flag on a running cluster?

hsanjuan:
No.

First, `--consensus crdt` is not available on `v0.10.1`, it's on master (even though `ipfs-cluster-service` will report 0.10.1 on master  - we need to change that).

On the released `v0.10.1` (and `v0.10.0) the `--upgrade` flag does the necessary to run Raft with the new state formats when the user was running `v0.9.x` and below.

If you want to run with `--consensus crdt` using the old pinset that was running before with Raft, you need to do `ipfs-cluster-service state export --consensus raft` to extract that pinset and `state import --consensus crdt` to import it to the crdt side of things, on `master`. It won't work to export with a previous version and import with the newer either.

PR #690, opened 2019-03-01
ipfsproxy: use PinPath to match IPFS behaviour

hsanjuan:
License: MIT

### DIFF
--- a/rpc_api.go
+++ b/rpc_api.go
@@ -55,8 +55,8 @@ func (rpcapi *RPCAPI) PinPath(ctx context.Context, in *api.PinPath, out *api.Pin
 }
 
 // UnpinPath resolves path into a cid and runs Cluster.Unpin().
-func (rpcapi *RPCAPI) UnpinPath(ctx context.Context, in string, out *api.Pin) error {
-	pin, err := rpcapi.c.UnpinPath(ctx, in)
+func (rpcapi *RPCAPI) UnpinPath(ctx context.Context, in *api.PinPath, out *api.Pin) error {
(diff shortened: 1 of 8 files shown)


test/rpc_api_mock.go
```diff
@@ -54,20 +55,50 @@ func (mock *mockService) Unpin(ctx context.Context, in *api.Pin, out *struct{})
...
+		}
+		if c.Equals(ErrorCid) {
+			return ErrBadCid
+		}
+		pin = api.PinWithOpts(c, in.PinOptions)
+	} else {
+		pin = api.PinWithOpts(CidResolved, in.PinOptions)
+	}
+
+	*out = *pin
```
kishansagathiya: This part can be moved to a function and be used in both PinPath() and UnpinPath()

rpc_api.go
```diff
@@ -55,8 +55,8 @@ func (rpcapi *RPCAPI) PinPath(ctx context.Context, in *api.PinPath, out *api.Pin
 }
 
 // UnpinPath resolves path into a cid and runs Cluster.Unpin().
-func (rpcapi *RPCAPI) UnpinPath(ctx context.Context, in string, out *api.Pin) error {
-	pin, err := rpcapi.c.UnpinPath(ctx, in)
+func (rpcapi *RPCAPI) UnpinPath(ctx context.Context, in *api.PinPath, out *api.Pin) error {
```
kishansagathiya: How can unpin utilize pinOptions?

hsanjuan: it doesn't, but for parallelism (allows using same code to call pin/unpin) and if at some point in history we can unpin with options it's marginally better this way.

PR #699, opened 2019-03-05
Hide `extract_headers_path`, `extract_headers_ttl`

### LINKED ISSUE #698: proxy: "extract_headers_path" and "extract_headers_ttl" should be hidden options in the config
@kishansagathiya can you take care of this one? I think it's just setting the default to ""  for the path. For the ttl, probably set to 0 and omit set manually to "" if value is 0 (instead of doing String()).

kishansagathiya:
proxy: "extract_headers_path" and "extract_headers_ttl" should be
hidden options in the config

Fixes #698

### DIFF
--- a/api/ipfsproxy/config.go
+++ b/api/ipfsproxy/config.go
@@ -258,8 +258,12 @@ func (cfg *Config) toJSONConfig() (jcfg *jsonConfig, err error) {
 	jcfg.NodeHTTPS = cfg.NodeHTTPS
 
 	jcfg.ExtractHeadersExtra = cfg.ExtractHeadersExtra
-	jcfg.ExtractHeadersPath = cfg.ExtractHeadersPath
-	jcfg.ExtractHeadersTTL = cfg.ExtractHeadersTTL.String()
+	if cfg.ExtractHeadersPath != DefaultExtractHeadersPath {
+		jcfg.ExtractHeadersPath = cfg.ExtractHeadersPath
+	}
+	if ttl := cfg.ExtractHeadersTTL; ttl != DefaultExtractHeadersTTL {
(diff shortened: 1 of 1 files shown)


api/ipfsproxy/config.go
```diff
@@ -250,8 +258,14 @@ func (cfg *Config) toJSONConfig() (jcfg *jsonConfig, err error) {
 	jcfg.NodeHTTPS = cfg.NodeHTTPS
 
 	jcfg.ExtractHeadersExtra = cfg.ExtractHeadersExtra
-	jcfg.ExtractHeadersPath = cfg.ExtractHeadersPath
-	if !(cfg.ExtractHeadersTTL > 0) {
+	if cfg.ExtractHeadersPath == DefaultExtractHeadersPath {
```
hsanjuan: The first case is not necessary because the json object is empty. So the `if` should only have the second case.

api/ipfsproxy/config.go
```diff
@@ -250,8 +258,14 @@ func (cfg *Config) toJSONConfig() (jcfg *jsonConfig, err error) {
...
+		jcfg.ExtractHeadersPath = ""
+	} else {
+		jcfg.ExtractHeadersPath = cfg.ExtractHeadersPath
+	}
+
+	if cfg.ExtractHeadersTTL == DefaultExtractHeadersTTL {
+		jcfg.ExtractHeadersTTL = ""
 		return
```
hsanjuan: Same as before. Keep consistency.

`if ttl := cfg.ExtractHeadersTTL; ttl != DefaultExtractHeadersTTL { jcfg.ExtractHeadersTTL = ttl.String() }`

do not return

PR #700, opened 2019-03-06
Types: make AddedOutput carry a cid.Cid

hsanjuan:
This was a leftover. For consisency, all CIDs coming out of the API
should have the canonical cid form ( { "/": "cid" } ), otherwise
we are inconsistent.

### DIFF
--- a/api/add.go
+++ b/api/add.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/url"
 	"strconv"
+
+	cid "github.com/ipfs/go-cid"
 )
 
 // DefaultShardSize is the shard size for params objects created with DefaultParams().
@@ -13,10 +15,10 @@ var DefaultShardSize = uint64(100 * 1024 * 1024) // 100 MB
 // AddedOutput carries information for displaying the standard ipfs output
 // indicating a node of a file has been added.
 type AddedOutput struct {
-	Name  string `json:"name"`
-	Cid   string `json:"cid,omitempty"`
-	Bytes uint64 `json:"bytes,omitempty"`
-	Size  uint64 `json:"size,omitempty"`
+	Name  string  `json:"name" codec:"n,omitempty"`
+	Cid   cid.Cid `json:"cid,omitempty" codec:"c"`
+	Bytes uint64  `json:"bytes,omitempty" codec:"b,omitempty"`
+	Size  uint64  `json:"size,omitempty" codec:"s,omitempty"`
 }
 
 // AddParams contains all of the configurable parameters needed to specify the
--- a/api/add.go
+++ b/api/add.go
@@ -16,7 +16,7 @@ var DefaultShardSize = uint64(100 * 1024 * 1024) // 100 MB
 // indicating a node of a file has been added.
 type AddedOutput struct {
 	Name  string  `json:"name" codec:"n,omitempty"`
-	Cid   cid.Cid `json:"cid,omitempty" codec:"c"`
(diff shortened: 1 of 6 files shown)


api/add.go
```diff
@@ -13,10 +15,10 @@ var DefaultShardSize = uint64(100 * 1024 * 1024) // 100 MB
 // AddedOutput carries information for displaying the standard ipfs output
 // indicating a node of a file has been added.
 type AddedOutput struct {
-	Name  string `json:"name"`
-	Cid   string `json:"cid,omitempty"`
-	Bytes uint64 `json:"bytes,omitempty"`
-	Size  uint64 `json:"size,omitempty"`
+	Name  string  `json:"name" codec:"n,omitempty"`
+	Cid   cid.Cid `json:"cid,omitempty" codec:"c"`
```
kishansagathiya: Does `omitempty` work for struct fields?

api/add.go
```diff
@@ -13,10 +15,10 @@ var DefaultShardSize = uint64(100 * 1024 * 1024) // 100 MB
 // AddedOutput carries information for displaying the standard ipfs output
 // indicating a node of a file has been added.
 type AddedOutput struct {
-	Name  string `json:"name"`
-	Cid   string `json:"cid,omitempty"`
-	Bytes uint64 `json:"bytes,omitempty"`
-	Size  uint64 `json:"size,omitempty"`
+	Name  string  `json:"name" codec:"n,omitempty"`
```
kishansagathiya: You want to omit this from codec, but not from json, why?

api/rest/restapi_test.go
```diff
@@ -447,7 +447,7 @@ func TestAPIAddFileEndpointLocal(t *testing.T) {
 		makeStreamingPost(t, rest, localURL, body, mpContentType, &resp)
 
 		// resp will contain the last object from the streaming
-		if resp.Cid != test.ShardingDirBalancedRootCID {
+		if resp.Cid.String() != test.ShardingDirBalancedRootCID {
 			t.Error("Bad Cid after adding: ", resp.Cid)
```
kishansagathiya:
```suggestion
			t.Error("Bad Cid after adding: ", resp.Cid.String())
```

api/rest/restapi_test.go
```diff
@@ -512,7 +512,7 @@ func TestAPIAddFileEndpoint_StreamChannelsFalse(t *testing.T) {
 
 		makePostWithContentType(t, rest, shardURL, fullBody, mpContentType, &resp)
 		lastHash := resp[len(resp)-1]
-		if lastHash.Cid != test.ShardingDirBalancedRootCID {
+		if lastHash.Cid.String() != test.ShardingDirBalancedRootCID {
 			t.Error("Bad Cid after adding: ", lastHash.Cid)
```
kishansagathiya:
```suggestion
			t.Error("Bad Cid after adding: ", lastHash.Cid.String())
```

I think lastHash.Cid would print something like `{"Qm...."}`

api/add.go
hsanjuan: removed `omitempty`

api/add.go
hsanjuan: not sure. I don't think we use the codec anyway. The json always carrying `Name` is probably because that's what ipfs did with the equivalent type. It's nice if things come out of the API with Name and Cid at least, even if name is empty.

api/rest/restapi_test.go
hsanjuan: I don't think so. It would use `String()`.

PR #706, opened 2019-03-08
Enable go.mod for ipfs-cluster

hsanjuan:


.travis.yml
```diff
@@ -12,9 +12,14 @@ sudo: required
 cache:
   directories:
     - $GOPATH/src/gx
+    - $GOPATH/pkg/mod
+    - $HOME/.cache/go-build
 
 install: true
 
+env:
```
lanzafame: This should follow the [gomod/ci-example](https://github.com/ipfs/ci-helpers/blob/master/travis-ci/travis.example.yml)

Makefile
```diff
@@ -15,52 +15,56 @@ clean: rwundo clean_sharness
...
 	$(MAKE) -C cmd/ipfs-cluster-service ipfs-cluster-service
-ctl: deps
+ctl: gx-deps
 	$(MAKE) -C cmd/ipfs-cluster-ctl ipfs-cluster-ctl
 
 gx-clean: clean
 	$(MAKE) -C $(deptools) gx-clean
 
 gx:
+ifeq ($(GX),yes)
```
lanzafame: A comment somewhere saying the env var `GX` needs to be set to `yes` for the gx make targets to work.

.travis.yml
hsanjuan: no, we have a proper multi-stage build. Plus not making cluster CI depend on a complex external script that can change under our feet. ok for small libraries.

lanzafame: @hsanjuan actually forget about using the example, but there ci-helpers script does have the functionality of untangling gx from go-mod so that things run in the proper order and things are rewritten when they shouldn't be.

lanzafame:
Happy for this to be merged once #708 is merged into this branch.

PR #713, opened 2019-03-11
Change the underlying data structure of a metrics.Window

lanzafame:
Replace the underlying slice in metrics.Window with a ring.Ring and add Distribution method for use by accural failure detection, which will follow in the next PR.


monitor/metrics/window.go
```diff
@@ -28,52 +37,76 @@ func NewWindow(windowCap int) *Window {
...
-	}
+	m.TS = time.Now().UnixNano()
+
+	mw.wMu.Lock()
+	mw.window.Value = m
+	mw.window = mw.window.Next()
+	mw.wMu.Unlock()
+
+	mw.lMu.Lock()
+	mw.last = m
```
hsanjuan: Can't `mw.window.Value` the same as `mw.last`? 

I mean, if instead of doing `Next()` after inserting, you do it before inserting, you can use `Value` as last value and save a variable and a mutex. (At least at first sight).

monitor/metrics/window.go
```diff
@@ -28,52 +37,76 @@ func NewWindow(windowCap int) *Window {
...
+// Distribution returns the deltas between all the current
+// values contained in the current window. This will
+// only return values if the api.Metric.Type() is "ping",
+// which are used for accural failure detection.
+func (mw *Window) Distribution() []int64 {
+	ms := mw.All()
+	dist := make([]int64, 0, len(ms)-1)
+	for i, v := range ms {
+		// the last value can't be used to calculate a delta
+		if i == len(ms)-1 {
```
hsanjuan: `for i, v := range ms[0:len(ms)-1])` ?

api/types.go
```diff
@@ -768,12 +768,15 @@ func (n *NodeWithMeta) Size() uint64 {
...
 // the Value, which should be interpreted by the PinAllocator.
+// The TS value is a timestamp representing when a peer has received
+// the metric value.
 type Metric struct {
 	Name   string  `json:"name" codec:"n,omitempty"`
 	Peer   peer.ID `json:"peer" codec:"p,omitempty"`
 	Value  string  `json:"value" codec:"v,omitempty"`
 	Expire int64   `json:"expire" codec:"e,omitempty"`
 	Valid  bool    `json:"valid" codec:"d,omitempty"`
+	TS     int64   `json:"ts" codec:"t,omitempty"` // TS contains a UnixNano timestamp
```
hsanjuan: Seems you are making a distribution using the times metrics were received. Wouldn't it be better (or the same) to use the deltas between the time they were sent (and for that you can use the `Expire` field)?

Alternatively, we should name this `ReceivedAt` or something more self-explanatory.

lanzafame: It needs to be between the received times as then it takes into account fluctuations in network latencies. There technical isn't going to be any difference between the expired value unless the peer disappears. 

Agreed with changing it to `ReceivedAt`.

hsanjuan:
@lanzafame one unaddressed comment  still

monitor/metrics/window.go
lanzafame: @hsanjuan addressed

hsanjuan:
RFM if tests pass

hsanjuan:
@lanzafame this looks fine doesn't it? without copying slices etc..

lanzafame:
rebased on top of #685.

hsanjuan:
@lanzafame I understand we can close this because it's in #714 ?

lanzafame:
@hsanjuan #714 should be merged into this one.

hsanjuan:
Speaking of "let's do small PRs from now"...

hsanjuan:
@lanzafame I can't get tests to pass here, specially the "replication overall" one. Can you have another look?

hsanjuan:
I'm going to spend time fixing tests in general

Issue #715, opened 2019-03-11
`peers ls` (and possibly other broadcast commands) fail when a peer is down


hsanjuan:
#### Pre-check

* [x] This is not a IPFS Cluster website content issue ([file those here](https://github.com/ipfs/ipfs-cluster-website/issues))
* [x] I read the [troubleshooting section](https://cluster.ipfs.io/documentation/troubleshooting/) of the website and it did not help
* [x] I searched for similar issues in the repo without luck
* [x] All my peers are running the same cluster version
* [x] All my peers are configured using the same cluster `secret`

#### Basic information

* [x] Version information (mark as appropiate):
  * [ ] Master
  * [ ] Release candidate for next version
  * [x] Latest stable version
  * [ ] An older version I should not be using
* [x] Type (mark as appropiate):
  * [x] Bug
  * [ ] Feature request
  * [ ] Enhancement
* [x] Operating system (mark as appropiate):
  * [x] Linux
  * [x] macOS
  * [x] Windows
  * [ ] Other: `which?`
* [x] Installation method (mark as appropiate):
  * [x] Binaries from dist.ipfs.io
  * [x] Built from sources
  * [x] Docker
  * [ ] Snap
  * [ ] Other: `which?`

#### Description

Pretty sure this is null-multiaddress related: https://discuss.ipfs.io/t/has-anyone-seen-this-error-cannot-unmarshal-object-into-go-struct-field/5007/4

70% sure that fix involves fixing deserialization of our `types.go/Multiaddr` or is somewhat related (either Undef cids or multiaddresses).

kishansagathiya:
> 70% sure that fix involves fixing deserialization of our types.go/Multiaddr or is somewhat related (either Undef cids or multiaddresses).

Yup, failing during unmarshaling https://github.com/ipfs/ipfs-cluster/blob/508131081124bc8289b762ffb7e0e0572798e9be/api/rest/client/request.go#L118

kishansagathiya:
This is happening because if a peer in down, on `peer ls` IPFSID is empty. 
```
type IPFSID struct {
	ID        peer.ID     `json:"id" codec:"i,omitempty"`
	Addresses []Multiaddr `json:"addresses" codec:"a,omitempty"`
	Error     string      `json:"error" codec:"e,omitempty"`
}
```

If I just omitempty `ID` from `IPFSID`, thing would work with a message for the peer which is down, like
```
[...]

PR #719, opened 2019-03-14
Bug fix: input isn't valid multihash

kishansagathiya:
Some command were failing with `input isn't valid multihash`, due to
empty peer.ID. This commit omit peer id from IPFSID when empty, thus
avoiding the error state

#715


hsanjuan:
It seems `Peers()` wraps errors with empty ID objects with the `Error` field set, but these cannot be re-parsed after convering to json because IPFSID.ID is empty? Is this the case? Probably the whole IPFSID object should be omitted in these cases because it does not make sense to marshal any part of it (might need to make it pointer and leave it nil).

There also the case when the IPFS daemon errors and IPFSID carries the error only. That will be fixed by the `omitempty` here. 

The `PinInfo` object also carries a `peer.ID` and an Error. We should make sure the `Peer` field is always set, even when it is used to wrap an error, or that `omitempty` is also used for that field.

kishansagathiya:
> It seems Peers() wraps errors with empty ID objects with the Error field set, but these cannot be re-parsed after convering to json because IPFSID.ID is empty? Is this the case?

Yes

kishansagathiya:
There isn't any problem because of peerID not being empty in `PinInfo`(may be because it is used at pointer at most places), at least in this scenario. I can `omitempty` it to be safe. But if we keep it we would know about those failure cases(when this should be omitted) instead of silently not showing peer ID.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -512,6 +512,9 @@ func (ipfs *Connector) ConnectSwarms(ctx context.Context) error {
 
 	for _, id := range ids {
 		ipfsID := id.IPFS
+		if ipfsID == nil {
```
hsanjuan: should `continue` if `id.Error != ""`, `id.IPFS == nil` (hopefully this should never happen), or `id.IPFS.Error != ""`

api/types.go
```diff
@@ -344,14 +344,14 @@ type ID struct {
 	Commit                string      `json:"commit" codec:"c,omitempty"`
 	RPCProtocolVersion    protocol.ID `json:"rpc_protocol_version" codec:"rv,omitempty"`
 	Error                 string      `json:"error" codec:"e,omitempty"`
-	IPFS                  IPFSID      `json:"ipfs" codec:"ip,omitempty"`
+	IPFS                  *IPFSID     `json:"ipfs" codec:"ip,omitempty"`
```
hsanjuan: add `omitempty` to the json tag.

PR #723, opened 2019-03-18
docs: fix download link in readme

olizilla:


hsanjuan:
I added an alias in the website too

PR #728, opened 2019-03-23
fix nil pointer panic

### LINKED ISSUE #727: nil pointer panic on add
#### Basic information

  * [x] Master
  * [ ] Latest stable version
  * [x] Bug
  * [ ] Feature request
  * [ ] Enhancement
  * [x] Linux
  * [ ] macOS
  * [ ] Windows
  * [ ] Other: `which?`
  * [ ] Built from sources
  * [x] Docker
  * [ ] Snap
  * [ ] Other: `which?`

#### Description

I encounter a nil pointer panic when adding a file or URL:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x98973e]

goroutine 9 [running]:
main.main.func6.1(0xc00025a834, 0xc0000acc60, 0xc000066fc0)
        /go/src/github.com/ipfs/ipfs-cluster/cmd/ipfs-cluster-ctl/main.go:436 +0x35e
created by main.main.func6
        /go/src/github.com/ipfs/ipfs-cluster/cmd/ipfs-cluster-ctl/main.go:417 +0x5b3
```
The panic occurs here: https://github.com/ipfs/ipfs-cluster/blob/master/cmd/ipfs-cluster-ctl/main.go#L436
Last changed by: https://github.com/ipfs/ipfs-cluster/pull/637

jmank88:
This PR prevents a nil pointer panic, and attempts to make the offending code more clear.

Fixes #727

### DIFF
--- a/cmd/ipfs-cluster-ctl/main.go
+++ b/cmd/ipfs-cluster-ctl/main.go
@@ -418,13 +418,13 @@ cluster "pin add".
 					defer wg.Done()
 
 					var buffered []*addedOutputQuiet
-					var lastBuf = make([]*addedOutputQuiet, 1, 1)
+					var lastBuf *addedOutputQuiet
 					var qq = c.Bool("quieter")
 					var q = c.Bool("quiet") || qq
 					var bufferResults = c.Bool("no-stream")
 					for v := range out {
 						added := &addedOutputQuiet{v, q}
-						lastBuf[0] = added
+						lastBuf = added
 						if bufferResults {
 							buffered = append(buffered, added)
 							continue
@@ -433,18 +433,18 @@ cluster "pin add".
 							formatResponse(c, added, nil)
 						}
 					}
-					if lastBuf[0].added == nil {
+					if lastBuf == nil || lastBuf.added == nil {
 						return // no elements at all
 					}
 					if bufferResults { // we buffered.
 						if qq { // [last elem]
-							formatResponse(c, lastBuf, nil)
+							formatResponse(c, []*addedOutputQuiet{lastBuf}, nil)
 							return
 						}
 						// [all elems]
 						formatResponse(c, buffered, nil)
 					} else if qq { // we already printed unless Quieter
-						formatResponse(c, lastBuf[0], nil)
+						formatResponse(c, lastBuf, nil)
 						return
 					}
 				}()


cmd/ipfs-cluster-ctl/main.go
```diff
@@ -433,18 +433,18 @@ cluster "pin add".
 							formatResponse(c, added, nil)
 						}
 					}
-					if lastBuf[0].added == nil {
+					if lastBuf == nil || lastBuf.added == nil {
```
jmank88: This entire change could be reduced to just `if lastBuf[0] == nil || lastBuf[0].added == nil {`, but I thought dropping the slice made things simpler and more readable.

jmank88:
Happy to, although I didn't actually specify `--no-stream`, so perhaps more tests are required to fully cover this path.

lanzafame:
Ah yeah I was going off the commit you mentioned in the issue.

hsanjuan:
I have trouble reproducing this. Can you provide the command that causes the issue?

jmank88:
I was testing a newly deployed cluster to see if ctl could talk to service. With ipfs and cluster running via docker-compose, both of these cause the panic (consistently):
```
> docker-compose exec cluster ipfs-cluster-ctl add test
> docker-compose exec cluster ipfs-cluster-ctl add http://test
```
The file `test` does not exist, and the URL is obviously bogus, so I'm guessing the panic is a secondary problem after some other error.

On my local machine, a non-dockerized local call will produce the panic _inconsistently_, and otherwise returns expected errors:
```
> ipfs-cluster-ctl add test
error : lstat test: no such file or directory
> ipfs-cluster-ctl add http://test
An error occurred:
  Code: 0
  Message: Post http://127.0.0.1:9094/add?chunker=size-262144&cid-version=0&hash=sha2-256&hidden=false&layout=&name=&progress=false&raw-leaves=false&recursive=false&replication-max=0&replication-min=0&shard=false&shard-size=104857600&stream-channels=true&wrap-with-directory=false: dial tcp 127.0.0.1:9094: connect: connection refused
```
So maybe there is a race hiding here as well.

hsanjuan:
Thanks!

PR #729, opened 2019-03-23
Fix: ctl: --enc=json shows empty output on "add"

hsanjuan:
The problem is that addedOutputQuiet produced `{}` when converted to JSON.


hsanjuan:
voice approval

PR #735, opened 2019-03-26
support --nocopy when adding URLs

jmank88:
https://github.com/ipfs/go-ipfs/issues/6065 added URL support to the `add` command, along with `--nocopy` as well, replacing the need for the experimental `urlstore`. This PR proposes supporting `ipfs-cluster-ctl add --nocopy <url>`.

### DIFF
--- a/cmd/ipfs-cluster-ctl/main.go
+++ b/cmd/ipfs-cluster-ctl/main.go
@@ -349,6 +349,10 @@ cluster "pin add".
 					Value: defaultAddParams.ReplicationFactorMax,
 					Usage: "Sets the maximum replication factor for pinning this file",
 				},
+				cli.BoolFlag{
+					Name:  "nocopy",
+					Usage: "Add the URL using filestore. Implies raw-leaves. (experimental)",
+				},
 				// TODO: Uncomment when sharding is supported.
 				// cli.BoolFlag{
 				//	Name:  "shard",
@@ -408,6 +412,10 @@ cluster "pin add".
 					p.CidVersion = 1
 				}
 				if p.CidVersion > 0 {
+					p.RawLeaves = true
+				}
+				p.NoCopy = c.Bool("nocopy")
+				if p.NoCopy {
 					p.RawLeaves = true
 				}
 
--- a/adder/adder.go
+++ b/adder/adder.go
@@ -117,6 +117,7 @@ func (a *Adder) FromFiles(ctx context.Context, f files.Directory) (cid.Cid, erro
 	ipfsAdder.Chunker = a.params.Chunker
 	ipfsAdder.Out = a.output
 	ipfsAdder.Progress = a.params.Progress
+	ipfsAdder.NoCopy = a.params.NoCopy
 
 	// Set up prefix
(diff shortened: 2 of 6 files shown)


cmd/ipfs-cluster-ctl/main.go
```diff
@@ -349,6 +349,10 @@ cluster "pin add".
 					Value: defaultAddParams.ReplicationFactorMax,
 					Usage: "Sets the maximum replication factor for pinning this file",
 				},
+				cli.BoolFlag{
+					Name:  "nocopy",
+					Usage: "Add the URL using filestore. Implies raw-leaves. (experimental)",
```
hsanjuan: Doesn't it add to the urlstore? I've had some problems figuring out how this is now working in ipfs (I haven' t been able to find the place where the URL is taken from the multipart request and used for the urlstore - I suspect it's somewhere in `unixfs`), but if urlstore is not enabled, adding urls like this on ipfs fails.

jmank88: `WebFile` now implements `FileInfo`, and returns the URL via `FileInfo.AbsPath()`, which is why the upgrade to `go-ipfs-files v0.0.2` is required. This is passed along via the `abspath` header, which ends up in the `ReaderFile`. It is then extracted and set as `DagBuilderParams.fullpath` https://github.com/ipfs/go-unixfs/blob/master/importer/helpers/dagbuilder.go#L85-L88, which in combination with the `nocopy` flag, triggers the same underlying mechanism as `urlstore`, but `urlstore` is not directly used. Only `filestore` is required https://github.com/ipfs/go-ipfs/blob/master/core/coreapi/unixfs.go#L52-L54. Here is the read logic: https://github.com/ipfs/go-ipfs/blob/master/filestore/fsrefstore.go#L138

jmank88: As far as I can tell, enabling `urlstore` makes the command available, but that command just leverages `filestore` internals to actually do the work.

jmank88:
> I wonder, when adding a file like this, is the DAG builder outputting a single special unixfs block (symlink)? This is my intuition but I haven't verified it.

Not sure I fully grok your terminology, but it creates a `FilestoreNode` whose `PosInfo.FullPath` contains the URL, then checks for URLs on reads.

cmd/ipfs-cluster-ctl/main.go
hsanjuan: Thanks, that's very useful. I have opened https://github.com/ipfs/go-ipfs/issues/6139

hsanjuan:
> > I wonder, when adding a file like this, is the DAG builder outputting a single special unixfs block (symlink)? This is my intuition but I haven't verified it.
> 
> Not sure I fully grok your terminology, but it creates a `FilestoreNode` whose `PosInfo.FullPath` contains the URL, then checks for URLs on reads.

Ah right, I understand now. Adding will chunk and build the resulting DAG blocks in the same way, but those blocks are filestore nodes, and instead of having the data payload they have a link to source+offset. Thanks!

Issue #448, opened 2018-06-01
There should be an api.Client which supports load balancing among multiple cluster peers


hsanjuan:
#### Basic information

* [x] Type (mark as appropiate):
  * [ ] Bug
  * [x] Feature request
  * [ ] Enhancement

#### Description

Following #447, a new client implementation would take multiple `*client.Configs` (or take one with multiple peer addresses) and maintain multiple `api/rest/client`s internally . When a request comes, it should round-robin, or randomly select a client. When a network error occurs, it should retry with a different peer.

hsanjuan:
Blocked on #447

hsanjuan:
No longer blocked as #447 is resolved.

lanzafame:
Blocked by #561.

kishansagathiya:
https://github.com/ipfs/ipfs-cluster/pull/561 is merged, removing `blocked` tag

kishansagathiya:
@hsanjuan Let me know if I understand this correctly.

You want a new client implementation whose job would be load balancing. Any request would first come to cluster leader's load balancing client which would pass it on(reverse proxy?) to rest api clients of cluster peers.

lanzafame:
We are trying to get rid of any kind of cluster leader, so it wouldn't make sense to give the current one any more responsibility. Not sure what the solution is in this case.

hsanjuan:
Hi, the leader has nothing to do here. We have a `client` interface:

https://github.com/ipfs/ipfs-cluster/blob/master/api/rest/client/client.go#L40

and a default implementation for it:

https://github.com/ipfs/ipfs-cluster/blob/master/api/rest/client/client.go#L174

The default implementation uses a single API endpoint (APIAddr).

This feature is about having a `client` implementation to which you can provide multiple cluster REST API endpoints and which will do load balancing between them. Internally, it will keep a default client pointing to each of the given APIAddrs. The logic of this is mostly around choosing which client to use.

DarkBlaez:
Fundamentally, load balancing should be done external to ipfs-cluster. There is absolutely no reason to put any level of lad balancing within the software. This can easily be handled external. Remember even Kubernetes will balance across POD instances if this is spun up in such. API calls to any node in the cluster could be fronted with a load balancer using any number of methods.

-DB

hsanjuan:
[...]

PR #737, opened 2019-04-02
Add load balancing client

kishansagathiya:
The load balancing client introduced in this commit would pick a client
randomly to execute the REST API request

#448


kishansagathiya:
Let me know if this looks good.

- Add an option `--lbclient` to `ipfs-cluster-ctl`
- Get a list of peer IDs using `consensus.Peers`
- creates configs of each client with multiaddr being `/p2p/<peer.ID>`
- Use the array of configs to create a load balancing client

@hsanjuan @lanzafame

api/rest/client/lbclient.go
```diff
@@ -13,27 +12,44 @@ import (
...
-		return *lc.clients[n], nil
+func (lc *loadBalancingClient) whichClient() Client {
+	switch lc.strategy {
+	case Random:
+		return *lc.clients[rand.Intn(len(lc.clients))]
+	case RoundRobin:
+		clientIndex++
+		if clientIndex == lc.length {
+			clientIndex = 0
+		}
```
kishansagathiya: `clientIndex` is racy if multiple client requests are made by the peer at the same time. Should I assume that it would happen?

api/rest/client/lbclient.go
```diff
@@ -42,112 +58,96 @@ func NewLBClient(cfgs []*Config) (Client, error) {
...
+	var id *api.ID
+	for {
+		id, err = lc.whichClient().ID(ctx)
+		if err != nil {
+			break
+		}
 	}
-
-	return dc.ID(ctx)
+	return id, err
```
kishansagathiya: I can run a loop like this in every method, but I am going to look around a bit and see if I can take advantage of similarities in all the functions.

api/rest/client/lbclient.go
```diff
@@ -42,112 +58,96 @@ func NewLBClient(cfgs []*Config) (Client, error) {
...
 // ID returns information about the cluster Peer.
 func (lc *loadBalancingClient) ID(ctx context.Context) (*api.ID, error) {
-	dc, err := lc.whichClient()
-	if err != nil {
-		return nil, err
+	var err error
+	var id *api.ID
+	for {
+		id, err = lc.whichClient().ID(ctx)
+		if err != nil {
```
kishansagathiya: Retry if client errors

api/rest/client/lbclient.go
hsanjuan: Yes

hsanjuan:
> Most functions have similar code. Wonder if something can be done to remove that redundancy, probably using reflect?

Good point.

Better do like:

```go
func retryCall(call func(client) error) {
   err := errors.New("first run")
   for err != nil {
        err = call(nextClient()) // need to handle that already tried with all clients and then do error
   }
}
```

and then each method would do:

```go
func ID(...) {
   call := func(client) error {
      client.ID(...)
   }
   retryCall(call)
}
```

No reflect needed.

api/rest/client/lbclient.go
```diff
@@ -13,27 +12,44 @@ import (
...
+	strategy Strategy
+	retries  int
+	length   int
 }
 
-func (lc *loadBalancingClient) whichClient() (Client, error) {
-	if lc.shedulingStrategy == "random" {
-		r := rand.New(rand.NewSource(time.Now().UnixNano()))
-		n := r.Intn(len(lc.clients))
+var clientIndex int
```
hsanjuan: can this be a local variable to the client?

api/rest/client/lbclient.go
```diff
@@ -13,27 +12,44 @@ import (
...
+
+// Strategy represents load balancing strategy
+type Strategy int
+
 // loadBalancingClient a client to interact with IPFS Cluster APIs.
 // It balances the load by distributing it among peers.
 type loadBalancingClient struct {
-	clients           []*Client
-	shedulingStrategy string
+	clients  []*Client
```
hsanjuan: No need to make a pointer to an interface, you can use `Client` directly (the underlying type is probably already a pointer).

kishansagathiya:
> Add an option --lbclient to ipfs-cluster-ctl

wdyt? 

I think this PR should be blocked until https://github.com/ipfs/ipfs-cluster/issues/772 is done

api/rest/client/lbclient.go
```diff
@@ -0,0 +1,383 @@
...
+}
+
+func (lc *loadBalancingClient) retryRoundRobin(i int, count int, call func(Client) error) error {
+	err := call(lc.clients[i%lc.length])
+	i++
+	count++
+	if count == lc.retries || err == nil {
+		return err
+	}
+	return lc.retryRoundRobin(i, count, call)
```
kishansagathiya: I can may be convert this logic to a simple for loop. That would be more readable.

hsanjuan: Deal first with concurrent calls using the same client which is not load balancing. Concurrent calls should use two different clients. If I'm reading well you are only load balancing retries for a single endpoint.

hsanjuan:
> Add an option --lbclient to ipfs-cluster-ctl
> 
> 
> wdyt?
> 
> I think this PR should be blocked until #772 is done

I'm not sure that makes sense. If the local peer is down, it cannot receive configurations from other peers. If it's not down, there is no reason to make a slow remote call instead of a local one. And the configurations from other peers say nothing about the actual IPs where the REST APIs will be reachable (that's called service discovery and it's done by things like Consul).

The idea here came so it can be used by pinbots so they don't depend on a single peer. The case for `ctl` supporting that is weaker and the approach in #772 doesn't work as far as I understand the description. Please continue working on this which is very scoped and without any reason to be blocked.

api/rest/client/lbclient.go
kishansagathiya: No, It uses different client(thus different endpoints)
For random strategy https://github.com/ipfs/ipfs-cluster/pull/737/files#diff-5b5f584f1d1c387d644f626937013360R62
For round robin strategy https://github.com/ipfs/ipfs-cluster/pull/737/files#diff-5b5f584f1d1c387d644f626937013360R71

I am not using a common counter anymore. Every client call (say c.ID(ctx)) maintains it's own separate counter. First value is selected at random even in roundrobin, so that we don't end up using the first client a lot.

kishansagathiya:
> I'm not sure that makes sense. If the local peer is down, it cannot receive configurations from other peers. If it's not down, there is no reason to make a slow remote call instead of a local one. And the configurations from other peers say nothing about the actual IPs where the REST APIs will be reachable (that's called service discovery and it's done by things like Consul).

Yes, I got confused because I have passed an array of configs to create a loadBalancingClient. May be I should change that (should take multiaddr instead of config)

hsanjuan:
> Yes, I got confused because I have passed an array of configs to create a loadBalancingClient. May be I should change that (should take multiaddr instead of config)

Why would you change that? If anything, create another additional New* method which initializes with default configs by just passing multiaddresses. But not a must.

api/rest/client/lbclient.go
hsanjuan: ok, we can go that way

api/rest/client/lbclient.go
```diff
@@ -0,0 +1,383 @@
...
+		return lc.retryRandom(0, call)
+	case RoundRobin:
+		return lc.retryRoundRobin(rand.Intn(len(lc.clients)), 0, call)
+	default:
+		return nil
+	}
+}
+
+func (lc *loadBalancingClient) retryRandom(count int, call func(Client) error) error {
+	err := call(lc.clients[rand.Intn(len(lc.clients))])
```
hsanjuan: The issue specified that retries should only happen in the face of a network error. The errors from clients are actually wrapping `api.Error` (or should be). A network error would be when:

* The error cannot be casted to `api.Error` (log this as this is abnormal), or
* The `error.Code == 0` (meaning no http request happened).

api/rest/client/lbclient.go
hsanjuan: It's not strictly load-balancing but it saves having to lock/unlock the counters which I like.

kishansagathiya:
> Note another strategy could be a Failover one, where we try the clients in order, always starting with the first one.

By first client you mean current machine? and start with first client for every request? otherwise this is similar to roundrobin.

kishansagathiya:
> Add an option --lbclient to ipfs-cluster-ctl

It is still not clear how we want ctl to decide which client to use, without this option or ctl.json

lanzafame:
@kishansagathiya @hsanjuan why is the load balancing functionality being put into a separate `Client` implementation, instead of creating a `net.Transport` implementation that supports load balancing? This way we wouldn't be wrapping the `Client` implementation with a `Next` call. I am sorry to bring this up so late @kishansagathiya

lanzafame:
Not an exact implementation of what we want but an example of something similar https://github.com/olivere/balancers/blob/master/transport.go

hsanjuan:
> @kishansagathiya @hsanjuan why is the load balancing functionality being put into a separate `Client` implementation, instead of creating a `net.Transport` implementation that supports load balancing? This way we wouldn't be wrapping the `Client` implementation with a `Next` call. I am sorry to bring this up so late @kishansagathiya

It's an idea but a custom transport is not flexible enough? If you want to load-balance between several cluster peers which are configured with different username/passwords, or some listening on libp2p and others on http ? Those things are outside the `net.Transport` scope...

lanzafame:
> If you want to load-balance between several cluster peers which are configured with different username/passwords, or some listening on libp2p and others on http ? 

@hsanjuan Firstly, are these necessary or requested use cases? I agree that those are use cases beyond what can be implemented in a `net.Transport` but with a focus on v1, it seems like not something we really need to support right now. 

Secondly, can we start with simple http load balancing via a `net.Transport`?

hsanjuan:
> > If you want to load-balance between several cluster peers which are configured with different username/passwords, or some listening on libp2p and others on http ?
> 
> @hsanjuan Firstly, are these necessary or requested use cases? I agree that those are use cases beyond what can be implemented in a `net.Transport` but with a focus on v1, it seems like not something we really need to support right now.
> 
> Secondly, can we start with simple http load balancing via a `net.Transport`?

I don't understand how load balancing via `net.Transport` is any simpler. Involves changing the client's configuration, and adds complexity to the base client (which is already tricky because of the 3 transports already supported). Supporting multiple strategies will be as complex. And it's way less flexible. It would not even allow to distinguish between cluster errors and network errors without artificially sniffing an http response. The thing is, you cannot do HTTP load balancing in `net.Transport` because `net.Transport` is for TCP (etc) load balancing at most. You're one layer too deep.

This is the simple way.

hsanjuan:
> I don't understand how load balancing via `net.Transport` is any simpler. Involves changing the client's configuration, and adds complexity to the base client (which is already tricky because of the 3 transports already supported). Supporting multiple strategies will be as complex. And it's way less flexible. ~It would not even allow to distinguish between cluster errors and network errors without artificially sniffing an http response. The thing is, you cannot do HTTP load balancing in `net.Transport` because `net.Transport` is for TCP (etc) load balancing at most. You're one layer too deep.~
> 
> This is the simple way.

I updated my comment because the last part is incorrect. You don't need to distinguish between cluster errors and network errors on a network transport because you only see network errors. So that would work. Still find it simpler to balance between clients.

api/rest/client/lbclient.go
```diff
@@ -0,0 +1,431 @@
...
+
+// loadBalancingClient a client to interact with IPFS Cluster APIs.
+// It balances the load by distributing it among peers.
+type loadBalancingClient struct {
+	strategy LBStrategy
+}
+
+// LBStrategy is a strategy to load balance request among clients
+type LBStrategy interface {
+	Next(count int, call func(Client) error) error
```
hsanjuan: This does not correspond to what I suggested: `Next() Client`

The strategy is not in charge of calling the method, it's just in charge of providing clients in certain order so that the load-balancing client can do that. This way you have ended up with duplicated code blocks in the strategies (creating the clients, handling the errors..).

I guess the `LBStrategy` will also need to implement a `SetClients([]Client)` method as a way to pass in the list of clients as created by the load balancing client.

api/rest/client/lbclient.go
```diff
@@ -0,0 +1,431 @@
...
+		if err != nil {
+			return nil, err
+		}
+		clients = append(clients, defaultClient)
+	}
+
+	rand.Seed(time.Now().UnixNano())
+
+	counter := make(chan int)
+	counter <- 0
```
hsanjuan: I think this is hanging forever and this function will never return. `make(chan int, 1)` might help.

The problem with using channels is that you will leak them as you have no way of closing them when the client is not used anymore and they will be happily stay open and holding data. Consider using the stuff `atomic` for this, which is also possibly way faster.

api/rest/client/lbclient.go
kishansagathiya: > This way you have ended up with duplicated code blocks in the strategies (creating the clients, handling the errors..).

Changing this to `Next() Client` would also result in similar amount of duplicate code, since to retry the request a `r.retries` you will need to add a loop in every method.

The idea was to have that similar code at a common place.

> The strategy is not in charge of calling the method

If not for this, we would be able to add a `Retry() err` method to the interface, make `Next() Client` and avoid duplicate code from all methods.

hsanjuan: In the previous version of this code, the load balancer client had a method that did the retry logic for all methods in a single place. That's a problem you had already solved.

kishansagathiya: > I guess the LBStrategy will also need to implement a SetClients([]Client) method as a way to pass in the list of clients as created by the load balancing client.

The clients are added when a strategy is created, I don't see a need for SetClients()

hsanjuan: You have each strategy creating the clients this duplication client creation code. I don't want users writing alternative strategies to have to create and initialize clients when it can be done for them.

api/rest/client/lbclient.go
kishansagathiya: https://github.com/ipfs/ipfs-cluster/pull/737/commits/c24f0004fe9ccdf7d365d8709d498c2fdad5a635

Uses atomic now

api/rest/client/lbclient.go
kishansagathiya: https://github.com/ipfs/ipfs-cluster/pull/737/commits/c24f0004fe9ccdf7d365d8709d498c2fdad5a635

api/rest/client/lbclient.go
```diff
@@ -12,74 +13,87 @@ import (
...
-	err := call(r.clients[i])
-	apiErr, ok := err.(*api.Error)
-	if !ok {
-		logger.Error("could not cast error into api.Error")
-		return err
-	}
+// Next return the next client to be used.
+func (r *RoundRobin) Next(count int) Client {
+	i := atomic.LoadInt32(r.counter)
+	atomic.StoreInt32(r.counter, (i+1)%int32(r.length))
```
hsanjuan: This will return the same client to two concurrent callers because the steps between load and atomic are not "atomic". You can use atomic add and do the module at the end. I don't think overflow is a problem (but check), as long as you use uints..

kishansagathiya: > I don't think overflow is a problem (but check), as long as you use uints

true, starts from 0 after max https://play.golang.org/p/XKh3hYS6zPO

api/rest/client/lbclient.go
```diff
@@ -0,0 +1,419 @@
...
+type RoundRobin struct {
+	clients []Client
+	counter uint32
+}
+
+// Next return the next client to be used.
+func (r *RoundRobin) Next(count int) Client {
+	i := atomic.AddUint32(&r.counter, 1)
+
+	return r.clients[i]
```
hsanjuan: Should this be `i%len(clients)` ?

api/rest/client/lbclient.go
```diff
@@ -0,0 +1,419 @@
...
+	err := call(lc.strategy.Next(count))
+	count++
+
+	// successful request
+	if err == nil {
+		return nil
+	}
+
+	apiErr, ok := err.(*api.Error)
+	if !ok {
```
hsanjuan: I think this is wrong. Auto-retries should happen when the error is NOT an `*api.Error` (you don't want to retry on 404s.

api/rest/client/lbclient.go
kishansagathiya: yes

api/rest/client/lbclient.go
```diff
@@ -0,0 +1,419 @@
...
+	counter uint32
+}
+
+// Next returns the next client to be used.
+func (f *Failover) Next(count int) Client {
+	if count == 0 {
+		return f.clients[0]
+	}
+
+	i := atomic.AddUint32(&f.counter, 1)
```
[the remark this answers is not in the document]
kishansagathiya: You get i from f.counter

You want to skip if i is 0, because if we are here, we already tried 0th client.

api/rest/client/lbclient.go
kishansagathiya: Looking at client.do,
All errors returned by client methods are of type `api.Error`, even if error is not coming from the api as response (i.e, something to do with the client or how we are making requests). We check the code to differentiate. If code is 0, then it is from the client and not from API.
https://github.com/ipfs/ipfs-cluster/blob/53f1a99feb8f070b87fec3eb0c8eda13f555f211/api/rest/client/request.go#L28

It is a safety check. This error should never occur. It must always cast into api.Error.

kishansagathiya:
This has few more errors, will push changes along with tests

api/rest/client/lbclient.go
hsanjuan: ok but then:

```
if apiErr.Code != 0 {
		return err
}
```

should come earlier because you don't want to check retry count if you're not planning to retry?

Maybe put a comment to indicate the safeguard too.

Also note that you don't need to check for `ok`, if it does not cast, `apiErr.Code` will stay as 0 and you can retry, so everything can happen with a single check.

api/rest/client/lbclient.go
```diff
@@ -0,0 +1,421 @@
...
+// If the local call fail it would try other client in a round robin fashion.
+type Failover struct {
+	clients []Client
+	counter uint32
+	length  uint32
+}
+
+// Next returns the next client to be used.
+func (f *Failover) Next(count int) Client {
+	if count == 0 {
```
hsanjuan: I think this function can be summarized to `return f.clients[count]`. You are sort of load-balancing all clients except the first. Failover just means `try the next in the list`.

kishansagathiya: > Failover just means try the next in the list

ok, I can change this. What I have for Failover, I think, should be better called RoundRobin.

kishansagathiya: I will make RoundRobin as well, always use the first client.

hsanjuan: The other strategy is round robin because it's a global roundrobin. This is not a global roundrobin in the sense that clients are always used in order based on the current count of retries.

Don't touch the RoundRobin client, it does not have to always use the first client (why would it).

kishansagathiya: > (why would it)

RoundRobin currently, for the first request try, makes a slower remote call instead of local one.

hsanjuan: who said the first client is a local one?

PR #738, opened 2019-04-04
Pin timeouts should start from the last block

### LINKED ISSUE #497: Pin timeouts should happen from the moment the last block was received
#### Basic information

  * [x] Master
  * [ ] Latest stable version
  * [ ] Bug
  * [ ] Feature request
  * [x] Enhancement

#### Description

ipfsconnector Pin operation should not have a timeout since the start of the operation, but a timeout since the last block was received, thus only cancelling operations when there are no more blocks provided in the network.

kishansagathiya:
IPFSConnector Pin operation should not have a timeout since the start
of the operation, but a timeout since the last block was received,
thus only cancelling operations when there are no more blocks provided
in the network

Fixes #497

### DIFF
--- a/ipfsconn/ipfshttp/ipfshttp.go
+++ b/ipfsconn/ipfshttp/ipfshttp.go
@@ -423,7 +423,7 @@ func (ipfs *Connector) refsProgress(ctx context.Context, hash cid.Cid, maxDepth 
 
 		// We have a Ref!
 		if errStr := ref.Err; errStr != "" {
-			return errors.New(errStr)
+			logger.Error(errStr)
 		}
 
 		select { // do not lock
--- a/ipfsconn/ipfshttp/ipfshttp.go
+++ b/ipfsconn/ipfshttp/ipfshttp.go
@@ -3,7 +3,6 @@
 package ipfshttp
 
 import (
-	"bufio"
 	"context"
 	"encoding/json"
 	"errors"
@@ -305,7 +304,7 @@ func (ipfs *Connector) Pin(ctx context.Context, hash cid.Cid, maxDepth int) erro
 	timer := time.NewTimer(ipfs.config.PinTimeout)
 	defer timer.Stop()
 
-	reset := make(chan int)
+	reset := make(chan struct{})
 	go checkTimeout(ctx, cancel, timer, ipfs.config.PinTimeout, reset)
 
 	switch ipfs.config.PinMethod {
@@ -317,7 +316,7 @@ func (ipfs *Connector) Pin(ctx context.Context, hash cid.Cid, maxDepth int) erro
 		}
 		defer res.Body.Close()
 
-		err = handleRefsProgress(json.NewDecoder(bufio.NewReader(res.Body)), reset)
+		err = handleRefsProgress(json.NewDecoder(res.Body), reset)
 		if err != nil {
 			return err
 		}
@@ -332,7 +331,7 @@ func (ipfs *Connector) Pin(ctx context.Context, hash cid.Cid, maxDepth int) erro
 	}
 	defer res.Body.Close()
 
-	err = handlePinsProgress(json.NewDecoder(bufio.NewReader(res.Body)), reset)
+	err = handlePinsProgress(json.NewDecoder(res.Body), reset)
 	if err == nil {
 		logger.Info("IPFS Pin request succeeded: ", hash)
 	}
@@ -484,7 +483,7 @@ func checkResponse(path string, code int, body []byte) error {
 	return fmt.Errorf("IPFS-post '%s' unsuccessful: %d: %s", path, code, body)
 }
 
-func handleRefsProgress(dec *json.Decoder, reset chan int) error {
+func handleRefsProgress(dec *json.Decoder, reset chan struct{}) error {
 	ticker := time.NewTicker(5 * time.Second)
 	var val atomic.Value
 	var done bool
@@ -494,7 +493,7 @@ func handleRefsProgress(dec *json.Decoder, reset chan int) error {
 			case <-ticker.C:
 				{
 					if val.Load().(bool) {
-						reset <- 1
+						reset <- struct{}{}
 					}
 					val.Store(false)
 				}
@@ -528,7 +527,7 @@ func handleRefsProgress(dec *json.Decoder, reset chan int) error {
 	return nil
 }
 
-func handlePinsProgress(dec *json.Decoder, reset chan int) error {
+func handlePinsProgress(dec *json.Decoder, reset chan struct{}) error {
 	ticker := time.NewTicker(5 * time.Second)
 	var val atomic.Value
 	var progress int
@@ -539,7 +538,7 @@ func handlePinsProgress(dec *json.Decoder, reset chan int) error {
 			case <-ticker.C:
 				{
 					if val.Load().(bool) {
-						reset <- 1
+						reset <- struct{}{}
 					}
 					val.Store(false)
 				}
@@ -570,7 +569,7 @@ func handlePinsProgress(dec *json.Decoder, reset chan int) error {
 	return nil
 }
 
-func checkTimeout(ctx context.Context, cancel context.CancelFunc, timer *time.Timer, timeout time.Duration, reset chan int) {
+func checkTimeout(ctx context.Context, cancel context.CancelFunc, timer *time.Timer, timeout time.Duration, reset chan struct{}) {
 	var done bool
 	for {
 		select {
--- a/ipfsconn/ipfshttp/ipfshttp.go
+++ b/ipfsconn/ipfshttp/ipfshttp.go
@@ -431,7 +431,6 @@ func (ipfs *Connector) refsProgress(ctx context.Context, hash cid.Cid, maxDepth 
 		default:
 		}
 	}
-	return nil
 }
 
 // pinProgress pins an item and sends fetched node's progress on a
@@ -478,7 +477,6 @@ func (ipfs *Connector) pinProgress(ctx context.Context, hash cid.Cid, maxDepth i
 		default:
 		}
 	}
-	return nil
 }
 
 // Unpin performs an unpin request against the configured IPFS
--- a/ipfsconn/ipfshttp/ipfshttp.go
+++ b/ipfsconn/ipfshttp/ipfshttp.go
@@ -91,13 +91,13 @@ type ipfsResolveResp struct {
 }
 
 type ipfsRefsResp struct {
-	Ref string `json:"Ref"`
-	Err string `json:"Err"`
+	Ref string
+	Err string
 }
 
 type ipfsPinsResp struct {
-	Pins     []string `json:"Pins"`
-	Progress int      `json:"Progress"`
+	Pins     []string
+	Progress int
 }
 
 type ipfsSwarmPeersResp struct {
--- a/ipfsconn/ipfshttp/ipfshttp.go
+++ b/ipfsconn/ipfshttp/ipfshttp.go
@@ -304,8 +304,8 @@ func (ipfs *Connector) Pin(ctx context.Context, hash cid.Cid, maxDepth int) erro
 	timer := time.NewTimer(ipfs.config.PinTimeout)
 	defer timer.Stop()
 
-	elapsed := time.Now()
-	checkTimeout(ctx, cancel, timer, &elapsed, ipfs.config.PinTimeout)
+	reset := make(chan int)
+	checkTimeout(ctx, cancel, timer, ipfs.config.PinTimeout, reset)
 
 	switch ipfs.config.PinMethod {
 	case "refs": // do refs -r first
@@ -316,7 +316,7 @@ func (ipfs *Connector) Pin(ctx context.Context, hash cid.Cid, maxDepth int) erro
 		}
 		defer res.Body.Close()
 
-		err = handleRefsProgress(json.NewDecoder(bufio.NewReader(res.Body)), &elapsed)
+		err = handleRefsProgress(json.NewDecoder(bufio.NewReader(res.Body)), reset)
 		if err != nil {
 			return err
 		}
@@ -331,7 +331,7 @@ func (ipfs *Connector) Pin(ctx context.Context, hash cid.Cid, maxDepth int) erro
 	}
 	defer res.Body.Close()
 
-	err = handlePinsProgress(json.NewDecoder(bufio.NewReader(res.Body)), &elapsed)
+	err = handlePinsProgress(json.NewDecoder(bufio.NewReader(res.Body)), reset)
 	if err == nil {
 		logger.Info("IPFS Pin request succeeded: ", hash)
 	}
@@ -483,7 +483,7 @@ func checkResponse(path string, code int, body []byte) error {
 	return fmt.Errorf("IPFS-post '%s' unsuccessful: %d: %s", path, code, body)
 }
 
-func handleRefsProgress(dec *json.Decoder, elapsed *time.Time) error {
+func handleRefsProgress(dec *json.Decoder, reset chan int) error {
 	var last string
 	for {
 		var ref ipfsRefsResp
@@ -492,20 +492,21 @@ func handleRefsProgress(dec *json.Decoder, elapsed *time.Time) error {
 		} else if err != nil {
 			return err
 		}
+		// logger.Infof("Ref %s", ref.Ref)
 		if ref.Err != "" {
 			logger.Error(ref.Err)
 		}
 
 		if ref.Ref != last {
 			last = ref.Ref
-			*elapsed = time.Now()
+			reset <- 1
 		}
 	}
 
 	return nil
 }
 
-func handlePinsProgress(dec *json.Decoder, elapsed *time.Time) error {
+func handlePinsProgress(dec *json.Decoder, reset chan int) error {
 	var progress int
 	for {
 		var pins ipfsPinsResp
@@ -514,28 +515,31 @@ func handlePinsProgress(dec *json.Decoder, elapsed *time.Time) error {
 		} else if err != nil {
 			return err
 		}
+		logger.Infof("Progress: %d", pins.Progress)
 		if pins.Progress > progress {
 			progress = pins.Progress
-			*elapsed = time.Now()
+			reset <- 1
 		}
 	}
 
 	return nil
 }
 
-func checkTimeout(ctx context.Context, cancel context.CancelFunc, timer *time.Timer, elapsed *time.Time, timeout time.Duration) {
+func checkTimeout(ctx context.Context, cancel context.CancelFunc, timer *time.Timer, timeout time.Duration, reset chan int) {
 	var done bool
 	go func() {
 		for {
 			select {
+			case <-reset:
+				{
+					//timer.Stop()
+					timer.Reset(timeout)
+				}
 			case <-timer.C:
 				{
-					if time.Since(*elapsed) >= timeout {
-						done = true
-						cancel()
-						break
-					}
-					timer.Reset(timeout - time.Since(*elapsed))
+					done = true
+					cancel()
+					break
 				}
 			case <-ctx.Done():
 				{
--- a/ipfsconn/ipfshttp/ipfshttp.go
+++ b/ipfsconn/ipfshttp/ipfshttp.go
@@ -14,6 +14,7 @@ import (
 	"net/url"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	gopath "github.com/ipfs/go-path"
@@ -305,7 +306,7 @@ func (ipfs *Connector) Pin(ctx context.Context, hash cid.Cid, maxDepth int) erro
 	defer timer.Stop()
 
 	reset := make(chan int)
-	checkTimeout(ctx, cancel, timer, ipfs.config.PinTimeout, reset)
+	go checkTimeout(ctx, cancel, timer, ipfs.config.PinTimeout, reset)
 
 	switch ipfs.config.PinMethod {
 	case "refs": // do refs -r first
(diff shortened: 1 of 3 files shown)


kishansagathiya:
Was it really this simple or did I miss something here?

lanzafame:
I have had a quick look, and I think you have missed something or I have misunderstood what @hsanjuan meant in the issue. My interpretation is that the timeout should be occurring in the IPFSConnector but I didn't think that IPFS exposed enough insight into how the blocks are coming across the network for this.

hsanjuan:
The `--progress` flag when pinning should send updates for every pinned block (?? verify this claim). If blocks are still coming through, the pin timeout should be reset.

This change is just letting PinLs potentially hang forever which does not make much sense.

kishansagathiya:
When you say `--progress`, you mean `ipfs pin add --progress`, right?

It didn't show any updates on pinned blocked

```sh
$ ipfs ls QmWVc1hqrbmYGaYqA9sudZgA7YLyc1DRMPeTvZo3eSES4D
QmdqJSYx4Lc4Zqpv8omY4uDPenS58JBhsBHfVzJk5AqRNJ 45623854 
QmUgk6i1vtMx51xkKJxiMxPV7z8ihenRhB7Kt8skoFAZqX 45623854 
QmeuerZd3mME5Te1ooreCzjeVkWSpN4pqsWM5HizkgVHg8 36689280 
$ ipfs pin add --progress QmWVc1hqrbmYGaYqA9sudZgA7YLyc1DRMPeTvZo3eSES4D
pinned QmWVc1hqrbmYGaYqA9sudZgA7YLyc1DRMPeTvZo3eSES4D recursively
$ ipfs -D pin add --progress QmWVc1hqrbmYGaYqA9sudZgA7YLyc1DRMPeTvZo3eSES4D
13:48:15.276 DEBUG   cmd/ipfs: config path is /home/kishan/.ipfs main.go:104
13:48:15.291 DEBUG blockservi: BlockService GetBlock: 'QmRD2GqwpVUvZoxfRNNwdEgyh182pEFVNHq8aNEr78cmCt' blockservice.go:202
13:48:15.292 DEBUG blockservi: BlockService GetBlock: 'QmWsu5HeZPPMqSboGe4LVVmAX9nRjyWXcoD6DTkgHTyS9r' blockservice.go:202
13:48:15.293 DEBUG blockservi: BlockService GetBlock: 'QmSKboVigcD3AY4kLsob117KJcMHvMUu6vNFqk1PQzYUpp' blockservice.go:202
13:48:15.294 DEBUG blockservi: BlockService GetBlock: 'QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn' blockservice.go:202
13:48:15.294 DEBUG blockservi: BlockService GetBlock: 'QmWVc1hqrbmYGaYqA9sudZgA7YLyc1DRMPeTvZo3eSES4D' blockservice.go:202
pinned QmWVc1hqrbmYGaYqA9sudZgA7YLyc1DRMPeTvZo3eSES4D recursively
13:48:15.296  INFO    command: Shutting down node... request.go:108
13:48:15.296 DEBUG       core: core is shutting down... core.go:640
13:48:15.296 DEBUG blockservi: blockservice is shutting down... blockservice.go:324
```

kishansagathiya:
Same with the API as well
```
$ curl "http://localhost:5001/api/v0/pin/add?arg=QmWVc1hqrbmYGaYqA9sudZgA7YLyc1DRMPeTvZo3eSES4D&recursive=true&progress=true"
{"Pins":["QmWVc1hqrbmYGaYqA9sudZgA7YLyc1DRMPeTvZo3eSES4D"]}
```

hsanjuan:
It does show updates (with large enough pins at least).

```sh
$ ipfs --enc=json pin add --progress /ipns/dist.ipfs.io 
Fetched/Processed 1002 nodes
```

shows a growing count of blocks as they get fetched.

```
curl "localhost:5001/api/v0/pin/add?arg=/ipns/dist.ipfs.io&progress=true"
{"Pins":null,"Progress":527}
{"Pins":null,"Progress":652}
{"Pins":null,"Progress":914}
{"Pins":null,"Progress":1216}
{"Pins":null,"Progress":1438}
{"Pins":null,"Progress":1531}
{"Pins":null,"Progress":1624
...
```

shows json updates as expected. The timeout should count from the last of those updates.

Even clearer with `refs` since we do `refs` pinning by default. The timeout should count from the last ref received.

kishansagathiya:
It seems like updates are not per block, but on some interval. I am saying so because I see results like this
```
{"Pins":null,"Progress":13}
{"Pins":null,"Progress":13}
{"Pins":null,"Progress":13}
{"Pins":null,"Progress":14}
{"Pins":null,"Progress":14}
{"Pins":null,"Progress":14}
```

hsanjuan:
> It seems like updates are not per block, but on some interval. I am saying so because I see results like this
> 
> ```
> {"Pins":null,"Progress":13}
> {"Pins":null,"Progress":13}
> {"Pins":null,"Progress":13}
> {"Pins":null,"Progress":14}
> {"Pins":null,"Progress":14}
> {"Pins":null,"Progress":14}
> ```

yes, so take the `progress` number and only reset the timeout if it has grown...

kishansagathiya:
> This change is just letting PinLs potentially hang forever which does not make much sense.

Ok, so we need timeout for PinLs.

But, Pin timeouts should happen from the moment the last block was received. I am confused. I you asking to use a different context, or reset timeout(Is that possible?)

hsanjuan:
* PinLs should timeout as a normal `IPFSRequestTimeout`
* `refs -r` and `pin add` should timeout after `PinTimeout` when no progress is observed:
  * for `refs -r` that means, no new refs seen on the response stream
  * for `pin add` that means no progress seen on the progress updates

I can't tell the exact way to implement this because I'm not implementing it myself. I guess `time.Timer`s or `time.Sleep` or `time.Tickers` can be used to reset contexts and cancel requests upon expiration, but the idea of what needs to happen should be clear enough now.

kishansagathiya:
@hsanjuan 
> not parsing the json objects

So far we were ignoring the body. We just need to make sure that we are seeing new output (i.e., progress is happening). We can parse it, but it won't give us any useful information.
Also, the last entry in `/pin/add`, would be of different format (we can still parse it of course)

hsanjuan:
> So far we were ignoring the body. We just need to make sure that we are seeing new output (i.e., progress is happening). We can parse it, but it won't give us any useful information.

what if your output is error messages. Please parse the objects which are telling you the number of nodes fetched.

hsanjuan:
The whole part of `refs` pinning is missing by the way.

kishansagathiya:
@hsanjuan 
> what if your output is error messages. Please parse the objects which are telling you the number of nodes fetched.

Yes, Will add parsing.

> The whole part of refs pinning is missing by the way.

I am not sure what do you mean here (I have used the same post function for both `refs -r` and `pin add`)

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -450,6 +459,98 @@ func checkResponse(path string, code int, body []byte) error {
...
+			}
+			return err
+		}
+
+		err = json.Unmarshal(line, &ref)
+		if err != nil {
+			return err
+		}
+		if ref.Err != "" {
+			logger.Error(ref.Err)
```
kishansagathiya: Not returning since having an error in ref does not mean end of stream

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -450,6 +459,98 @@ func checkResponse(path string, code int, body []byte) error {
...
+
+	elapsed := time.Now()
+
+	go func() {
+		for {
+			select {
+			case <-timer.C:
+				{
+					if time.Since(elapsed) >= timeout {
+						cancel()
```
kishansagathiya: I could have explicitly returned once context is done, but I have relied on reader to end by throwing `context cancelled` error

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -89,6 +90,16 @@ type ipfsResolveResp struct {
 	Path string
 }
 
+type ipfsRefsResp struct {
+	Ref string `json:"Ref"`
```
hsanjuan: annotations here and below are redundant

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -265,8 +276,6 @@ func (ipfs *Connector) Pin(ctx context.Context, hash cid.Cid, maxDepth int) erro
 	ctx, span := trace.StartSpan(ctx, "ipfsconn/ipfshttp/Pin")
 	defer span.End()
 
-	ctx, cancel := context.WithTimeout(ctx, ipfs.config.PinTimeout)
-	defer cancel()
 	pinStatus, err := ipfs.PinLsCid(ctx, hash)
```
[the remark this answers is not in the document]
hsanjuan: Can you fix this in `Unpin` too?

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -450,6 +459,97 @@ func checkResponse(path string, code int, body []byte) error {
 	return fmt.Errorf("IPFS-post '%s' unsuccessful: %d: %s", path, code, body)
 }
 
+// postCtxStreaming makes a POST request against
+// the ipfs daemon, reads the body of the response line by line
+// at evey millisecond. It cancels the context if timeout duration
+// has passed since last new data on the buffer.
+func (ipfs *Connector) postCtxStreaming(ctx context.Context, path string, timeout time.Duration) error {
```
hsanjuan: this needs refactoring. a `doPostCtx` can happen in `Pin` and then pass the response reader to either a `handleRefsProgress` or `handlePinProgress` (or better name, but two separate functions).

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -450,6 +459,97 @@ func checkResponse(path string, code int, body []byte) error {
...
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+	reader := bufio.NewReader(res.Body)
+
+	if strings.HasPrefix(path, "pin/add") {
+		var pins ipfsPinsResp
+		var progress int
+		for {
```
hsanjuan: make a json `Decoder ` with the response reader and run `for { ... dec.Decode(&respObject) ... }` (or with `for dec.More() { ... }`) there is an example in the json module docs.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -450,6 +459,97 @@ func checkResponse(path string, code int, body []byte) error {
...
+				return err
+			}
+
+			err = json.Unmarshal(line, &pins)
+			if err != nil {
+				return err
+			}
+			if pins.Progress > progress {
+				progress = pins.Progress
+				elapsed = time.Now()
```
hsanjuan: this is racy, chances that it will panic

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -450,6 +459,97 @@ func checkResponse(path string, code int, body []byte) error {
...
+	elapsed := time.Now()
+
+	go func() {
+		for {
+			select {
+			case <-timer.C:
+				{
+					if time.Since(elapsed) >= timeout {
+						cancel()
+						break
```
hsanjuan: this is just breaking from the select, not from the for loop?

kishansagathiya: Yes, but once context is canceled our `reader` would throw an error `context cancelled` and the function would return.
I could break out of the loop by keeping some done flag, but that would add a bunch of lines of redundant code.

kishansagathiya: Unresolve if you think I should explicitly break out of this for loop and other infinite loops

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: This would create quite some duplicate code, but I guess we are trading that with readability, will do.

ipfsconn/ipfshttp/ipfshttp.go
hsanjuan: Your reader will return and the other function will end but this goroutine stays there forever (?)

hsanjuan: Which makes me realize that this goroutine is also not cleaned up when things don't time out.

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: https://github.com/ipfs/ipfs-cluster/pull/738/commits/58c3a98bd444ef85774296b2504863186b726600

kishansagathiya: Not fixed, time.Reset could also lead to racy behavior. Will add changes

kishansagathiya: This should take care of racy behaviour on using timer.Reset https://github.com/ipfs/ipfs-cluster/pull/738/commits/914b270c993bd373618fd39c5d6992e971ee6c16

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -450,6 +483,79 @@ func checkResponse(path string, code int, body []byte) error {
...
+
+func checkTimeout(ctx context.Context, cancel context.CancelFunc, timer *time.Timer, timeout time.Duration, reset chan int) {
+	var done bool
+	go func() {
+		for {
+			select {
+			case <-reset:
+				{
+					if timer.Stop() {
+						timer.Reset(timeout)
```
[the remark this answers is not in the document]
hsanjuan: We could end up being ressetting a timer every few milliseconds in default the case of `refs`, or every second or so with `pin add`. Would it be better to have a ticker that drains the reset channel once a second (or even once every 5) and then resets the timer? The timeouts here are at least minutes long. The sending part must not block on a full channel.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -289,22 +298,44 @@ func (ipfs *Connector) Pin(ctx context.Context, hash cid.Cid, maxDepth int) erro
 		pinArgs = fmt.Sprintf("recursive=true&max-depth=%d", maxDepth)
 	}
 
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	timer := time.NewTimer(ipfs.config.PinTimeout)
+	defer timer.Stop()
+
+	reset := make(chan int)
```
hsanjuan: channel never closed. Also, `make(chan struct{})`

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -450,6 +483,79 @@ func checkResponse(path string, code int, body []byte) error {
...
+		if pins.Progress > progress {
+			progress = pins.Progress
+			reset <- 1
+		}
+	}
+
+	return nil
+}
+
+func checkTimeout(ctx context.Context, cancel context.CancelFunc, timer *time.Timer, timeout time.Duration, reset chan int) {
```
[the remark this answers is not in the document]
kishansagathiya: All parameters are used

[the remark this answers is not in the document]
hsanjuan: ah ok, but the timer could just be created inside passing only the timeout ?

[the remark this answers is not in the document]
hsanjuan: you can probably avoid passing the context by closing the reset channel and checking if it has been closed with `case _, ok := reset:`

ipfsconn/ipfshttp/ipfshttp.go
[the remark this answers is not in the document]
kishansagathiya: > Would it be better to have a ticker that drains the reset channel once a second (or even once every 5)

This would add blocking time of one or five seconds because of the channels. I have some ideas, let me try.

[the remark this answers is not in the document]
hsanjuan: You should send like:

```
select {
case reset <- struct{}{}:
default:
}
```

so it doesn't block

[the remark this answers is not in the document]
kishansagathiya: I have tried https://github.com/ipfs/ipfs-cluster/pull/738/commits/d640fa8157d775bc3d14d541777e0fd4cc60f516
This reduces the number of times that we reset, but it still looks really horrible. There should be a better way.

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: > Also, make(chan struct{})

why?

ipfsconn/ipfshttp/ipfshttp.go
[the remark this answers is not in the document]
kishansagathiya: Using context let's me take care of the situation when we don't timeout and context is Done, because of something. May be this can be done without passing context, but why not pass context?

kishansagathiya:
@hsanjuan I am gonna move to something else and come back here after a day or two with fresh thoughts

ipfsconn/ipfshttp/ipfshttp.go
hsanjuan: https://dave.cheney.net/2014/03/25/the-empty-struct

hsanjuan:
@kishansagathiya I think I found a clean approach. Can you work on some tests for this? We need to modify the ipfs mock. 

`/refs` on `test.SlowCid1` should cause it to send 1 ref every two seconds (even if the same) and you can set the configuration timeout to 1 second and should see an error when testing pinning on "refs" mode.

"/pin/add" for `slowCid1` should send  something like 10 progress updates with no progress, once a second. You can set the configuration timeout to be 5 seconds (so that not affected by the slow "refs"), and then should get an error when pinning because there is no progress.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -288,57 +306,179 @@ func (ipfs *Connector) Pin(ctx context.Context, hash cid.Cid, maxDepth int) erro
...
+			}
 		}
-		defer res.Body.Close()
 
-		err = handleRefsProgress(json.NewDecoder(res.Body), reset)
-		if err != nil {
-			return err
+		// We have a Ref!
+		if errStr := ref.Err; errStr != "" {
+			return errors.New(errStr)
```
kishansagathiya: This doesn't mean end of stream. I am inclined to log this and continue reading the stream.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -288,57 +306,179 @@ func (ipfs *Connector) Pin(ctx context.Context, hash cid.Cid, maxDepth int) erro
...
+	res, err := ipfs.doPostCtx(ctx, ipfs.client, ipfs.apiURL(), path, "", nil)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+
+	_, err = checkResponse(path, res)
+	if err != nil {
+		return err
+	}
```
kishansagathiya: `checkResponse` should be removed

kishansagathiya: https://github.com/ipfs/ipfs-cluster/pull/738/commits/c3ca53080c03ca7846cb6d50de9a040cd7bdc4d5

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: https://github.com/ipfs/ipfs-cluster/pull/738/commits/c3ca53080c03ca7846cb6d50de9a040cd7bdc4d5

hsanjuan: ok

ipfsconn/ipfshttp/ipfshttp.go
hsanjuan: Why would you do that? Please revert

hsanjuan:
There was a force-push here. Mentioned https://github.com/ipfs/ipfs-cluster/commit/c3ca53080c03ca7846cb6d50de9a040cd7bdc4d5 does not seem to exist anymore. I'm not sure why it was forced-push since it seems it was not a rebase on master either?

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: checkResponse reads the entire body with `readAll`(which reads till the stream ends) and throws error if status not `OK`. Why do we need this when we are reading the body using jsonDecoder?

kishansagathiya:
> There was a force-push here. Mentioned c3ca530 does not seem to exist anymore. I'm not sure why it was forced-push since it seems it was not a rebase on master either?

I screwed up the branch by resolving changes and pushing them here. What I forcepushed is your last commit `Remove some unreachable code `

ipfsconn/ipfshttp/ipfshttp.go
hsanjuan: `checkResponse` reads an IPFS error message body when the request status is not 200. This is so you can read what the actual error message is (if any). If the body is empty the method will wrongly succeed.

test/ipfs_mock.go
```diff
@@ -339,7 +346,14 @@ func (m *IpfsMock) handler(w http.ResponseWriter, r *http.Request) {
 			Ref: arg,
 		}
 		j, _ := json.Marshal(resp)
-		w.Write(j)
+		if arg == SlowCid1.String() {
+			for i := 0; i <= 5; i++ {
+				time.Sleep(2 * time.Second)
+				w.Write(j)
```
hsanjuan: Can you make sure the testing hits the missing lines here: https://codecov.io/gh/ipfs/ipfs-cluster/compare/e523215ee29b833c7cfdfb9315dd4a825a292f12...33602b9188b2edb621b9443f3c77dd016c2fccd3/diff#D1-367

The written object needs to have a `Progress` field that increases on every write.

ipfsconn/ipfshttp/ipfshttp_test.go
```diff
@@ -94,6 +94,25 @@ func testPin(t *testing.T, method string) {
 	if err == nil {
 		t.Error("expected error pinning cid")
 	}
+
+	if method == "refs" {
```
hsanjuan: use a switch. Makes flow clearer since these are excluding code paths.

PR #751, opened 2019-04-10
add swagger spec

lanzafame:


hsanjuan:
There should be an API endpoint that serves this (and also one to serve the pretty html representation). @kishansagathiya can you take over this?

kishansagathiya:
> @kishansagathiya can you take over this?

Sure thing

kishansagathiya:
I think this should be ready now. Let me know if this looks good to you @lanzafame.

I with have to raise a PR against the cluster-website, but do we serve the documentation as part of cluster as well?

hsanjuan:
> I with have to raise a PR against the cluster-website, but do we serve the documentation as part of cluster as well?

Yes, this would be most useful. An `/api` endpoint (or something else if there's a different convention). If `Accept` is set to json in the request, return the JSON of the spec. If set to `yaml` the yaml. Otherwise return the swagger API html website thing that is autogenerated from the spec (I haven't used swagger in many years but this was a thing).

hsanjuan:
It doesn't need to happen in this PR though

kishansagathiya:
You can also go through https://app.swaggerhub.com/apis-docs/kishansagathiya/ipfs-cluster_api/0.10.1 instead of going through the code.

api/rest/swagger.yaml
```diff
@@ -0,0 +1,791 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PinInfo'
+              example: {"cid":{"/":"zdj7WcGff6a9WbgRjz37svL6BoQRyzYAe2GZLBtkPULUk6rkS"},"peer_map":{"QmSJz3VA8cQVeegqZGsTkKQ4uiBXzWBE4r43KPb8woZbtH":{"cid":{"/":"zdj7WcGff6a9WbgRjz37svL6BoQRyzYAe2GZLBtkPULUk6rkS"},"peer":"QmSJz3VA8cQVeegqZGsTkKQ4uiBXzWBE4r43KPb8woZbtH","peername":"omou","status":"pinning","timestamp":"2019-04-08T16:21:38.680858768+10:00","error":""}}}
+        '400':
+          $ref: '#/components/responses/PinBadRequest'
+          
+  /health/graph:
+    get:
+      summary: Query all connected cluster peers and their ipfs peers to generate a graph of the connections
```
lanzafame: Summaries should be summaries, with full detail going into the `description`.

api/rest/swagger.yaml
```diff
@@ -0,0 +1,791 @@
...
+  /monitor/metrics/{name}:
+    parameters:
+      - name: name
+        in: path
+        required: true
+        schema:
+          type: string
+          enum: [ping, freespace]
+    get:
+      summary: Get the latest valid metrics of the given type logged by this peer for all current cluster peers
```
lanzafame: Summaries should be summaries, with full detail going into the `description`.

api/rest/swagger.yaml
```diff
@@ -0,0 +1,791 @@
...
+                $ref: '#/components/schemas/PinOptions'
+              example: {"replication_factor_min":-1,"replication_factor_max":-1,"name":"","shard_size":0,"user_allocations":null,"metadata":null,"cid":{"/":"zdj7WcGff6a9WbgRjz37svL6BoQRyzYAe2GZLBtkPULUk6rkS"},"type":2,"allocations":[],"max_depth":-1,"reference":null}
+        '400':
+          $ref: '#/components/responses/BadRequest'
+  /pins:
+    get:
+      summary: Get a list of Pins
+      parameters:
+        - $ref: '#/components/parameters/local'
+        - $ref: '#/components/parameters/filter'
```
lanzafame: An enum of possible values is definitely needed for the `filter` parameter.

lanzafame:
@kishansagathiya Also, as much as defining the common definitions for things is useful, sometimes it isn't worth it as it means another name for something or sometimes several things, i.e. we have a filter parameter that can't be reused if we want to specify what the accepted values are as they are different between the endpoints that use a filter parameter.

kishansagathiya:
@lanzafame Addressed comments in https://github.com/ipfs/ipfs-cluster/pull/751/commits/61b4d235abc275f5e31cc8716269e94a4ba38f4b

kishansagathiya:
Docs downloaded from swagger online tool doesn't look so good
https://kishansagathiya.github.io/swagger1.html

pkafei:
Are the swagger docs going to replace the API docs we currently have on the Cluster website? https://cluster.ipfs.io/documentation/developer/api/

kishansagathiya:
> Are the swagger docs going to replace the API docs we currently have on the Cluster website? https://cluster.ipfs.io/documentation/developer/api/

Yup, They are much more informative that what we have right now

meiqimichelle:
Do we have thoughts yet on how to display? ie, I saw https://app.swaggerhub.com/apis-docs/kishansagathiya/ipfs-cluster_api/0.10.1#/ -- SwaggerHub is new to me. Is that a tool we're thinking about buying to display the docs? Or just something you're using for development @kishansagathiya ?  

(I'm guessing I'll want to open an issue in the Cluster website repo re: display, unless there's already been a decision on that)

kishansagathiya:
@meiqimichelle 
> Is that a tool we're thinking about buying to display the docs? 

No need to buy. The UI there is swagger UI. We would just need to add that code(same as static files in this commit https://github.com/ipfs/ipfs-cluster/pull/751/commits/004b8b467c775c311175391c3dd3036ae8ba5f53)

When you say display do you mean design/UI of the page?

meiqimichelle:
Got it! And yes, on design/UI. I'll make an issue in the Cluster website repo about this so we (well, I) make sure not to forget it.

api/rest/restapi.go
```diff
@@ -258,6 +262,15 @@ func (api *API) addRoutes(router *mux.Router) {
...
 	}
+
+	statikFS, err := fs.New()
+	if err != nil {
+		panic(err)
+	}
+	staticServer := http.FileServer(statikFS)
+	sh := http.StripPrefix("/swaggerui/", staticServer)
+	router.PathPrefix("/swaggerui/").Handler(sh)
+
```
kishansagathiya: This is different than other routes, but it seems like to make the logic similar(write a HandlerFunc), I will have to re-implement or copy things that are already done in http.FileServer.
Can we keep it like this? Anyway this this endpoint is much different than rest.
@lanzafame

lanzafame: I would put it in a function `addSwaggerRoute(router *mux.Router)` and also it would be good to add the `ochttp` route tag:
```
router.PathPrefix("/swaggerui/").Handler(
    ochttp.WithRouteTag(
            http.HandlerFunc(sh),
	    "/swaggerui/",
    ),
)
```

kishansagathiya:
Trying to get it working without the trailing slash

kishansagathiya:
Problems with serving swaggerui with handlerFunc

- You have a file server where you are putting a directory, you want
  the route to be starting from `/swaggerui/` so that all contents in
  the directory(index.html, js, css) can be read.
- If your route starts from `/swaggerui` you can read only index.html,
  which would result in no output on the screen.
- Right now logic for serving swaggerui is differently written than
  other endpoints. You can write a handlerFunc with path like `/swaggerui/{path:.*}`,
  but in that you can serve everything, but index.html. Fileserver would always redirect
  `/index.html` to `/` and gorilla mux would redirect `/swaggerui/` to
  `/swaggerui` which is 404
- I can add an extra route say `/swaggerui`, and use the same handlerFunc
  as earlier, but some how that ends up in a redirection problem (page
  not being redirected properly)

Until, I find something better, we have hacky way where `/swaggerui` is
redirected to `/swaggerui/` to support without trailing slashes as well
(inside addSwaggerRoute). Somehow doing this along with every other
route also would cause redirection problems.

hsanjuan:
> Problems with serving swaggerui with handlerFunc

Have you tried https://golang.org/pkg/net/http/#ServeFile . It specifies what you have to do to serve index.html too.

kishansagathiya:
> Have you tried https://golang.org/pkg/net/http/#ServeFile . It specifies what you have to do to serve index.html too.

This could work, but in this case static files are read from the system (they will have to exist separately from the binary). I think it's better if we can compile static files with the binary and can pass a `http.FileSystem`.

Something like https://golang.org/src/net/http/fs.go?s=19279:19336#L540 would help, but it is not exported.

hsanjuan:
> > Have you tried https://golang.org/pkg/net/http/#ServeFile . It specifies what you have to do to serve index.html too.
> 
> This could work, but in this case static files are read from the system (they will have to exist separately from the binary). I think it's better if we can compile static files with the binary and can pass a `http.FileSystem`.
> 
> Something like https://golang.org/src/net/http/fs.go?s=19279:19336#L540 would help, but it is not exported.

True. Embedding a large js library in our binary is not something I'm fond of. Can we move this part to a new PR and merge the specs for the moment?

kishansagathiya:
@hsanjuan Have removed everything else except specs, commits don't make sense anymore, this should be squashed.

lanzafame:
@hsanjuan could you please squash+merge? I approve but can't actually approve as I authored the PR :expressionless:

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+  /version:
+    get:
+      summary: Get Version of Cluster
```
hsanjuan: Get the version of the cluster peer daemon.

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+      tags:
+        - cluster
+      description: Get the current version of the IPFS Cluster.
```
hsanjuan: `This endpoint returns the version of the IPFS Cluster daemon responding to the request.`

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
```
hsanjuan: put it in component/schemas like all the rest?

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+                type: object
+                properties:
+                  Version:
+                    type: string
+              example: {"Version":"0.10.0"}
+              
+  /add:
+    post:
+      summary: Add a File
```
hsanjuan: `Add a file`. I'm not sure if it's corret to Capitalize? @meiqimichelle ?

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+        - $ref: '#/components/parameters/wrap'
+        - $ref: '#/components/parameters/shard'
+        - $ref: '#/components/parameters/progress'
+        - $ref: '#/components/parameters/cidVersion'
+        - $ref: '#/components/parameters/streamChannels'
+        - $ref: '#/components/parameters/noCopy'
+        - $ref: '#/components/parameters/chunker'
+      tags:
+        - add
+      description: Add a file to IPFS via IPFS Cluster
```
hsanjuan: `This endpoint adds a file to the allocated IPFS daemons and pins it in the cluster. The cluster peer receiving the request will chunk and build the DAG, forwarding the resulting blocks to their final destination.`

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+      requestBody:
+        content:
+          multipart/form-data:
+            schema:
```
hsanjuan: move to schemas with the rest

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+              type: string
+              format: binary
+      responses:
```
hsanjuan: I don't know if swagger allows to express that this is a streaming response and that, in that case, a trailing header is used for errors. Otherwise this should be noted in the description at least.

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
```
hsanjuan: move to the schemas list.

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+                properties:
+                  size:
+                    type: integer
+                  name:
+                    type: string
+                  cid:
```
hsanjuan: maybe worth defining this part of the schema separately for all the places in which a CID is returned (I'm not sure swagger will allow nesting schemas like that)

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+                    type: object
+                    properties:
+                      /:
+                        type: string
+              example: {"name":"QmePw8gVcBMb8x6kAep6aMBAX23hCSk6iZW3i9VKkiFhu1","cid":{"/":"QmePw8gVcBMb8x6kAep6aMBAX23hCSk6iZW3i9VKkiFhu1"},"size":20}
+        '400':
+          $ref: '#/components/responses/BadRequest'
+  /peers:
+    get:
+      summary: Get list of Peers
```
hsanjuan: `Get the list of cluster peers`

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+      tags:
+        - peers
+      description: Get the current list of IPFS Cluster and IPFS peers that are in the cluster.
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
```
hsanjuan: move

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+
+    post:
+      summary: Add a new Peer to the Cluster
```
hsanjuan: Add a new peer to the cluster

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+      tags:
+        - peers
+      description: Add a new Peer to the Cluster
```
hsanjuan: The description should be a description not a copy of the summary.

That said, remove this from the docs altogether. This should not be used.

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+              
+  /peers/{peer}:
+    parameters:
+      - name: peer
+        in: path
+        required: true
+        schema:
+          type: string
+    delete:
+      summary: Remove a Peer from the Cluster
```
hsanjuan: Remove a peer from the Cluster

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+      tags:
+        - peers
+      description: Remove a Peer from the Cluster
```
hsanjuan: `This endpoint will attempt to remove a peer from a Raft-based cluster. If the removed peer is running, it will automatically shutdown. The operation can only succeed if a Raft leader exists. This operation has no  effect for CRDT-based Clusters.`

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+      responses:
+        '204':
+          description: No Content
+        '400':
```
hsanjuan: Likely missing a 500 when `timeout waiting for leader` I think.

api/rest/swagger.yaml
```diff
@@ -0,0 +1,796 @@
...
+        peer:
+          type: string
+        value:
+          type: string
+        expire:
+          type: integer
+        valid:
+          type: boolean
+          
+  parameters:
```
hsanjuan: descriptions missing from most parameters

kishansagathiya:
> If needed, we could even run a "fake", open, cluster peer for the task.

What do you mean, by "fake" here? We would need to do some tricks to make sure that users cannot actually add pins or peers, but still see proper response. Doing all that might be much more work compared to just maintaining examples.

Best way to let users try out the api is giving the swagger endpoint in cluster apis. 

Hosted Swagger UI would throw errors (CORS). One that swaggerhub gives (like https://app.swaggerhub.com/apis/kishansagathiya/ipfs-cluster_api/0.10.1) would always throw 403

People can use inspector.swagger.io, but I found it really slow, and they also recommend to download an extension for it to work well.

hsanjuan:
> What do you mean, by "fake" here? We would need to do some tricks to make sure that users cannot actually add pins or peers, but still see proper response. Doing all that might be much more work compared to just maintaining examples.

Compiling a cluster peer with the mock RPC (or better the mock IPFS connector) is just changing a couple of lines.

We can keep the examples for the moment but then they need to be formatted.

api/rest/swagger.yaml
kishansagathiya: seems like swagger ui is not capable of handling these headers `Trailer`, `Transfer-Encoding` as they are forbidden by browsers 
https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name 
https://swagger.io/docs/open-source-tools/swagger-ui/usage/limitations/

hsanjuan: Those are response headers. The docs you link are not related since they are about browsers not allowing some headers to be modified programatically (when making requests).

kishansagathiya:
At a bunch of places PinOptions is used as response instead of Pins, changing that.

api/rest/swagger.yaml
kishansagathiya: I would add a description for progress, but it doesn't seem like it is being used. 
Not sure what I should write for shard size

hsanjuan: progress is used (but handled in the `addutils/` helper probably). `progress=false` disables streaming and returns an array of status at the end (progress=false actually makes the /add endpoint restful).

hsanjuan: you can leave sharding options undocumented, those are not used.

kishansagathiya: sharding options are not used by cli, but they are still used by the apis.

> progress is used (but handled in the addutils/ helper probably). progress=false disables streaming and returns an array of status at the end (progress=false actually makes the /add endpoint restful).

I see `params.StreamChannels` doing this.https://github.com/ipfs/ipfs-cluster/blob/e523215ee29b833c7cfdfb9315dd4a825a292f12/adder/adderutils/adderutils.go#L59-L90

hsanjuan: sorry, you are right, we don't yet use progress for anything.

kishansagathiya:
@hsanjuan Do you have more comments for this?

api/rest/swagger.yaml
```diff
@@ -0,0 +1,792 @@
...
+
+  /allocations/{cid}:
+    parameters:
+      - name: cid
+        in: path
+        required: true
+        schema:
+          type: string
+    get:
+      summary: Get Allocation information of a Pin.
```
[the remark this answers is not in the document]
meiqimichelle: Agree. We should default to lower-case. Upper-case should be the exception (at the moment, I can't think of any exceptions).

meiqimichelle:
Please let me know when this is ready for a copyedit review, and if there's a UI, when that's ready for review (wasn't sure if that was included in this PR). If there *is* a UI that should be reviewed here, please point me to where I can preview it, or instructions to run it locally. Thx! (Might also separate UI review into a different PR/review cycle.)

lanzafame:
@abn fyi, this is IPFS Cluster’s WIP swagger spec.

hsanjuan:
I'm at a point in life where I don't want to deal with Swagger documentation by hand ever again, so closing this. It's a pain to have it both be accurate and up-to-date, and I prefer not having it at all.

PR #762, opened 2019-04-24
Default allocations filter should be all

kishansagathiya:
Currently `curl -X GET http://localhost:9094/allocations` results in an
empty array, because no filter is provided. Solution to this would be
either 1) to consider filter as all if no filter is provided by the user
or 2) to make the `filter` parameter mandatory and reply with BadRequest
if no filter is provided

This commit makes the default filter as all

### DIFF
--- a/api/rest/restapi.go
+++ b/api/rest/restapi.go
@@ -743,6 +743,12 @@ func (api *API) allocationsHandler(w http.ResponseWriter, r *http.Request) {
 	for _, f := range strings.Split(filterStr, ",") {
 		filter |= types.PinTypeFromString(f)
 	}
+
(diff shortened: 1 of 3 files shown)


hsanjuan:
Good catch here. In case when the filter turns out to be AllFilter, we should shortcut and not loop all the pins comparing each one.

I'm ok changing `""` to mean `AllType`. We should probably check that the resulting pin is not `BadType`. If we had checked this before, the problem would have likely come up. And we need a test (extend the test making a request without filter).

hsanjuan:
If the filter results `BadType`, I think we can respond with `BadRequest`.

PR #768, opened 2019-04-29
Proxy: hijack pin/update

### LINKED ISSUE #732: [Proxy] Intercept pin/update
#### Pre-check

#### Basic information

  * [x] Master
  * [ ] Latest stable version
  * [ ] Bug
  * [x] Feature request
  * [ ] Enhancement
  * [x] Linux
  * [ ] macOS
  * [ ] Windows
  * [ ] Other: `which?`
  * [ ] Built from sources
  * [ ] Docker
  * [ ] Snap
  * [ ] Other: `which?`

#### Description
It would be great to also intercept the pin update command. For us in [Peergos](https://github.com/peergos/peergos), 99% of our calls are pin update, and we'd love to use ipfs-cluster for large deployments.

hsanjuan:
The IPFS pin/update endpoint takes two arguments and usually
unpins the first and pins the second. It is a bit more efficient
to do it in a single operation than two separate ones.

This will make the proxy endpoint hijack pin/update requests.

First, the FROM pin is fetched from the state. If present, we
set the options (replication factors, actual allocations) from
that pin to the new one. Then we pin the TO item and proceed
to unpin the FROM item when `unpin` is not false.

We need to support path resolving, just like IPFS, therefore
it was necessary to expose IPFSResolve() via RPC.

Goes on top of #767 

Fixes #732

### DIFF
--- a/api/ipfsproxy/ipfsproxy.go
+++ b/api/ipfsproxy/ipfsproxy.go
@@ -288,10 +288,14 @@ func (proxy *Server) run() {
 }
 
 // ipfsErrorResponder writes an http error response just like IPFS would.
-func ipfsErrorResponder(w http.ResponseWriter, errMsg string) {
+func ipfsErrorResponder(w http.ResponseWriter, errMsg string, code int) {
 	res := ipfsError{errMsg}
 	resBytes, _ := json.Marshal(res)
-	w.WriteHeader(http.StatusInternalServerError)
+	if code > 0 {
+		w.WriteHeader(code)
+	} else {
+		w.WriteHeader(http.StatusInternalServerError)
+	}
 	w.Write(resBytes)
 	return
 }
@@ -302,7 +306,7 @@ func (proxy *Server) pinOpHandler(op string, w http.ResponseWriter, r *http.Requ
 	arg := r.URL.Query().Get("arg")
 	p, err := path.ParsePath(arg)
 	if err != nil {
-		ipfsErrorResponder(w, "Error parsing IPFS Path: "+err.Error())
+		ipfsErrorResponder(w, "Error parsing IPFS Path: "+err.Error(), -1)
 		return
 	}
 
@@ -316,7 +320,7 @@ func (proxy *Server) pinOpHandler(op string, w http.ResponseWriter, r *http.Requ
 		&pin,
 	)
 	if err != nil {
-		ipfsErrorResponder(w, err.Error())
+		ipfsErrorResponder(w, err.Error(), -1)
 		return
 	}
 
@@ -347,7 +351,7 @@ func (proxy *Server) pinLsHandler(w http.ResponseWriter, r *http.Request) {
 	if arg != "" {
 		c, err := cid.Decode(arg)
 		if err != nil {
-			ipfsErrorResponder(w, err.Error())
+			ipfsErrorResponder(w, err.Error(), -1)
 			return
 		}
 		var pin api.Pin
@@ -359,7 +363,7 @@ func (proxy *Server) pinLsHandler(w http.ResponseWriter, r *http.Request) {
 			&pin,
 		)
 		if err != nil {
-			ipfsErrorResponder(w, fmt.Sprintf("Error: path '%s' is not pinned", arg))
+			ipfsErrorResponder(w, fmt.Sprintf("Error: path '%s' is not pinned", arg), -1)
 			return
 		}
 		pinLs.Keys[pin.Cid.String()] = ipfsPinType{
@@ -375,7 +379,7 @@ func (proxy *Server) pinLsHandler(w http.ResponseWriter, r *http.Request) {
 			&pins,
 		)
 		if err != nil {
-			ipfsErrorResponder(w, err.Error())
+			ipfsErrorResponder(w, err.Error(), -1)
 			return
 		}
 
@@ -401,11 +405,11 @@ func (proxy *Server) pinUpdateHandler(w http.ResponseWriter, r *http.Request) {
 	q := r.URL.Query()
 	args := q["arg"]
 	if len(args) == 0 {
-		ipfsErrorResponder(w, "argument \"from-path\" is required")
+		ipfsErrorResponder(w, "argument \"from-path\" is required", http.StatusBadRequest)
 		return
 	}
 	if len(args) == 1 {
-		ipfsErrorResponder(w, "argument \"to-path\" is required")
+		ipfsErrorResponder(w, "argument \"to-path\" is required", http.StatusBadRequest)
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


api/ipfsproxy/ipfsproxy_test.go
```diff
@@ -314,6 +314,108 @@ func TestIPFSProxyUnpin(t *testing.T) {
 	}
 }
 
+func TestIPFSProxyPinUpdate(t *testing.T) {
```
lanzafame: This may be a stupid idea, but for the proxy tests, it may be 'beneficial' to make the same requests against an ipfs api and use the response for the comparison with the cluster proxy response.

hsanjuan: We can't mimic IPFS responses fully, particularly cluster error responses will need to be with cluster errors. In any case, we don't even use the IPFS mock to test this component, and rely fully on the RPC mock. It would not hurt to have end-2-end tests here. And in general, switch to using a real IPFS daemon and not the mock.

api/ipfsproxy/ipfsproxy_test.go
```diff
@@ -314,6 +314,108 @@ func TestIPFSProxyUnpin(t *testing.T) {
...
+		var resp ipfsPinOpResp
+		resBytes, _ := ioutil.ReadAll(res.Body)
+		err = json.Unmarshal(resBytes, &resp)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if len(resp.Pins) != 2 ||
+			resp.Pins[0] != test.PathIPFS1 ||
+			resp.Pins[1] != test.PathIPFS2 {
+			t.Errorf("bad response: %s", string(resBytes))
```
kishansagathiya: If `to` is recursively pinned and `from` is not, ipfs throws an error, but our hijacked API succeeds.
Not sure what would be the right thing to here, us to mimick ipfs or otherwise.
```
$ curl "http://localhost:5101/api/v0/pin/update?arg=QmNMgLz2jbCNpsqoBvYzzHZkkTbMbiuAxgjhcztrserJXC&arg=Qmdejgw83ccxAGb2ALDRJCso8YVwkqT992KTr9yhkX4ynk&unpin=false"
{"Message":"'from' cid was not recursively pinned already","Code":0,"Type":"error"}
```

ianopolous:
Note that using pin update rather than pin add + pin rm in ipfs is not just a little bit more efficient. It is exponentially faster (O(D) vs O(e^D) for the original use case - updating a mapping in a CHAMP). This speed is critical for our usage in Peergos as operations on even small champs end up being 100s of times faster.

lanzafame:
@ianopolous unfortunately using the actual pin/update call will require more extensive changes as the calls to the underlying ipfs node are handled by the pintracker via the way of the consensus layer.

ianopolous:
@lanzafame Yep, I realise this is a stop gap until an optimised version can be written. Just wanted to point out the huge speed up available.

api/ipfsproxy/ipfsproxy_test.go
hsanjuan: Our API fails when from is not pinned (or should, as it is now). There's a test for this.

hsanjuan:
> @lanzafame Yep, I realise this is a stop gap until an optimised version can be written. Just wanted to point out the huge speed up available.

I realize I'm probably not up to date on the ipfs pinning system internals anymore. Let's get this in first so that it can be used and then work on the optimized version (cluster could also optimize the 2 operations a bit).

api/ipfsproxy/ipfsproxy_test.go
kishansagathiya: > Our API fails when from is not pinned (or should, as it is now). There's a test for this.

It's not just about being pinned or not. It's about whether it is recursively pinned or not.

api/ipfsproxy/ipfsproxy.go
```diff
@@ -387,6 +391,122 @@ func (proxy *Server) pinLsHandler(w http.ResponseWriter, r *http.Request) {
...
+	ctx, span := trace.StartSpan(r.Context(), "ipfsproxy/pinUpdateHandler")
+	defer span.End()
+
+	proxy.setHeaders(w.Header(), r)
+
+	// Check that we have enough arguments and mimic ipfs response when not
+	q := r.URL.Query()
+	args := q["arg"]
+	if len(args) == 0 {
+		ipfsErrorResponder(w, "argument \"from-path\" is required")
```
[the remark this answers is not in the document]
hsanjuan: This is consistent with how the ipfs daemon responds in any case right?

api/ipfsproxy/ipfsproxy_test.go
hsanjuan: Yes I checked this. Re-open if I missed something.

api/ipfsproxy/ipfsproxy_test.go
```diff
@@ -314,6 +314,108 @@ func TestIPFSProxyUnpin(t *testing.T) {
...
+
+	t.Run("pin/update bad args", func(t *testing.T) {
+		res, err := http.Post(fmt.Sprintf("%s/pin/update", proxyURL(proxy)), "", nil)
+		if err != nil {
+			t.Fatal("request should complete: ", err)
+		}
+
+		defer res.Body.Close()
+		if res.StatusCode != http.StatusInternalServerError {
+			t.Error("request should not be successful with a no arguments")
```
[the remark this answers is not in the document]
hsanjuan: Fixed

api/ipfsproxy/ipfsproxy_test.go
```diff
@@ -314,6 +314,108 @@ func TestIPFSProxyUnpin(t *testing.T) {
...
+		}
+
+		res2, err := http.Post(fmt.Sprintf("%s/pin/update?arg=%s", proxyURL(proxy), test.PathIPFS1), "", nil)
+		if err != nil {
+			t.Fatal("request should complete: ", err)
+		}
+
+		defer res2.Body.Close()
+		if res2.StatusCode != http.StatusInternalServerError {
+			t.Error("request should not be successful with a single argument")
```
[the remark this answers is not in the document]
hsanjuan: Fixed

api/ipfsproxy/ipfsproxy.go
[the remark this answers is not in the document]
kishansagathiya: No,
This is what I see after your last commit
```
$ curl http://localhost:5001/api/v0/pin/update
argument "from-path" is required
$ curl http://localhost:5001/api/v0/pin/update/
argument "to-path" is required
$ curl http://localhost:9095/api/v0/pin/update
{"Message":"argument \"from-path\" is required"}
$ curl http://localhost:9095/api/v0/pin/update/
argument "to-path" is required                                                                                                                                                        
```

I am hoping that these error would be in JSON once https://github.com/ipfs/go-ipfs/issues/6269 is resolved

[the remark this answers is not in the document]
hsanjuan: calls ending with `/` are not handled and forwarded to the ipfs daemon (which I think always error).

I'm fine not answering text errors in cluster even if ipfs does. This is an ipfs bug. I'm not sure what you want here, do you want to handle `pin/update/` ?

api/ipfsproxy/ipfsproxy_test.go
kishansagathiya: Strangely enough I can't reproduce it again on ipfs, even after using the same files. So, let it be.

api/ipfsproxy/ipfsproxy.go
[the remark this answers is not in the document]
kishansagathiya: Let it be, It's fine I think. We don't have to handle that.

hsanjuan:
Thanks! I'll have another run at that sharness test. It seems an unrelated failure :/

PR #770, opened 2019-04-29
RFC: CRDT next steps

hsanjuan:
This document is for discussion only. Do Not Merge.

### DIFF
--- a/rfcs/followers.md
+++ b/rfcs/followers.md
@@ -1,0 +1,106 @@
+RFC: Follower cluster peers
+===========================
+
+
+1. Per https://github.com/ipfs/ipfs-cluster/pull/710/files we will introduce
+   RPC authorization policies:
+
+  * `raft` policy is still pretty much fully open (all endpoints that can
+  potentially be used by someone else are open to everyone
+  * `crdtsoft` allows trusted peers to perform sync/recover/block put etc on any other peer
+  * `crdtstrict` allows only read-only methods to trusted peers and almost nothing to the rest (connect swarms (?))
+
+Running with `crdtsoft` policy where all peers trust all the other peers
+should be equivalent to running with `raft` policy (right?).
+
+Regarding the configuration:
+
+* `rpc_auth_policy` key will be introduced in the main `cluster` section
+* `trusted_peers` array for trusted peer IDs will be introduced in the main `cluster` section.
+
+This brings the question:
+
+##  Does it make sense to make `raft` policy selectable (with a different name) for "trusted" environments running with CRDTS (so that the user does not need to add/rm trusted peers when they join/leave the cluster?
+
+I'd say yes. Options:
+
+* We can settle in 3 `rpc_auth_policy`: `open`, `trusted` and `strict`. Unless
+  defined, `ipfs-cluster-service` will run with `open` for raft and `trusted`
+  for crdts.
+* We remove the `open` option and allow a `trusted_peers: all` setting
+  (default?). Maybe even letting `trusted_peers` be a subkey of the
+  `rpc_auth_policy`.
+* More? Overall I'm not too happy here. Maybe we can hide `rpc_auth_policy` completely:
+  * When trusted_peers == all -> we run with `trusted`
+  * When trusted_peers has some peer but peer started with
+    `ipfs-cluster-service daemon` we use `trusted` (raft users should never
+    set `trusted_peers`.
+  * When we start with `ipfs-cluster-service follow` we use `strict`.
+
+More options?
+
+---
+
+2. We are separating `ID+PrivateKey` into a separate `identity.json`
+   file. https://github.com/ipfs/ipfs-cluster/pull/766
+
+---
+
+3. `ipfs-cluster-service --config xxx` (or without the flag when using a
+   default config location) will autogenerate an identity when it does not
+   exist in `~/ipfs-cluster/` nor in the existing configuration.
+
+---
+
+4. Support for remote configs is added via `ipfs-cluster-service --config
+   <ipfspath>` so that `ipfs-cluster-service --config /ipfs/xxx daemon` will:
+  * Create a default ipfs connector config and apply env vars (to let the user
+    modify the node multiaddress)
+  * Fetch the config file from IPFS
+  * Use that to initqialize the IPFS Connector
+  * Use the identity and the fetched configuration to launch the cluster peer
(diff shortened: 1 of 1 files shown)


rfcs/followers.md
```diff
@@ -0,0 +1,106 @@
...
+should be equivalent to running with `raft` policy (right?).
+
+Regarding the configuration:
+
+* `rpc_auth_policy` key will be introduced in the main `cluster` section
+* `trusted_peers` array for trusted peer IDs will be introduced in the main `cluster` section.
+
+This brings the question:
+
+##  Does it make sense to make `raft` policy selectable (with a different name) for "trusted" environments running with CRDTS (so that the user does not need to add/rm trusted peers when they join/leave the cluster?
```
hsanjuan: I mean, when you want everyone in the cluster to be able to commit pins and unpins.

rfcs/followers.md
```diff
@@ -0,0 +1,106 @@
...
+---
+
+2. We are separating `ID+PrivateKey` into a separate `identity.json`
+   file. https://github.com/ipfs/ipfs-cluster/pull/766
+
+---
+
+3. `ipfs-cluster-service --config xxx` (or without the flag when using a
+   default config location) will autogenerate an identity when it does not
+   exist in `~/ipfs-cluster/` nor in the existing configuration.
```
kishansagathiya: If you mean this only for followers, it should be fine.

rfcs/followers.md
```diff
@@ -0,0 +1,106 @@
...
+  * Use the identity and the fetched configuration to launch the cluster peer
+  * Auto-generate and store an identity if no existing one can be found
+  * Note that the peer does not store the configuration on disk. It just runs
+    with it.
+  * It would be good to support `dnslink` here too, somehow.
+
+---
+
+5. On the same line, `ipfs-cluster-service init --from /ipfs/xxx` downloads a
+   config to `~/.ipfs-cluster`
```
kishansagathiya: `ipfs-cluster-service init --from configpath`

This would be helpful option in general (not just for followers)

rfcs/followers.md
```diff
@@ -0,0 +1,106 @@
...
+
+---
+
+6. The cluster section in the configuration re-introduces a `bootstrappers`
+   key. These bootstrappers are used when:
+  * No `--bootstrap` flag is provided
+  * Starting `raft` for the first time (but not when a raft state exists).
+  * Always when using CRDTs.
```
kishansagathiya: This could greatly improve the user flow. This could eliminate the need to initialize manually. 
Users would be able to start cluster directly with `ipfs-cluster-service --config configpath daemon/follow`

rfcs/followers.md
```diff
@@ -0,0 +1,106 @@
...
+
+---
+
+9. We remove the `secret` from the main `cluster` configuration section and
+   move it to the `raft` (visible, same usage as now, with warning when not
+   set).
+
+---
+
+10. We add a hidden `network_key` to the `crdt` configuration section. When
```
lanzafame:
```suggestion
10. We add a optional `network_key` to the `crdt` configuration section. When
```

rfcs/followers.md
```diff
@@ -0,0 +1,106 @@
+RFC: Follower cluster peers
+===========================
+
```
lanzafame: @hsanjuan This RFC is very implementation heavy and currently lack any context or definition of what we are attempting to implement. 
Could we get a draft of the following things:
- What a `follower` peer is?
- How a `follower` follows a cluster? Not the command invocation but logically.
- What part does a `follower` play in consensus, if any at all?
- A `follower` peer suggests there is a `non-follower` entity, what is it and what is its functionality/responsibilities?

This information will make it clear why, for example, we need different RPC auth policies etc. I think it is important that we get the non-implemenation details clarified so that we understand why the things in the rest of the doc are required and also so anyone outside of the dev team knows what this is about.

Sidenote: I would like to see the high-level specification of what follower peers are actually merged.

rfcs/followers.md
hsanjuan: Not only followers.

rfcs/followers.md
hsanjuan: It's not just for followeres

rfcs/followers.md
hsanjuan: Optional, hidden by default.

rfcs/followers.md
hsanjuan: So this is not user documentation. This builds on extensive previous discussions in the dev team and is for the dev team. Perhaps I've made a bad mistake using `followers` in the name because this is more about auth, shorthands and possibilities for initialization.

To recap: A follower is a `--consensus crdt` peer with defined `trusted peers`  and a strict rpc policy. It just like any other peer. A follower just an alias and a shorthand command to run a cluster peer with those things. I thought this was super clear at this point. There is nothing special in what a follower is vs. any other peer.

Now, we can align with the actual implemetation approaches here, we can maybe write proper explanations for everyone else and you can make the case for the nuances of not considering followers part of a "cluster" in the docs etc.

The main issue here is to settle on how to do `1.`.

rfcs/followers.md
```diff
@@ -0,0 +1,106 @@
...
+  * `crdtsoft` allows trusted peers to perform sync/recover/block put etc on any other peer
+  * `crdtstrict` allows only read-only methods to trusted peers and almost nothing to the rest (connect swarms (?))
+
+Running with `crdtsoft` policy where all peers trust all the other peers
+should be equivalent to running with `raft` policy (right?).
+
+Regarding the configuration:
+
+* `rpc_auth_policy` key will be introduced in the main `cluster` section
+* `trusted_peers` array for trusted peer IDs will be introduced in the main `cluster` section.
```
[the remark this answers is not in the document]
hsanjuan: The point in this is to be able to fetch the list of trusted peers as part of the template when you initialize with an `/ipfs/Qm...` configuration, for example.

rfcs/followers.md
hsanjuan: yes, essentially

rfcs/followers.md
lanzafame: I am not advocating for user docs but for implementers docs, i.e. from an implementers perspective what is a follower peer.

hsanjuan:
after feedback I will rewrite this with more explanations.

PR #773, opened 2019-05-04
RPC: Give each component a different RPC Service

hsanjuan:
I had thought of this for a very long time but there were no compelling
reasons to do it. Specifying RPC endpoint permissions becomes however
significantly nicer if each Component is a different RPC Service. This also
fixes some naming issues like having to prefix methods with the component name
to separate them from methods named in the same way in some other component
(Pin and IPFSPin).

Main changes here: in `rpc_api.go`. The rest is just renamings and adapting the tests.

### DIFF
--- a/adder/local/dag_service.go
+++ b/adder/local/dag_service.go
@@ -33,7 +33,7 @@ type DAGService struct {
 }
 
 // New returns a new Adder with the given rpc Client. The client is used
-// to perform calls to IPFSBlockPut and Pin content on Cluster.
+// to perform calls to IPFS.BlockPut and Pin content on Cluster.
 func New(rpc *rpc.Client, opts api.PinOptions) *DAGService {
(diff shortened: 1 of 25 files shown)


kishansagathiya:
> Specifying RPC endpoint permissions becomes however
significantly nicer if each Component is a different RPC Service.

Instead of giving permission to call a specific method, you want to give permissions for a rpc service?

kishansagathiya:
How does this impact performance?
Some rpc methods would involve calling few other rpc methods. Earlier a `cluster` method would call another `cluster` method, but now a `cluster` method say calls `pintracker` methods(where cluster and pintracker are rpc services). Does this impact performance of rpc calls?

hsanjuan:
This is just a renaming thing. It does not really affect performance or changes how we'll do auth. It just makes it prettier and clearer.

PR #710, opened 2019-03-08
RPC Authorization

kishansagathiya:
This PR adds permission policy for raft consensus and adds
permission policy to config(but hides it from user).
    
It proposes minimum set of RPC methods which we have to allow peers in
order for all cluster features to work as expected. It restricts peers
from accessing methods that are not part of the proposed RPC method set.
    
Proposed authorization would work only for raft consensus.
`authorizeWithPolicy` would need to be edited in order to support
 upcoming consensus mechanisms.

#666

PR #775, opened 2019-05-09
Feat: RPC Authorization

hsanjuan:
This enables RPC Authorization in IPFS Cluster.

For this we use an authorization policy which classifies each cluster peer RPC Endpoint in one of: Closed, Trusted, Open:

* Closed endpoints are always disallowed for remote calls.
* Trusted endpoints are only allowed for "trusted" peers.
* Open endpoints are allowed for everyone.

What a "Trusted" peer is established by the Consensus component:

* For Raft: all peers are "trusted". As before, Raft clusters must still run in a trusted environment and the cluster "secret" (swarm protector key) is the pre-shared-secret that locks the clusters from any other peer.
* For CRDT: "trusted_peers" are defined in the configuration and loaded on boot. It should be "fine" to run a CRDT cluster without a secret.

Online edition of the trusted peerset (for CRDTs) would be possible with something like `ipfs-cluster-ctl peers trust <pid>` but that's out of scope of this PR (see commit messages for explanation on why it cannot be part of `peers add`).

Finally, this also enables the pubsub topic validator in CRDTs to only consider messages from trusted peers.


hsanjuan:
Overseeds #710

hsanjuan:
Forgot to say: easier review by checking each commit separately.

hsanjuan:
> Looks good, seems to be missing a ctl cmd to trust a new node?

Will come in different PR. Not particularly pressed for that.

cluster.go
```diff
@@ -758,6 +718,30 @@ func (c *Cluster) Join(ctx context.Context, addr ma.Multiaddr) error {
 		return err
 	}
 
+	// Log a fake but valid metric from the peer we are
```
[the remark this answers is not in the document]
hsanjuan: It's "fake" because we pretend it comes from a different peer (the one we are bootstrapping to). We are not sending a metric, we are "receiving" it.

cluster.go
```diff
@@ -767,6 +751,12 @@ func (c *Cluster) Join(ctx context.Context, addr ma.Multiaddr) error {
 		c.dht.BootstrapOnce(ctx, dht.DefaultBootstrapConfig)
 	}()
 
+	// ConnectSwarms in the background after a while, when we have likely
+	// received some metrics.
```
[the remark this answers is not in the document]
hsanjuan: There might be several reasons for not having received metrics, but it does not really matter if ConnectSwarms does nothing then... it's a best effort thing. If we want it to enforce good ipfs swarm connectivity we should have cluster peers run it regularly (we can do this in a different PR, in fact it's probably very useful).

rpc_auth.go
```diff
@@ -0,0 +1,87 @@
...
+
+// A trick to find where something is used (i.e. Cluster.Pin):
+// grep -R -B 3 '"Pin"' | grep -C 1 '"Cluster"'.
+// This does not cover globalPinInfo*(...) broadcasts nor redirects to leader
+// in Raft.
+
+// DefaultRPCPolicy associates all rpc endpoints offered by cluster peers to an
+// endpoint type. See rpcutil/policygen.go as a quick way to generate this
+// without missing any endpoint.
+var DefaultRPCPolicy = map[string]RPCEndpointType{
```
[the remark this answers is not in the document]
hsanjuan: Indeed, yes good idea..

rpc_auth.go
```diff
@@ -0,0 +1,87 @@
...
+	"IPFSConnector.Resolve":       RPCClosed,
+	"IPFSConnector.SwarmPeers":    RPCTrusted, // connectGraph.
+	"IPFSConnector.Unpin":         RPCClosed,
+
+	// Consensus methods
+	"Consensus.AddPeer":  RPCTrusted, // Raft redirect to leader
+	"Consensus.LogPin":   RPCTrusted, // Raft redirect to leader
+	"Consensus.LogUnpin": RPCTrusted, // Raft redirect to leader
+	"Consensus.Peers":    RPCClosed,
+	"Consensus.RmPeer":   RPCTrusted, // Raft redirect to leader
```
kishansagathiya: In case of crdts, all consensus method could be `RPCClosed`, right?
This makes me wonder if we should have different default rpc policies for crdt and raft.

consensus/crdt/consensus.go
```diff
@@ -276,6 +282,28 @@ func (css *Consensus) Ready(ctx context.Context) <-chan struct{} {
 	return css.readyCh
 }
 
+// IsTrustedPeer returns whether the given peer is taken into account
+// when submitting updates to the consensus state.
```
kishansagathiya: If the given peer is not taken into account when submitting the updates to the state, it doesn't really help by being part of the cluster, does it?
Do you mean to say that the replication factor will achieved by the trusted peers and whatever followers do will be considered as bonus?

consensus/raft/consensus.go
```diff
@@ -229,6 +229,24 @@ func (cc *Consensus) Ready(ctx context.Context) <-chan struct{} {
...
 }
 
+// IsTrustedPeer returns true. In Raft we trust all peers.
+//
+// TODO: There is a potential improvement here by just trusting peers which
+// are currently part of the Raft peerset. But we cannot call Peers() every
+// single time and should keep a cache of trusted peers indexed by peer which
+// is difficult to maintain when peers are removed (we get no
+// notification). All in all, since PeerAdd is effectively open to everyone in
+// Cluster, it makes little difference.
```
kishansagathiya: I think for now only use-case for raft is where the cluster is private and all nodes are handled by one entity. Unless, some new use case come up, I think we should just trust all raft peers.

ipfscluster.go
```diff
@@ -53,6 +53,15 @@ type Consensus interface {
 	Clean(context.Context) error
 	// Peers returns the peerset participating in the Consensus
 	Peers(context.Context) ([]peer.ID, error)
+	// IsTrustedPeer returns true if the given peer is "trusted"
+	// This will grant access to more rpc endpoints and a
+	// non-trusted one. This should be fast as it will be
+	// called repeteadly for every remote RPC request.
```
kishansagathiya: // IsTrustedPeer returns true if the given peer is "trusted".

ipfscluster.go
```diff
@@ -53,6 +53,15 @@ type Consensus interface {
...
+	IsTrustedPeer(context.Context, peer.ID) bool
+	// Trust marks a peer as "trusted"
```
kishansagathiya: // Trust marks a peer as "trusted".
// This will grant access to more rpc endpoints than a
// non-trusted one. This should be fast as it will be
// called repeatedly for every remote RPC request.

rpcutil/policygen/policygen.go
```diff
@@ -0,0 +1,66 @@
...
+	"reflect"
+
+	cluster "github.com/ipfs/ipfs-cluster"
+)
+
+type svcIdComponent interface {
+	SvcID() string
+}
+
+func rpcTypeStr(t cluster.RPCEndpointType) string {
```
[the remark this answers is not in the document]
hsanjuan: A String() method should print human names like "closed", "trusted" but we have no use for those and I don't want to make an public API method for something internal (unfortunately I can't place all this in an `internal/` submodule due to cycle deps.

rpc_auth.go
hsanjuan: We make the assumption trusted peers are trusted and therefore can call those methods (which they call indirectly anyway by doing pubsub broadcasts). Also add/rm peers are noops.

consensus/crdt/consensus.go
hsanjuan: > If the given peer is not taken into account when submitting the updates to the state, it doesn't really help by being part of the cluster, does it?

It can be any cluster peer that this peer does not trust (like what we call a "follower"). Anyone can broadcast an update, but peers choose which ones to take into account (only those from trusted peers).

> Do you mean to say that the replication factor will achieved by the trusted peers and whatever followers do will be considered as bonus?

No

consensus/raft/consensus.go
hsanjuan: ok

rpcutil/policygen/policygen.go
```diff
@@ -29,28 +49,42 @@ func main() {
...
-// into rpc_auth.go and set the NEW endpoints to their correct type.`)
-	fmt.Println()
+// from the existing one, marking new endpoints as NEW. Redirect stdout
+// into rpc_policy.go and set the NEW endpoints to their correct type (make
+// sure you have recompiled this binary with the current version of the code).
+============================================================================`)
+	fmt.Fprintln(os.Stderr)
 
-	fmt.Println("var RPCPolicy = map[string]RPCEndpointType{")
+	fmt.Println("package ipfscluster")
```
lanzafame: I suggest writing this all to a byte slice and then running [`format.Source`](https://golang.org/pkg/go/format/#Source) for format the output. This would prevent compile time errors from wiping out the existing `rpc_policy.go` file as happened when I was testing this and the unused `fmt` import caused `go fmt` to bork (I ran `make install`).

hsanjuan: Thanks, that's a way better way

hsanjuan:
approved via slack

Issue #771, opened 2019-05-01
Daemon using CRDTs fails to start on Raspberry Pi


lanzafame:
**Additional information:**
 - OS: linux
 - IPFS Cluster version: master
 - Installation method: built from source for ARM

Running ipfs-cluster-service daemon on a Raspberry Pi 3 B with CRDT consensus fails due to badger not being able to `mmap` memory. 
```
$ ipfs-cluster-service daemon --consensus=crdt
06:07:40.790  INFO    service: Initializing. For verbose output run with "-l debug". Please wait... daemon.go:48
error creating datastore: Mmap value log file. Path=/home/pi/.ipfs-cluster/badger/000000.vlog. Error=cannot allocate memory
```

I found this [badgerdb issue](https://github.com/dgraph-io/badger/issues/610) which is resolved and if I ran that main.go on the RPi, it worked but if I changed the `options.ValueLogLoadingMode` back to the default, which is to use `options.MemoryMap`, that example program fails. 

@hsanjuan when you get the chance could you please confirm this? 

The fix, if this is a problem, is to allow passing through the configuration of value for `options.ValueLogLoadingMode` to badger.

hsanjuan:
@lanzafame `ValueLogFileSize` is set to 1GB by default. Can you try reducing it? How much RAM is there in these Raspberry Pis ? Ideally we would automatically handle this in systems with lower memory somehow. We can always expose [needed] badger options though.

hsanjuan:
Closed by #776 and configuring file loading mode for badger.

PR #776, opened 2019-05-10
expose badger loading mode conf

lanzafame:
Resolves: #771.

For simplicity, I haven't exposed the entire
badger config. Also, due to go default values and
the badger implementation, I have had to extract the
two loading mode config values so we can determine
whether it has actually has been set by the user in
the json config file.

License: MIT


datastore/badger/config.go
```diff
@@ -81,6 +112,12 @@ func (cfg *Config) LoadJSON(raw []byte) error {
 
 func (cfg *Config) applyJSONConfig(jcfg *jsonConfig) error {
 	config.SetIfNotDefault(jcfg.Folder, &cfg.Folder)
+	if v, ok := correctLoadingMode(jcfg.TableLoadingMode); ok {
+		cfg.BadgerOptions.TableLoadingMode = options.FileLoadingMode(v)
```
hsanjuan: casting `options.FileLoadingMode` twice. Also, shouldn't it error ?

hsanjuan: You can make `jsonConfig.TableLoadingMode` a pointer to decide whether it's set at all (with a possible wrong value) or not.

datastore/badger/config_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"testing"
+
+	"github.com/dgraph-io/badger"
+	"github.com/dgraph-io/badger/options"
+)
+
+var cfgJSON = []byte(`
+{
+    "folder": "test",
+	"value_log_loading_mode": 1
```
hsanjuan: indentation

go.mod
```diff
@@ -10,7 +10,8 @@ require (
 	github.com/boltdb/bolt v1.3.1 // indirect
 	github.com/btcsuite/btcd v0.0.0-20190427004231-96897255fd17 // indirect
 	github.com/coreos/go-semver v0.3.0 // indirect
-	github.com/dgryski/go-farm v0.0.0-20190423205320-6a90982ecee2 // indirect
+	github.com/dgraph-io/badger v2.0.0-rc.2+incompatible
+	github.com/dgryski/go-farm v0.0.0-20190416075124-e1214b5e05dc // indirect
```
hsanjuan: interesting, why is this one changing?

lanzafame:
> I thought we decided to try reducing the value log size so that it fits in memory, and this it not exposing that at all ?

I tried that before making these changes but even significantly reduced it didn't work.

hsanjuan:
> > I thought we decided to try reducing the value log size so that it fits in memory, and this it not exposing that at all ?
> 
> I tried that before making these changes but even significantly reduced it didn't work.

Can we expose it in any case, since we're at it?

go.mod
lanzafame: hmm not sure, I will check it out.

lanzafame:
> > > I thought we decided to try reducing the value log size so that it fits in memory, and this it not exposing that at all ?
> > 
> > 
> > I tried that before making these changes but even significantly reduced it didn't work.
> 
> Can we expose it in any case, since we're at it?

That was painful but it is now exposing the full config now.

hsanjuan:
agh I forgot to merge this

PR #739, opened 2019-04-04
Feature Support: cluster gc

0zAND1z:
This pull request is for a feature request #628 

**Tasks to be completed before merge:**

- [ ] Implement & review the API client for `/repo/gc`
- [ ] Implement  & review the RPC Call that needs to be executed in all the cluster nodes
- [ ] Write tests

---
This change is [Reviewable](https://reviewable.io/reviews/ipfs/ipfs-cluster/739)

PR #777, opened 2019-05-14
Add ability to run Garbage Collector(ipfs repo gc) on all peers

### LINKED ISSUE #628: Add ability to run Garbage Collector(ipfs repo gc) on all peers
Hi)

It will be very useful to have an ability run IPFS garbage collector through IPFS-Cluster API. 

https://discuss.ipfs.io/t/is-it-possible-to-completely-delete-file-from-the-cluster/4482/7

kishansagathiya:
Supersedes https://github.com/ipfs/ipfs-cluster/pull/739 

Fix #628

### DIFF
--- a/api/rest/client/client.go
+++ b/api/rest/client/client.go
@@ -119,6 +119,11 @@ type Client interface {
 	// Metrics returns a map with the latest metrics of matching name
 	// for the current cluster peers.
 	Metrics(ctx context.Context, name string) ([]*api.Metric, error)
+
+	// RepoGC runs garbage collection on IPFS daemons of cluster peers and
+	// returns collected CIDs. If local is true, it would garbage collect
+	// only on contacted peer, otherwise on all peers' IPFS daemons.
+	RepoGC(ctx context.Context, local bool) (*api.GlobalRepoGC, error)
 }
 
 // Config allows to configure the parameters to connect
--- a/api/rest/client/methods.go
+++ b/api/rest/client/methods.go
@@ -334,6 +334,26 @@ func (c *defaultClient) Metrics(ctx context.Context, name string) ([]*api.Metric
 	return metrics, err
 }
 
+// RepoGC runs garbage collection on IPFS daemons of cluster peers and
+// returns collected CIDs. If local is true, it would garbage collect
+// only on contacted peer, otherwise on all peers' IPFS daemons.
+func (c *defaultClient) RepoGC(ctx context.Context, local bool) (*api.GlobalRepoGC, error) {
+	ctx, span := trace.StartSpan(ctx, "client/RepoGC")
+	defer span.End()
+
+	var repoGC api.GlobalRepoGC
+	err := c.do(
+		ctx,
+		"POST",
+		fmt.Sprintf("/ipfs/gc?local=%t", local),
+		nil,
+		nil,
+		&repoGC,
+	)
+
+	return &repoGC, err
+}
+
 // WaitFor is a utility function that allows for a caller to wait for a
 // paticular status for a CID (as defined by StatusFilterParams).
 // It returns the final status for that CID and an error, if there was.
--- a/api/rest/restapi.go
+++ b/api/rest/restapi.go
@@ -457,6 +457,12 @@ func (api *API) routes() []route {
 			api.unpinPathHandler,
 		},
 		{
+			"RepoGC",
+			"POST",
+			"/ipfs/gc",
+			api.repoGCHandler,
+		},
+		{
 			"ConnectionGraph",
 			"GET",
 			"/health/graph",
@@ -1062,6 +1068,45 @@ func (api *API) recoverHandler(w http.ResponseWriter, r *http.Request) {
 			)
 			api.sendResponse(w, autoStatus, err, pinInfo)
 		}
+	}
+}
+
+func (api *API) repoGCHandler(w http.ResponseWriter, r *http.Request) {
+	queryValues := r.URL.Query()
+	local := queryValues.Get("local")
+
+	if local == "true" {
+		var localRepoGC types.RepoGC
+		err := api.rpcClient.CallContext(
+			r.Context(),
+			"",
+			"Cluster",
+			"RepoGCLocal",
+			struct{}{},
+			&localRepoGC,
+		)
+
+		api.sendResponse(w, autoStatus, err, repoGCToGlobal(&localRepoGC))
+		return
+	}
+
+	var repoGC types.GlobalRepoGC
+	err := api.rpcClient.CallContext(
+		r.Context(),
+		"",
+		"Cluster",
+		"RepoGC",
+		struct{}{},
+		&repoGC,
+	)
+	api.sendResponse(w, autoStatus, err, repoGC)
+}
+
+func repoGCToGlobal(r *types.RepoGC) types.GlobalRepoGC {
+	return types.GlobalRepoGC{
+		PeerMap: map[string]*types.RepoGC{
+			peer.IDB58Encode(r.Peer): r,
+		},
 	}
 }
 
--- a/api/types.go
+++ b/api/types.go
@@ -897,3 +897,23 @@ type IPFSRepoStat struct {
 	RepoSize   uint64 `codec:"r,omitempty"`
 	StorageMax uint64 `codec:"s, omitempty"`
 }
+
+// IPFSRepoGC represents the streaming response sent from repo gc API of IPFS.
+type IPFSRepoGC struct {
+	Key   cid.Cid `json:"key,omitempty" codec:"k,omitempty"`
+	Error string  `json:"error,omitempty" codec:"e,omitempty"`
+}
+
+// RepoGC contains garbage collected CIDs from a cluster peer's IPFS daemon.
+type RepoGC struct {
+	Peer     peer.ID      `json:"peer" codec:"p,omitempty"` // the Cluster peer ID
+	Peername string       `json:"peername" codec:"pn,omitempty"`
+	Keys     []IPFSRepoGC `json:"keys" codec:"k"`
+	Error    string       `json:"error,omitempty" codec:"e,omitempty"`
+}
+
+// GlobalRepoGC contains cluster-wide information about garbage collected CIDs
+// from IPFS.
+type GlobalRepoGC struct {
+	PeerMap map[string]*RepoGC `json:"peer_map" codec:"pm,omitempty"`
+}
--- a/cluster.go
+++ b/cluster.go
@@ -1883,3 +1883,73 @@ func diffPeers(peers1, peers2 []peer.ID) (added, removed []peer.ID) {
 	}
 	return
 }
+
+// RepoGC performs garbage collection sweep on all peers' IPFS repo.
+func (c *Cluster) RepoGC(ctx context.Context) (*api.GlobalRepoGC, error) {
+	_, span := trace.StartSpan(ctx, "cluster/RepoGC")
+	defer span.End()
+	ctx = trace.NewContext(c.ctx, span)
+
+	members, err := c.consensus.Peers(ctx)
+	if err != nil {
+		logger.Error(err)
+		return nil, err
+	}
+	lenMembers := len(members)
+
+	ctxs, cancels := rpcutil.CtxsWithCancel(ctx, lenMembers)
+	defer rpcutil.MultiCancel(cancels)
+
+	repoGCsResp := make([]*api.RepoGC, lenMembers, lenMembers)
+
+	errs := c.rpcClient.MultiCall(
+		ctxs,
+		members,
+		"Cluster",
+		"RepoGCLocal",
+		struct{}{},
+		rpcutil.CopyRepoGCSliceToIfaces(repoGCsResp),
+	)
+
+	// clubbing `RepoGCLocal` responses of all peers into one
+	globalRepoGC := api.GlobalRepoGC{PeerMap: make(map[string]*api.RepoGC)}
+	for i, resp := range repoGCsResp {
+		e := errs[i]
+
+		if e == nil {
+			globalRepoGC.PeerMap[peer.IDB58Encode(members[i])] = resp
+			continue
+		}
+
+		if rpc.IsAuthorizationError(e) {
+			logger.Debug("rpc auth error:", e)
+			continue
+		}
+
+		logger.Errorf("%s: error in broadcast response from %s: %s ", c.id, members[i], e)
+
+		globalRepoGC.PeerMap[peer.IDB58Encode(members[i])] = &api.RepoGC{
+			Peer:     members[i],
+			Peername: peer.IDB58Encode(members[i]),
+			Keys:     []api.IPFSRepoGC{},
+			Error:    err.Error(),
+		}
+	}
+
+	return &globalRepoGC, nil
+}
+
+// RepoGCLocal performs garbage collection only on the local IPFS deamon.
+func (c *Cluster) RepoGCLocal(ctx context.Context) (*api.RepoGC, error) {
+	_, span := trace.StartSpan(ctx, "cluster/RepoGCLocal")
+	defer span.End()
+	ctx = trace.NewContext(c.ctx, span)
+
+	resp, err := c.ipfs.RepoGC(ctx)
+	if err != nil {
+		return nil, err
+	}
+	resp.Peer = c.id
+	resp.Peername = c.config.Peername
+	return resp, nil
+}
--- a/cluster.go
+++ b/cluster.go
@@ -1897,19 +1897,21 @@ func (c *Cluster) RepoGC(ctx context.Context) (*api.GlobalRepoGC, error) {
 	}
 	lenMembers := len(members)
 
-	ctxs, cancels := rpcutil.CtxsWithCancel(ctx, lenMembers)
-	defer rpcutil.MultiCancel(cancels)
-
 	repoGCsResp := make([]*api.RepoGC, lenMembers, lenMembers)
+	errs := make([]error, lenMembers, lenMembers)
 
-	errs := c.rpcClient.MultiCall(
-		ctxs,
-		members,
-		"Cluster",
-		"RepoGCLocal",
-		struct{}{},
-		rpcutil.CopyRepoGCSliceToIfaces(repoGCsResp),
-	)
+	for i, member := range members {
+		var repoGC api.RepoGC
+		errs[i] = c.rpcClient.CallContext(
+			ctx,
+			member,
+			"Cluster",
+			"RepoGCLocal",
+			struct{}{},
+			&repoGC,
+		)
+		repoGCsResp[i] = &repoGC
+	}
 
 	// clubbing `RepoGCLocal` responses of all peers into one
 	globalRepoGC := api.GlobalRepoGC{PeerMap: make(map[string]*api.RepoGC)}
--- a/cluster.go
+++ b/cluster.go
@@ -1895,14 +1895,12 @@ func (c *Cluster) RepoGC(ctx context.Context) (*api.GlobalRepoGC, error) {
 		logger.Error(err)
 		return nil, err
 	}
-	lenMembers := len(members)
 
-	repoGCsResp := make([]*api.RepoGC, lenMembers, lenMembers)
-	errs := make([]error, lenMembers, lenMembers)
-
-	for i, member := range members {
+	// to club `RepoGCLocal` responses of all peers into one
+	globalRepoGC := api.GlobalRepoGC{PeerMap: make(map[string]*api.RepoGC)}
+	for _, member := range members {
 		var repoGC api.RepoGC
-		errs[i] = c.rpcClient.CallContext(
+		err = c.rpcClient.CallContext(
 			ctx,
 			member,
 			"Cluster",
@@ -1910,29 +1908,21 @@ func (c *Cluster) RepoGC(ctx context.Context) (*api.GlobalRepoGC, error) {
 			struct{}{},
 			&repoGC,
 		)
-		repoGCsResp[i] = &repoGC
-	}
-
-	// clubbing `RepoGCLocal` responses of all peers into one
-	globalRepoGC := api.GlobalRepoGC{PeerMap: make(map[string]*api.RepoGC)}
-	for i, resp := range repoGCsResp {
-		e := errs[i]
-
-		if e == nil {
-			globalRepoGC.PeerMap[peer.IDB58Encode(members[i])] = resp
+		if err == nil {
+			globalRepoGC.PeerMap[peer.IDB58Encode(member)] = &repoGC
 			continue
 		}
 
-		if rpc.IsAuthorizationError(e) {
-			logger.Debug("rpc auth error:", e)
+		if rpc.IsAuthorizationError(err) {
+			logger.Debug("rpc auth error:", err)
 			continue
 		}
 
-		logger.Errorf("%s: error in broadcast response from %s: %s ", c.id, members[i], e)
+		logger.Errorf("%s: error in broadcast response from %s: %s ", c.id, member, err)
 
-		globalRepoGC.PeerMap[peer.IDB58Encode(members[i])] = &api.RepoGC{
-			Peer:     members[i],
-			Peername: peer.IDB58Encode(members[i]),
+		globalRepoGC.PeerMap[peer.IDB58Encode(member)] = &api.RepoGC{
+			Peer:     member,
+			Peername: peer.IDB58Encode(member),
 			Keys:     []api.IPFSRepoGC{},
 			Error:    err.Error(),
 		}
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -109,6 +109,16 @@ func (ipfs *mockConnector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, err
 	return &api.IPFSRepoStat{RepoSize: 100, StorageMax: 1000}, nil
 }
 
+func (ipfs *mockConnector) RepoGC(ctx context.Context) (*api.RepoGC, error) {
+	return &api.RepoGC{
+		Keys: []api.IPFSRepoGC{
+			{
+				Key: test.Cid1,
+			},
+		},
+	}, nil
+}
+
 func (ipfs *mockConnector) Resolve(ctx context.Context, path string) (cid.Cid, error) {
 	_, err := gopath.ParsePath(path)
 	if err != nil {
@@ -849,7 +859,7 @@ func TestClusterUnpinPath(t *testing.T) {
 	// Unpin after pin should succeed
 	pin, err := cl.PinPath(ctx, test.PathIPFS2, api.PinOptions{})
 	if err != nil {
-		t.Fatal("pin with should have worked:", err)
+		t.Fatal("pin with path should have worked:", err)
 	}
 	if !pin.Cid.Equals(test.CidResolved) {
 		t.Error("expected a different cid, found", pin.Cid.String())
@@ -919,5 +929,64 @@ func TestClusterRecoverAllLocal(t *testing.T) {
 	}
 	if recov[0].Status != api.TrackerStatusPinned {
 		t.Errorf("the pin should have been recovered, got = %v", recov[0].Status)
+	}
+}
+
+func TestClusterRepoGC(t *testing.T) {
+	ctx := context.Background()
+	cl, _, _, _ := testingCluster(t)
+	defer cleanState()
+	defer cl.Shutdown(ctx)
+
+	gRepoGC, err := cl.RepoGC(ctx)
+	if err != nil {
+		t.Fatal("gc should have worked:", err)
+	}
+
+	if gRepoGC.PeerMap == nil {
+		t.Fatal("expected a non-nil peer map")
+	}
+
+	if len(gRepoGC.PeerMap) != 1 {
+		t.Error("expected repo gc information for one peer")
+	}
+	for _, repoGC := range gRepoGC.PeerMap {
+		testRepoGC(t, repoGC)
+	}
+
+}
+
+func TestClusterRepoGCLocal(t *testing.T) {
+	ctx := context.Background()
+	cl, _, _, _ := testingCluster(t)
+	defer cleanState()
+	defer cl.Shutdown(ctx)
+
+	repoGC, err := cl.RepoGCLocal(ctx)
+	if err != nil {
+		t.Fatal("gc should have worked:", err)
+	}
+
+	testRepoGC(t, repoGC)
+}
+
+func testRepoGC(t *testing.T, repoGC *api.RepoGC) {
+	if repoGC.Peer == "" {
+		t.Error("expected a cluster ID")
+	}
+	if repoGC.Error != "" {
+		t.Error("did not expect any error")
+	}
+
+	if repoGC.Keys == nil {
+		t.Fatal("expected a non-nil array of IPFSRepoGC")
+	}
+
+	if len(repoGC.Keys) == 0 {
+		t.Fatal("expected at least one key, but found none")
+	}
+
+	if !repoGC.Keys[0].Key.Equals(test.Cid1) {
+		t.Errorf("expected a different cid, expected: %s, found: %s", test.Cid1, repoGC.Keys[0].Key)
 	}
 }
--- a/cmd/ipfs-cluster-ctl/formatters.go
+++ b/cmd/ipfs-cluster-ctl/formatters.go
@@ -86,6 +86,8 @@ func textFormatObject(resp interface{}) {
 		for _, item := range resp.([]*api.Metric) {
 			textFormatObject(item)
 		}
+	case *api.GlobalRepoGC:
+		textFormatPrintGlobalRepoGC(resp.(*api.GlobalRepoGC))
 	default:
 		checkErr("", errors.New("unsupported type returned"))
 	}
@@ -213,6 +215,38 @@ func textFormatPrintAddedOutputQuiet(obj *addedOutputQuiet) {
 func textFormatPrintMetric(obj *api.Metric) {
 	date := time.Unix(0, obj.Expire).UTC().Format(time.RFC3339)
 	fmt.Printf("%s: %s | Expire: %s\n", peer.IDB58Encode(obj.Peer), obj.Value, date)
+}
+
+func textFormatPrintGlobalRepoGC(obj *api.GlobalRepoGC) {
+	peers := make(sort.StringSlice, 0, len(obj.PeerMap))
+	for peer := range obj.PeerMap {
+		peers = append(peers, peer)
+	}
+	peers.Sort()
+
+	for _, peer := range peers {
+		item := obj.PeerMap[peer]
+		// If peer name is set, use it instead of peer ID.
+		if len(item.Peername) > 0 {
+			peer = item.Peername
+		}
+		if item.Error != "" {
+			fmt.Printf("%-15s | ERROR: %s\n", peer, item.Error)
+		} else {
+			fmt.Printf("%-15s\n", peer)
+		}
+
+		fmt.Printf("  > CIDs:\n")
+		for _, key := range item.Keys {
+			if key.Error != "" {
+				// key.Key will be empty
+				fmt.Printf("    - ERROR: %s\n", key.Error)
+				continue
+			}
+
+			fmt.Printf("    - %s\n", key.Key)
+		}
+	}
 }
 
 func textFormatPrintError(obj *api.Error) {
--- a/cmd/ipfs-cluster-ctl/main.go
+++ b/cmd/ipfs-cluster-ctl/main.go
@@ -953,6 +953,32 @@ but usually are:
 			},
 		},
 		{
+			Name:        "ipfs",
+			Usage:       "Manage IPFS daemon",
+			Description: "Manage IPFS daemon",
+			Subcommands: []cli.Command{
+				{
+					Name:  "gc",
+					Usage: "run garbage collection on IPFS repos of cluster peers",
+					Description: `
+This command will instruct current Cluster peers to run "repo gc" on their
+respective IPFS daemons.
+
+When --local flag is passed, it will garbage collect only on the local IPFS
+deamon, otherwise on all IPFS daemons.
+`,
+					Flags: []cli.Flag{
+						localFlag(),
+					},
+					Action: func(c *cli.Context) error {
+						resp, cerr := globalClient.RepoGC(ctx, c.Bool("local"))
+						formatResponse(c, resp, cerr)
+						return nil
+					},
+				},
+			},
+		},
+		{
 			Name:      "commands",
 			Usage:     "List all commands",
 			ArgsUsage: " ",
--- a/ipfsconn/ipfshttp/config_test.go
+++ b/ipfsconn/ipfshttp/config_test.go
@@ -9,11 +9,12 @@ import (
 
 var cfgJSON = []byte(`
 {
-      "node_multiaddress": "/ip4/127.0.0.1/tcp/5001",
-      "connect_swarms_delay": "7s",
-      "ipfs_request_timeout": "5m0s",
-      "pin_timeout": "24h",
-      "unpin_timeout": "3h"
+	"node_multiaddress": "/ip4/127.0.0.1/tcp/5001",
+	"connect_swarms_delay": "7s",
+	"ipfs_request_timeout": "5m0s",
+	"pin_timeout": "24h",
+	"unpin_timeout": "3h",
+	"repogc_timeout": "24h"
 }
 `)
 
--- a/ipfsconn/ipfshttp/ipfshttp.go
+++ b/ipfsconn/ipfshttp/ipfshttp.go
@@ -93,6 +93,11 @@ type ipfsResolveResp struct {
 	Path string
 }
 
+type ipfsRepoGCResp struct {
+	Key   cid.Cid
+	Error string
+}
+
 type ipfsRefsResp struct {
 	Ref string
 	Err string
@@ -755,6 +760,47 @@ func (ipfs *Connector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, error) 
 		return nil, err
 	}
 	return &stats, nil
+}
+
+// RepoGC performs a garbage collection sweep on the cluster peer's IPFS repo.
+func (ipfs *Connector) RepoGC(ctx context.Context) (*api.RepoGC, error) {
+	ctx, span := trace.StartSpan(ctx, "ipfsconn/ipfshttp/RepoGC")
+	defer span.End()
+
+	ctx, cancel := context.WithTimeout(ctx, ipfs.config.RepoGCTimeout)
+	defer cancel()
+
+	res, err := ipfs.doPostCtx(ctx, ipfs.client, ipfs.apiURL(), "repo/gc?stream-errors=true", "", nil)
+	if err != nil {
+		logger.Error(err)
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	dec := json.NewDecoder(res.Body)
+	repoGC := &api.RepoGC{
+		Keys: []api.IPFSRepoGC{},
+	}
+	for {
+		resp := ipfsRepoGCResp{}
+
+		if err := dec.Decode(&resp); err != nil {
+			// If we cancelled the request we should tell the user
+			// (in case dec.Decode() exited cleanly with an EOF).
+			select {
+			case <-ctx.Done():
+				return repoGC, ctx.Err()
+			default:
+				if err == io.EOF {
+					return repoGC, nil // clean exit
+				}
+				logger.Error(err)
+				return repoGC, err // error decoding
+			}
+		}
+
+		repoGC.Keys = append(repoGC.Keys, api.IPFSRepoGC{Key: resp.Key, Error: resp.Error})
+	}
 }
 
 // Resolve accepts ipfs or ipns path and resolves it into a cid
--- a/ipfsconn/ipfshttp/ipfshttp_test.go
+++ b/ipfsconn/ipfshttp/ipfshttp_test.go
@@ -10,6 +10,7 @@ import (
 	logging "github.com/ipfs/go-log"
 	ma "github.com/multiformats/go-multiaddr"
 
+	merkledag "github.com/ipfs/go-merkledag"
 	"github.com/ipfs/ipfs-cluster/api"
 	"github.com/ipfs/ipfs-cluster/test"
 )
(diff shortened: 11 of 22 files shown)


ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -101,6 +101,13 @@ type ipfsStream struct {
 	Protocol string
 }
 
+// ipfsRepoGC contains `ipfs repo gc` api response for a single garbace
+// collected content.
+type ipfsRepoGC struct {
+	Key   cid.Cid `json:"Key"`
```
lanzafame: Decapitalise the json struct tags

cluster.go
```diff
@@ -1597,3 +1597,41 @@ func diffPeers(peers1, peers2 []peer.ID) (added, removed []peer.ID) {
 	}
 	return
 }
+
+// RepoGC performs garbage collection sweep on all peers' repo.
+func (c *Cluster) RepoGC(ctx context.Context) error {
```
lanzafame: Can we get an endpoint that allows for individual gc? We can then trigger gc on particular peers based metrics coming in

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: IPFS uses capitalized.

hsanjuan: Maybe, simply don't put a json tag when it does nothing.

cluster.go
hsanjuan: > Can we get an endpoint that allows for individual gc? We can then trigger gc on particular peers based metrics coming in

You mean, a public endpoint? Internally this should be no problem (talking directly to the IPFSConnetor).

kishansagathiya:
ipfs repo gc api has streaming output, wonder if we should do the same with our repo gc api?

hsanjuan:
> ipfs repo gc api has streaming output, wonder if we should do the same with our repo gc api?

you can't do streaming without streaming rpc which we don't have for the moment. Either we wait (a potentially long time for everyone to finish, making sure the requests do not time out (in a similar fashion to the pin PR, looking at the progress), or we return with 202 Accepted and leave it running in the background. I prefer the first option.

kishansagathiya:
> Either we wait (a potentially long time for everyone to finish, making sure the requests do not time out 

With that repo gc api of cluster's ipfs proxy would be different than that of ipfs api (since no streaming response)

>  streaming rpc which we don't have for the moment.

How doable is this?

hsanjuan:
> > Either we wait (a potentially long time for everyone to finish, making sure the requests do not time out
> 
> With that repo gc api of cluster's ipfs proxy would be different than that of ipfs api (since no streaming response)

we'll have to live with that for the moment

> 
> > streaming rpc which we don't have for the moment.
> 
> How doable is this?

not doable now, difficult

api/rest/client/client.go
```diff
@@ -119,6 +119,8 @@ type Client interface {
 	// Metrics returns a map with the latest metrics of matching name
 	// for the current cluster peers.
 	Metrics(ctx context.Context, name string) ([]*api.Metric, error)
+
```
lanzafame: Interface methods should have a comment.

api/types.go
```diff
@@ -838,3 +838,10 @@ type IPFSRepoStat struct {
 	RepoSize   uint64 `codec:"r,omitempty"`
 	StorageMax uint64 `codec:"s, omitempty"`
 }
+
+// IPFSRepoGC contains `ipfs repo gc` api response for a single garbace
```
lanzafame:
```suggestion
// IPFSRepoGC contains `ipfs repo gc` api response for a single garbage
```

kishansagathiya:
```
✗ ipfs-cluster-ctl --encoding json ipfs gc 
[
    {
        "Key": {
            "/": "QmeH3GhJf3Hj6o3K8Wvj4wAWS4gsiuopoSsx38711jpLt5"
        }
    },
    {
        "Key": {
            "/": "QmU3UbpCLLWftCvnXRQkhgK1aEFwnsqwtE6pKKR64ApiQS"
        }
    },
    {
        "Key": {
            "/": "QmUB8H5k465qDQjgubGhnjasVy513xLo8EW8pyZ5xeymie"
        }
    },
    {
        "Key": {
            "/": "QmZZRTyhDpL5Jgift1cHbAhexeE1m2Hw8x8g7rTcPahDvo"
        }
    },
    {
        "Key": {
            "/": "QmPhk6cJkRcFfZCdYam4c9MKYjFG9V29LswUnbrFNhtk2S"
        }
    }
]
✗ ipfs-cluster-ctl ipfs gc                             
QmU3UbpCLLWftCvnXRQkhgK1aEFwnsqwtE6pKKR64ApiQS | Error : 
✗ 
```

hsanjuan:
> ```
> ✗ ipfs-cluster-ctl --encoding json ipfs gc 
> [
>     {
>         "Key": {
>             "/": "QmeH3GhJf3Hj6o3K8Wvj4wAWS4gsiuopoSsx38711jpLt5"
>         }
>     },
>     {
>         "Key": {
>             "/": "QmU3UbpCLLWftCvnXRQkhgK1aEFwnsqwtE6pKKR64ApiQS"
>         }
>     },
>     {
>         "Key": {
>             "/": "QmUB8H5k465qDQjgubGhnjasVy513xLo8EW8pyZ5xeymie"
>         }
>     },
>     {
>         "Key": {
>             "/": "QmZZRTyhDpL5Jgift1cHbAhexeE1m2Hw8x8g7rTcPahDvo"
>         }
>     },
>     {
>         "Key": {
>             "/": "QmPhk6cJkRcFfZCdYam4c9MKYjFG9V29LswUnbrFNhtk2S"
>         }
>     }
> ]
> ✗ ipfs-cluster-ctl ipfs gc                             
> QmU3UbpCLLWftCvnXRQkhgK1aEFwnsqwtE6pKKR64ApiQS | Error : 
> ✗ 
> ```

I'm not sure what this means (I haven't read the PR) but the Cluster REST API never returns uppercased keys.

api/rest/client/methods.go
```diff
@@ -325,6 +325,25 @@ func (c *defaultClient) Metrics(ctx context.Context, name string) ([]*api.Metric
...
+// collected cids.
+func (c *defaultClient) RepoGC(ctx context.Context) ([]api.IPFSRepoGC, error) {
+	ctx, span := trace.StartSpan(ctx, "client/RepoGC")
+	defer span.End()
+
+	var repoGC []api.IPFSRepoGC
+	err := c.do(
+		ctx,
+		"POST",
+		"/repo/gc",
```
hsanjuan: `/peers/gc` or `peers/ipfs/gc`.

api/rest/restapi.go
```diff
@@ -421,6 +421,12 @@ func (api *API) routes() []route {
 			"/pins/{keyType:ipfs|ipns|ipld}/{path:.*}",
 			api.unpinPathHandler,
 		},
+		{
+			"RepoGC",
+			"POST",
+			"/repo/gc",
```
hsanjuan: `/peers/gc or peers/ipfs/gc.` (move up next to the `peers` endpoints).

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -208,6 +214,10 @@ func textFormatPrintMetric(obj *api.Metric) {
 	fmt.Printf("%s: %s | Expire: %s\n", peer.IDB58Encode(obj.Peer), obj.Value, date)
 }
 
+func textFormatPrintRepoGC(obj api.IPFSRepoGC) {
+	fmt.Printf("%s | Error : %s\n", obj.Key.String(), obj.Error)
```
hsanjuan: Don't print an error tag when there is no error. Look at other places with types containing errors. Keep consistency.

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -860,6 +860,27 @@ but usually are:
...
+		{
+			Name:        "ipfs",
+			Usage:       "Manage IPFS daemon",
+			Description: "Manage IPFS daemon",
+			Subcommands: []cli.Command{
+				{
+					Name:  "gc",
+					Usage: "run garbage collection on ipfs repos of cluster peers",
+					Description: `
+This command would sweep the local set of stored objects and remove ones that
```
hsanjuan: `This command will instruct all current Cluster peers to run "repo gc" on their respective IPFS daemons`.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -604,6 +604,39 @@ func (ipfs *Connector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, error)
 	return &stats, nil
 }
 
+// RepoGC performs a garbage collection sweep on the repo.
+func (ipfs *Connector) RepoGC(ctx context.Context) ([]api.IPFSRepoGC, error) {
+	ctx, span := trace.StartSpan(ctx, "ipfsconn/ipfshttp/RepoGC")
+	defer span.End()
+
+	ctx, cancel := context.WithTimeout(ctx, ipfs.config.IPFSRequestTimeout)
```
hsanjuan: repo GC is going to need a special timeout because it's an operation that can take hours.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -604,6 +604,39 @@ func (ipfs *Connector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, error)
...
+	res, err := ipfs.doPostCtx(ctx, ipfs.client, ipfs.apiURL(), "repo/gc", "", nil)
+	if err != nil {
+		logger.Error(err)
+		return repoGCResp, err
+	}
+	defer res.Body.Close()
+
+	dec := json.NewDecoder(res.Body)
+	for {
+		resp := api.IPFSRepoGC{}
```
hsanjuan: cluster API and IPFS api will certainly use different types. This should be a type defined here that matches what the IPFS api returns, like every other method does in this module.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -604,6 +604,39 @@ func (ipfs *Connector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, error)
...
+		if err := dec.Decode(&resp); err == io.EOF {
+			break
+		} else if err != nil {
+			return repoGCResp, err
+		}
+		if resp.Error != "" {
+			logger.Error("Error while repo gc: ", resp.Error)
+		}
+		repoGCResp = append(repoGCResp, resp)
```
hsanjuan: Can you investigate what the --quiet flag does, and if setting `stream-channels=false` has any effect on the output (actually making return an array instead of streaming).

kishansagathiya: None of these options give an array instead of streaming. Strangely, for `quiet` true or false doesn't make a difference. It gives the same results. 
`stream-errors` is for whether I want to see errors in my output or not, but those are also all streaming responses.

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: We can do something similar to what we are doing for pinTimeout.

api/rest/restapi.go
kishansagathiya: Didn't we agree on not linking this with peers and using `ipfs gc` https://github.com/ipfs/ipfs-cluster/pull/739#discussion_r278049953

It's nice to have your commands and APIs matching. We have done that so far. If we are using command as `ctl ipfs gc` the API should be probably be `/ipfs/gc`

hsanjuan: I also said that I did not like `/repo/gc`...

Thing is, `/ipfs/` is not a resource type (yet). A peer is a resource, and `POST /peers/ipfs` would indicate that we want to run an `/ipfs` action on our peers. That's my logic.

But `/ipfs/gc` might also be ok if we consider `/ipfs` will at some point be a resource (for example to return IPFSID objects directly), although IPFS ID objects are more a sub-resource of peers. I think this is slightly more human-friendly though.

Btw, it's not true commands match API endpoints. Pin add/rm talk to `/allocations` while `status` talks to `/pins`. Traditionally commands strived to keep familiarity while API strived to be consistent with data types and objects in cluster.

ipfsconn/ipfshttp/ipfshttp.go
hsanjuan: if you ever run repo GC, you will observe that there is a part where no output is produced. That parts corresponds to walking down all the pinned DAGs to mark them so they are not GC'ed. Only when that set is final, GC starts and the call produces output. The first part, without any output, can take hours.

hsanjuan: After that output is going to come out very fast very probably, meaning we can just go the simple way and read it.

ipfsconn/ipfshttp/ipfshttp.go
hsanjuan: hmm ok

kishansagathiya:
> type GlobalRepoGC struct {
    PeerMap map[string]*RepoGC
}

This repeats peerID (assuming that string here would be string of peerID).
How about 
```
type GlobalRepoGC struct {
    PeerMap map[peer.ID]*RepoGC
}
```
and RepoGC stays
```
type RepoGC struct {
   Keys []cid.Cid
   Error string
}
```

then the cluster method would be
```
func (c *Cluster) RepoGC(ctx context.Context) (GlobalRepoGC, error)
```

api/rest/restapi.go
kishansagathiya: I have changed this to `/ipfs/gc`. Is that fine?

kishansagathiya:
What output looks like
```
➜✗ ipfs-cluster-ctl ipfs gc
QmR4EQTUuwpcJAgZFPzT4W1NqBrnzx82VVJhWr6cYiuWEz
  > CIDs:
    - QmZZRTyhDpL5Jgift1cHbAhexeE1m2Hw8x8g7rTcPahDvo
    - QmVRQDVCjSY6Tp4U1H41DprKaKgs7NMmBt1Rg6eFoABNVg
    - QmPhk6cJkRcFfZCdYam4c9MKYjFG9V29LswUnbrFNhtk2S
QmXGEuS7ARYwnBCp86uKi5oGHMg8b1eefmrKnQuvcEWboW
  > CIDs:
    - QmZZRTyhDpL5Jgift1cHbAhexeE1m2Hw8x8g7rTcPahDvo
    - QmVRQDVCjSY6Tp4U1H41DprKaKgs7NMmBt1Rg6eFoABNVg
    - QmPhk6cJkRcFfZCdYam4c9MKYjFG9V29LswUnbrFNhtk2S
➜✗ 
```

hsanjuan:
> > type GlobalRepoGC struct {
> > PeerMap map[string]*RepoGC
> > }
> 
> This repeats peerID (assuming that string here would be string of peerID).
> How about
> 
> ```
> type GlobalRepoGC struct {
>     PeerMap map[string]*RepoGC
> }
> ```
> 
> and RepoGC stays
> 
> ```
> type RepoGC struct {
>    Keys []cid.Cid
>    Error string
> }
> ```
> 
> then the cluster method would be
> 
> ```
> func (c *Cluster) RepoGC(ctx context.Context) (GlobalRepoGC, error)
> ```

I think the purpose of including a peer.ID in the inner object is that it facilitates calls with `--local` to turn the local status object into the global one. It also facilitates doing some quick parsings. Given that each peer only returns one RepoGC element I'm not worried about the overhead (way worse in the case of GlobalPinInfos).

cluster.go
```diff
@@ -1737,3 +1737,56 @@ func diffPeers(peers1, peers2 []peer.ID) (added, removed []peer.ID) {
...
+	errs := c.rpcClient.MultiCall(
+		ctxs,
+		members,
+		"IPFSConnector",
+		"RepoGC",
+		struct{}{},
+		rpcutil.CopyRepoGCSliceToIfaces(repoGCsResp),
+	)
+
+	if len(errs) > 0 {
```
hsanjuan: This has a number of problems. Please copy response handling from `globalPinInfoCid` (for example).

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -208,6 +212,23 @@ func textFormatPrintMetric(obj *api.Metric) {
 	fmt.Printf("%s: %s | Expire: %s\n", peer.IDB58Encode(obj.Peer), obj.Value, date)
 }
 
+func textFormatPrintRepoGC(obj *api.RepoGC) {
+	if obj.Error != "" {
+		fmt.Printf("  > Error: %s\n", obj.Error)
```
hsanjuan: should return?

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -208,6 +212,23 @@ func textFormatPrintMetric(obj *api.Metric) {
 	fmt.Printf("%s: %s | Expire: %s\n", peer.IDB58Encode(obj.Peer), obj.Value, date)
 }
 
+func textFormatPrintRepoGC(obj *api.RepoGC) {
```
hsanjuan: If we add `--local` support we will have to change this. for PinInfo the `textFormatPrintPInfo` method creates a GlobalPinInfo object and calls the formatter for that one, so that output is always consistent. You could copy that (and also the peer sorting and could also carry peer names in the object).

go.sum
```diff
@@ -74,6 +74,7 @@ github.com/davidlazar/go-crypto v0.0.0-20170701192655-dcfb0a7ac018 h1:6xT9KW8zLC
 github.com/davidlazar/go-crypto v0.0.0-20170701192655-dcfb0a7ac018/go.mod h1:rQYf4tfk5sSwFsnDg3qYaBxSjsD9S8+59vW0dKUgme4=
 github.com/dgraph-io/badger v1.5.5-0.20190226225317-8115aed38f8f/go.mod h1:VZxzAIRPHRVNRKRo6AXrX9BJegn6il06VMTZVJYCIjQ=
 github.com/dgraph-io/badger/v2 v2.0.0-20190620211019-41d170b5158f/go.mod h1:jUaIjOV835xZ/mCLG/8P/38ZxiT4bG/K1khDNZJxuwU=
+github.com/dgraph-io/badger/v2 v2.0.0-rc2 h1:Lxx0mkjYCrqiqIp75+UOwN3pwRChXZyLQMz+oidMEMs=
```
hsanjuan: should not happen / keep current modules from master.

ipfsconn/ipfshttp/ipfshttp.go
hsanjuan: a special timeout setting for is still missing

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -612,6 +617,48 @@ func (ipfs *Connector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, error)
...
+	ctx, span := trace.StartSpan(ctx, "ipfsconn/ipfshttp/RepoGC")
+	defer span.End()
+
+	id, err := ipfs.ID(ctx)
+	if err != nil {
+		logger.Error(err)
+		return nil, err
+	}
+	repoGC := api.RepoGC{
+		Peer: id.ID,
```
hsanjuan: I thought we'd be setting this to the Cluster peer ID (in `cluster.go`). If you want to carry the ipfs id, you'll need an additional field in the `RepoGC`.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -612,6 +617,48 @@ func (ipfs *Connector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, error)
...
+	if err != nil {
+		logger.Error(err)
+		return &repoGC, err
+	}
+	defer res.Body.Close()
+
+	dec := json.NewDecoder(res.Body)
+	for {
+		resp := ipfsRepoGCResp{}
+		if err := dec.Decode(&resp); err == io.EOF {
```
hsanjuan: see https://github.com/ipfs/ipfs-cluster/blob/master/ipfsconn/ipfshttp/ipfshttp.go#L413 for a slightly improved error handling.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -612,6 +617,48 @@ func (ipfs *Connector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, error)
...
+			break
+		} else if err != nil {
+			return &repoGC, err
+		}
+		if resp.Error != "" {
+			logger.Error("Error while repo gc: ", resp.Error)
+			repoGC.Error += ", " + resp.Error
```
hsanjuan: put errors in a slice and join them later (by tagging them with the cid they belong to if present `<cid>: <error>, <cid2>: <error2>`? This will result in `, error 1`.

rpc_policy.go
```diff
@@ -34,33 +35,34 @@ var DefaultRPCPolicy = map[string]RPCEndpointType{
...
 
-	// IPFSConnector methods
+	//  IPFSConnector methods
 	"IPFSConnector.BlockGet":   RPCClosed,
 	"IPFSConnector.BlockPut":   RPCTrusted, // Called from Add()
 	"IPFSConnector.ConfigKey":  RPCClosed,
 	"IPFSConnector.Pin":        RPCClosed,
 	"IPFSConnector.PinLs":      RPCClosed,
 	"IPFSConnector.PinLsCid":   RPCClosed,
+	"IPFSConnector.RepoGC":     RPCTrusted,
```
hsanjuan: I don't think we want trusted peers to be allowed to RepoGC other peers even if they are trusted. We can leave this as RPC closed. Please add comment (in `comments` variable in policygen.go)

kishansagathiya: For Cluster.RepoGC to work it has to broadcast IPFSConnector.RepoGC. If this can't happen, there won't be cluster repoGC, only local one.

cmd/ipfs-cluster-ctl/formatters.go
kishansagathiya: No, it should just print errors saved in api.RepoGC to the terminal

rpc_policy.go
hsanjuan: Hmm I don't feel conformtable opening this up on follower-clusters because of the potential impact of a trusted peers mistakenly gc-ing a bunch of people. We should come with a way that users can opt-in to certain permissions but I guess it needs to stay as Trusted for the moment.

lanzafame: @hsanjuan do we have a place to document the current edges cases with the different cluster ‘modes’ like the one you just mentioned? I fear it will get lost in a PR and github search is sketchy at the best of times.

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: Have changed the struct for repo gc. With this we would be able to tag cids with errors.
```
// IPFSRepoGC represents the streaming response sent from ipfs' repo gc API.
type IPFSRepoGC struct {
	Key   cid.Cid `json:"key" codec:"k"`
	Error string  `json:"error,omitempty" codec:"e,omitempty"`
}

// RepoGC contains gc'ed CIDs from a cluster peer's IPFS daemon.
type RepoGC struct {
	Peer  peer.ID      `json:"peer" codec:"p,omitempty"` // the Cluster peer ID
	Keys  []IPFSRepoGC `json:"keys" codec:"k"`
	Error string       `json:"error,omitempty" codec:"e,omitempty"`
}
```

cmd/ipfs-cluster-ctl/formatters.go
kishansagathiya: added support for `--local`

Instead of converting an object into a global object in formatter, shouldn't this happen at the API level? That would make API output consistent as well.

kishansagathiya: Added peer name and sorted objects by peers.

But if this is about delivering same results(Is it?), with formatting we have done only for cli. 
Results in APIs are not sorted. Should we use array instead of map, if aim is to achieve consistent output?

api/types.go
```diff
@@ -863,3 +863,22 @@ type IPFSRepoStat struct {
 	RepoSize   uint64 `codec:"r,omitempty"`
 	StorageMax uint64 `codec:"s, omitempty"`
 }
+
+// IPFSRepoGC represents the streaming response sent from ipfs' repo gc API.
+type IPFSRepoGC struct {
+	Key   cid.Cid `json:"key" codec:"k"`
```
hsanjuan: you can `omitempty` I think.

cluster.go
```diff
@@ -1741,3 +1741,42 @@ func diffPeers(peers1, peers2 []peer.ID) (added, removed []peer.ID) {
...
+		members,
+		"IPFSConnector",
+		"RepoGC",
+		struct{}{},
+		rpcutil.CopyRepoGCSliceToIfaces(repoGCsResp),
+	)
+
+	// clubbing repo gc responses of all peers into one
+	globalRepoGC := api.GlobalRepoGC{PeerMap: make(map[string]*api.RepoGC)}
+	for i := 0; i < len(members); i++ {
```
hsanjuan: I asked to copy handling from `globalPinInfoCid` but you seem to have disregarded, why?

You missed ignoring auth errors as `globalPinInfoCid` does. Also why doing the loop like `for i := 0; i < len(members); i++ {` when you could `for i, resp := range repoGCResp[i] {` like the other function. It's more readable (you are processing responses not members). There is value in keeping similar things similar.

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -86,6 +86,8 @@ func textFormatObject(resp interface{}) {
 		for _, item := range resp.([]*api.Metric) {
 			textFormatObject(item)
 		}
+	case *api.GlobalRepoGC:
+		textFormatPrintGlobaleRepoGC(resp.(*api.GlobalRepoGC))
```
hsanjuan: There is a typo `GlobaleRepoGC` -> `GlobalRepoGC`.

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -208,6 +210,37 @@ func textFormatPrintMetric(obj *api.Metric) {
...
+		if item.Error != "" {
+			fmt.Printf("%-15s | ERROR: %s\n", peer, item.Error)
+		} else {
+			fmt.Printf("%-15s\n", peer)
+		}
+
+		fmt.Printf("  > CIDs:\n")
+		for _, key := range item.Keys {
+			if key.Error != "" {
+				fmt.Printf("    - %s | ERROR: %s\n", key.Key.String(), key.Error)
```
hsanjuan: Unfortunately ipfs will not set the `Key` when an error happens, it will make it part of the error message (next time please look things like this up before deciding on an approach).

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -208,6 +210,37 @@ func textFormatPrintMetric(obj *api.Metric) {
...
+				continue
+			}
+
+			fmt.Printf("    - %s\n", key.Key.String())
```
hsanjuan: The `%s` formatter automatically calls the `String()` method for you (applies to multiple places)

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -860,6 +860,30 @@ but usually are:
...
+		{
+			Name:        "ipfs",
+			Usage:       "Manage IPFS daemon",
+			Description: "Manage IPFS daemon",
+			Subcommands: []cli.Command{
+				{
+					Name:  "gc",
+					Usage: "run garbage collection on ipfs repos of cluster peers",
+					Description: `
+This command will instruct current Cluster peers to run "repo gc" on their respective IPFS daemons.
```
hsanjuan: We try to wrap most descriptions on 80 char-per-line. These are longer.

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -860,6 +860,30 @@ but usually are:
...
+
+When --local flag is passed, it would garbage collect only on the local peer, otherwise on all peers.
```
hsanjuan: it ***will*** only on the local ipfs ***daemon***... on all ***daemons***.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -806,6 +811,103 @@ func (ipfs *Connector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, error)
 	return &stats, nil
 }
 
+// RepoGC performs a garbage collection sweep on the cluster peer's IPFS repo.
+func (ipfs *Connector) RepoGC(ctx context.Context) (*api.RepoGC, error) {
+	ctx, span := trace.StartSpan(ctx, "ipfsconn/ipfshttp/RepoGC")
+	defer span.End()
+
+	ctx1, cancel1 := context.WithTimeout(ctx, ipfs.config.IPFSRequestTimeout)
```
hsanjuan: Sorry if it was not clear before, this needs a special timeout, by that I mean a new IPFSGCTimeout option in the config..

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -806,6 +811,103 @@ func (ipfs *Connector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, error)
...
+	defer cancel1()
+
+	id := &api.ID{}
+	err := ipfs.rpcClient.CallContext(
```
hsanjuan: You can fill the peerID/peerName information of this object in `cluster.go` when you process the answers of the broadcast (like `globalPinInfoCid` does). Slightly simpler, but mostly for consistency.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -806,6 +811,103 @@ func (ipfs *Connector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, error)
...
+		}
+	}()
+
+	keys, err := ipfs.repoGCProgress(ctx, outGCs)
+	repoGC.Keys = keys
+
+	return &repoGC, err
+}
+
+func (ipfs *Connector) repoGCProgress(ctx context.Context, out chan<- string) ([]api.IPFSRepoGC, error) {
```
hsanjuan: Why did you add this? Our previous convo about exactly this here:

image

In the last review I did not raise any comments about this part needing a full rework. Adopting this approach here does not help because the GC output is not likely to hang long once you start getting it. Most of the wait will happen with no output at all. This overcomplicates. Revert to the previous form please.

Please next time, concentrate on resolving review comments. Do not move forward into reworking full parts without asking first or being asked to.

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -213,18 +213,21 @@ func textFormatPrintMetric(obj *api.Metric) {
...
 func textFormatPrintRepoGC(obj *api.RepoGC) {
-	if obj.Error != "" {
-		fmt.Printf("  > Error: %s\n", obj.Error)
-	}
-
 	fmt.Printf("  > CIDs:\n")
 	for _, key := range obj.Keys {
-		fmt.Printf("    - %s\n", key.String())
+		if key.Error != "" {
+			fmt.Printf("    - %s | ERROR: %s\n", key.Key.String(), key.Error)
```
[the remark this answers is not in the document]
kishansagathiya: On going through related IPFS code to see how this happens.

One more(important) thing that I have missed is that, `/repo/gc` api would keep error empty if not passed `stream-errors`. This can be seen in code
https://github.com/ipfs/go-ipfs/blob/cd2611db377fecca28140fa1723c7a5a13295e0a/core/commands/repo.go#L79-L106 

I tried to confirm this by manually adding an error, running with `?stream-errors=true`, but it would just get stuck instead of provide output.

I would pass this argument while making calls to IPFS so that we can see errors.

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: That would mean adding one more cluster method and rpc method say `RepoGCLocal`. So far I have directly called IPFSConnector's rpc call. 
But I should change this and add `RepoGCLocal`, since that's how everywhere we have handled `?local=true` option

hsanjuan: previous comment is unrelated to `local`. In general yes, copy from how `Status*` things are done. Previous comments just asks to save this rpc call since you can set that information later.

kishansagathiya:
Still some changes remaining, will add them soon

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: In `globalPinInfoCid`, peername is set to the peer string. I don't see any way to get that accept calling `ID` (in cluster method or connector method). If skipping that rpc call is more important, there is no reason to keep peer name in repo gc struct, since it will be same as peer id.

hsanjuan: Ah yes, on error it sets the peer ID as peer name. But otherwise the pintracker has set it because it's initialized with the peername (and does not need to make an rpc). In any case, you can set the peername correctly in there directly too in cluster.go. (sorry, I thought globalPinInfo did the same).

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: I tried to check if IPFS keeps any time out when `repo gc` command or API is called, but didn't find anything helpful
have kept this 24 hours, but not sure what should be the right timeout here

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: Added a special timeout, `RepoGCTimeout` set at 24 hours by default

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: Uses cluster peer ID now

cluster.go
kishansagathiya: It is now similar to `globalPinInfoCid`

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: Have removed checking timeout with progress (`repoGCProgress`)

ipfsconn/ipfshttp/ipfshttp.go
kishansagathiya: It now uses the suggested error handling

cluster.go
kishansagathiya: Error handling is now similar to `globalPinInfoCid`

cmd/ipfs-cluster-ctl/formatters.go
kishansagathiya: Have added support for local, peers are now sorted and peername is added

cluster.go
```diff
@@ -1741,3 +1741,76 @@ func diffPeers(peers1, peers2 []peer.ID) (added, removed []peer.ID) {
...
+		rpcutil.CopyRepoGCSliceToIfaces(repoGCsResp),
+	)
+
+	// clubbing repo gc responses of all peers into one
+	globalRepoGC := api.GlobalRepoGC{PeerMap: make(map[string]*api.RepoGC)}
+	for i, resp := range repoGCsResp {
+		e := errs[i]
+
+		if e == nil {
+			globalRepoGC.PeerMap[peer.IDB58Encode(resp.Peer)] = resp
```
hsanjuan: I think you should set Peername/ID here for every received resp? (also same for errors below)

hsanjuan: ah no, nevermind

cluster.go
```diff
@@ -1741,3 +1741,76 @@ func diffPeers(peers1, peers2 []peer.ID) (added, removed []peer.ID) {
...
+
+		if rpc.IsAuthorizationError(e) {
+			logger.Debug("rpc auth error:", e)
+			continue
+		}
+
+		logger.Errorf("%s: error in broadcast response from %s: %s ", c.id, members[i], e)
+
+		resp.Error = e.Error()
+		globalRepoGC.PeerMap[peer.IDB58Encode(resp.Peer)] = resp
```
hsanjuan: I think you can't do this because the resuIting resp will be an empty object. Either RPC has failed to contact that peer, or it did contact it but the method returned error. In any case, the response will not be set at all -at least for remote peers- even if you have returned it in `RepoGCLocal`.

***This is handled in `globalPinInfoCid`*** (third time I am correcting something because it was not done in the same way as that method). You need to set the peer ID and Peername etc manually (you don't have the peername for the remote peer so you can leave it empty or set it to peer ID).

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -806,6 +811,47 @@ func (ipfs *Connector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, error)
...
+		Keys: make([]api.IPFSRepoGC, 0),
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, ipfs.config.RepoGCTimeout)
+	defer cancel()
+
+	res, err := ipfs.doPostCtx(ctx, ipfs.client, ipfs.apiURL(), "repo/gc?stream-errors=true", "", nil)
+	if err != nil {
+		logger.Error(err)
+		return &repoGC, err
```
hsanjuan: return nil,err

You can move the `repoGC` declaration down, closer where it is used.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -806,6 +811,47 @@ func (ipfs *Connector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, error)
 	return &stats, nil
 }
 
+// RepoGC performs a garbage collection sweep on the cluster peer's IPFS repo.
+func (ipfs *Connector) RepoGC(ctx context.Context) (*api.RepoGC, error) {
+	ctx, span := trace.StartSpan(ctx, "ipfsconn/ipfshttp/RepoGC")
+	defer span.End()
+
+	repoGC := api.RepoGC{
+		Keys: make([]api.IPFSRepoGC, 0),
```
hsanjuan: More idiomatic to do `Keys: []api.IPFSRepoGC{}`

hsanjuan: Also I see no reasons for it to not be a pointer from here (now doing `&repoGC` in multiple places).

rpc_api.go
```diff
@@ -513,6 +533,16 @@ func (rpcapi *IPFSConnectorRPCAPI) RepoStat(ctx context.Context, in struct{}, ou
 	return err
 }
 
+// RepoGC performs garbage a collection sweep on the cluster peer's IPFS repo.
+func (rpcapi *IPFSConnectorRPCAPI) RepoGC(ctx context.Context, in struct{}, out *api.RepoGC) error {
```
hsanjuan: With the latest changes this endpoint is not necessary

cluster.go
```diff
@@ -1741,3 +1741,76 @@ func diffPeers(peers1, peers2 []peer.ID) (added, removed []peer.ID) {
...
+	return &globalRepoGC, nil
+}
+
+// RepoGCLocal perform garbage collection only on the local IPFS deamon.
+func (c *Cluster) RepoGCLocal(ctx context.Context) (*api.RepoGC, error) {
+	_, span := trace.StartSpan(ctx, "cluster/RepoGCLocal")
+	defer span.End()
+	ctx = trace.NewContext(c.ctx, span)
+
+	repoGC := &api.RepoGC{
```
hsanjuan: This is overly convoluted. 

```go
resp, err := c.ipfs.RepoGC(ctx)
if err != nil {
  return nil, err
}
resp.Peer = c.id
resp.Peername = c.config.Peername
return resp, nil
```

hsanjuan:
Sorry, I now realize you had not requested re-review on this one.

cluster.go
kishansagathiya: can't `return nil, err`
finding an error doesn't mean response is nil https://github.com/ipfs/ipfs-cluster/pull/777/files#diff-66308423c5638970cde2dd02574201c3R847

When cluster's `RepoGC` calls `RepoGCLocal`, it will fill RepoGC struct with peer and error. I have tried to produce same response, even if `RepoGCLocal` rpc call is called directly.

cluster.go
kishansagathiya: didn't consider that `resp` could be nil here.

In `globalPinInfoCid`, what's the reason for not setting authorization errors in PinInfo structs?
https://github.com/ipfs/ipfs-cluster/blob/45e9a313db0d312970bbc847e51824be0dcede22/cluster.go#L1530-L1533

kishansagathiya:
Should I hijack `/repo/gc` endpoint in proxy, but can't produce streaming responses as in IPFS?

cluster.go
hsanjuan: I re-iterate my ask to do this this way. If you had an error, forget about the response. It's an error. This makes assumptions about how something is implemented somewhere else when it does not need to.

When RepoGC calls RepoGCLocal, if there's an error, the response will be discarded anyways (see rpc_api.go).

cluster.go
hsanjuan: Resp is not nil. Resp is simply not set to anything other than the empty object you made the call with.

> In globalPinInfoCid, what's the reason for not setting authorization errors in PinInfo structs?

An auth error is not something out of the ordinary in follower modes or that the user can fix. I chose to ignore them. We can reverse this decision if needed, but for the moment we should keep consistency.

kishansagathiya:
Will go through this once more to see if more tests could be added.

hsanjuan:
> Should I hijack `/repo/gc` endpoint in proxy, but can't produce streaming responses as in IPFS?

Make a PR to this PR for that, please.

hsanjuan:
Assume stream-errors=false

cluster.go
```diff
@@ -1741,3 +1741,73 @@ func diffPeers(peers1, peers2 []peer.ID) (added, removed []peer.ID) {
...
+			logger.Debug("rpc auth error:", e)
+			continue
+		}
+
+		logger.Errorf("%s: error in broadcast response from %s: %s ", c.id, members[i], e)
+
+		globalRepoGC.PeerMap[peer.IDB58Encode(members[i])] = &api.RepoGC{
+			Peer:     members[i],
+			Peername: peer.IDB58Encode(members[i]),
+			Keys:     []api.IPFSRepoGC{},
```
hsanjuan: I don't see an initial advantage in carrying a non-nil object here (other than not being able to omit it when encoding).

cluster_test.go
```diff
@@ -908,3 +918,54 @@ func TestClusterRecoverAllLocal(t *testing.T) {
...
+		t.Error("expected a cluster ID")
+	}
+	if repoGC.Error != "" {
+		t.Error("did not expect any error")
+	}
+
+	if repoGC.Keys == nil {
+		t.Error("expected a non-nil array of IPFSRepoGC")
+	} else {
+		if !repoGC.Keys[0].Key.Equals(test.Cid1) {
```
hsanjuan: same comment as other tests

ipfsconn/ipfshttp/config_test.go
```diff
@@ -14,7 +14,8 @@ var cfgJSON = []byte(`
       "pin_method": "pin",
       "ipfs_request_timeout": "5m0s",
       "pin_timeout": "24h",
-      "unpin_timeout": "3h"
+	  "unpin_timeout": "3h",
```
hsanjuan: Indenting seems bad here

ipfsconn/ipfshttp/ipfshttp_test.go
```diff
@@ -376,3 +376,26 @@ func TestConfigKey(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if res.Error != "" {
+		t.Errorf("expected error to be empty: %s", res.Error)
+	}
+
+	if res.Keys[0].Key.Equals(test.Cid1) {
+		t.Errorf("expected different cid, expected: %s, found: %s\n", test.Cid1, res.Keys[0].Key.String())
```
hsanjuan: `.String()` is not needed. Also, check that length of `res.Keys` is more than 0 to avoid panics.

rpc_policy.go
```diff
@@ -34,15 +36,15 @@ var DefaultRPCPolicy = map[string]RPCEndpointType{
 	"Cluster.UnpinPath":          RPCClosed,
 	"Cluster.Version":            RPCOpen,
 
-	// PinTracker methods
+	//  PinTracker methods
```
hsanjuan: might take the chance to fix the extra space in the generator

test/ipfs_mock.go
```diff
@@ -317,6 +322,29 @@ func (m *IpfsMock) handler(w http.ResponseWriter, r *http.Request) {
 			goto ERROR
 		}
 		w.Write(data)
+	case "repo/gc":
```
hsanjuan: add comment saying it assumes stream-errors=true. Actually, you may want to set one of the errors.

test/rpc_api_mock.go
```diff
@@ -307,6 +307,35 @@ func (mock *mockCluster) BlockAllocate(ctx context.Context, in *api.Pin, out *[]
 	return nil
 }
 
+func (mock *mockCluster) RepoGC(ctx context.Context, in struct{}, out *api.GlobalRepoGC) error {
+	*out = api.GlobalRepoGC{
+		PeerMap: map[string]*api.RepoGC{
+			peer.IDB58Encode(PeerID1): &api.RepoGC{
```
hsanjuan: you can rely on RepoGCLocal to produce the local object you use in RepoGCGlobal

cluster.go
kishansagathiya: I don't have a strong reason for this. But I have seen empty arrays being sent instead of nil at some places `GET /allocations` `GET /pins`. It also saves the nil check. In the output, I find it more informative to see `[]`, compared to `null`. I can change this if you want.

hsanjuan: If there' s an error, should not be checking if things are `nil`. But fine.

kishansagathiya:
I will add sharness tests and will go through this once more to look for any errors

hsanjuan:
@kishansagathiya : after conversations with Matt, we cannot run GC at once in all peers. 

Changes:

* Rebase and remove conflicts
* We should start by doing GC sequentially (only start with a new peer after the others have finished).

Given this is in very acceptable state, got a bit old and had a very long discussion, I suggest to close this and re-post on a new PR so that we can have a fresh slate.

kishansagathiya:
> We should start by doing GC sequentially (only start with a new peer after the others have finished).

RepoGC REST API should make calls (sequentially) to cluster.RepoGCLocal of each member (get ids using `GET /id`), instead of calling its own cluster.RepoGC. That way user would have a streaming response (more user friendly). sound good?

hsanjuan:
> > We should start by doing GC sequentially (only start with a new peer after the others have finished).
> 
> RepoGC REST API should make calls (sequentially) to cluster.RepoGCLocal of each member (get ids using `GET /id`), instead of calling its own cluster.RepoGC. That way user would have a streaming response (more user friendly). sound good?

No, the REST API is a wrapper on the Go API and does not do streaming (except for add). Therefore it calls cluster.RepoGC. Streaming should come with RPC streaming and it will be simpler if we have consistency on how the restapi does things (what you are proposing is not consistent with the rest of endpoints).

PR #781, opened 2019-05-17
Fix #765: Remove gx

### LINKED ISSUE #765: Fully remove Gx
They don't bother anyone but probably will never be used again
This would mean getting rid off `deptools` as well

pkafei:
Deleted `gx` from ipfs-cluster code base.


pkafei:
@kishansagathiya Can you tell me specifically which important targets should be kept in MAKEfile? It also seems like I need to also figure out why the build is failing.

hsanjuan:
@pkafei simply don't delete any targets that are not gx-specific. For example, the `install` target. Some tests are failing because of that.

kishansagathiya:
@pkafei The commits that you have made are from the user `Protocol Labs` instead of `pkafei`, I don't know if that can be a problem, but It's better if you change that.

pkafei:
Also will run tests before I push up code again!

Makefile
```diff
@@ -1,6 +1,5 @@
 deptools=deptools
 sharness = sharness/lib/sharness
-GXENABLED=no # Set to yes for Gx builds.
 gx=$(deptools)/gx
 gx-go=$(deptools)/gx-go
```
kishansagathiya: gx and gx-go should be unused now

Makefile
```diff
@@ -1,6 +1,5 @@
 deptools=deptools
```
kishansagathiya: unused, this was used only to store binaries of gx and gx-go

Makefile
```diff
@@ -34,32 +33,6 @@ service: gx-deps
 ctl: gx-deps
```
kishansagathiya: Don't use removed targets `docker_gx-deps`, `gx-deps` (here and at all other places)

Makefile
```diff
@@ -121,4 +82,4 @@ docker-compose:
 
 prcheck: gx-deps check service ctl test
 
-.PHONY: all gx gx-deps test test_sharness clean_sharness rw rwundo publish service ctl install clean gx-clean docker
+.PHONY: all gx gx-deps test test_sharness clean_sharness rw rwundo publish service ctl install clean gx-clean docker
```
lanzafame: Remove `gx gx-deps gx-clean`

lanzafame:
The go module env to expose is `GO111MODULE=one`

kishansagathiya:
I have added changes to clear doubts arising here (https://github.com/ipfs/ipfs-cluster/issues/825), hope that's okay

Makefile
```diff
@@ -1,49 +1,44 @@
...
 	@rm -rf ./test/testingData
 	@rm -rf ./compose
 
-install: gx-deps
+install:
 	$(MAKE) -C cmd/ipfs-cluster-service install
 	$(MAKE) -C cmd/ipfs-cluster-ctl install
 
-docker_install: docker_gx-deps
+docker_install:
```
hsanjuan: the `docker` target is missing

hsanjuan:
Needs rebase, also let's make sure `GO111Module=on` is exported on the `Makefile`. Can you finish this @kishansagathiya ?

Makefile
kishansagathiya: `docker_install` was install but gx would use local deps . Now with `gx`gone, `docker_install` is same as `install`, so probably we can get rid off this target.

lanzafame:
@kishansagathiya can we put the badger fixes in a separate PR?

kishansagathiya:
@lanzafame 
> @kishansagathiya can we put the badger fixes in a separate PR?

We can always put that in a separate commit to show it differently (and remove gx in a separate commit). If that's not good, I can surely make a new PR.

hsanjuan:
I have removed the deps update. We need to do a more complicated depth bubbling also involving libp2p-core etc so I'll do that in a different PR. I'll merge this one for the moment.

PR #792, opened 2019-05-23
Fix #787: Connectivity fixes

### LINKED ISSUE #787: Both consensus are bleeding peers
### Additional information
 - OS: Ubuntu
 - ~~IPFS Cluster version:~~
 - Installation method: built from source

### Description

If there's network disturbance between a node and the rest of the cluster, that node becomes its own cluster and doesn't attempt to re-attach to the bootstrap target. This quickly leaves the cluster in a fragmented or irrecoverable state. Manual intervention is required to restore the cluster.

### Suggestion

Alternative 1 : Hooks

Peers should remember every peers and attempt to get them to re-join the cluster if they still have the correct secret. This would make a cluster resistant to change because other nodes would tell them to join back. Those with a different secret can be safely forgotten.

Alternative 2 : Call for help

Peers should remember the node used for the bootstrap process. If that node becomes unreachable, then the peer should attempt every minute to bootstrap again. This would make change propagate quickly since every node would be dependent on their bootstrap.

---

In environment where network disturbance is common place, IPFS-Cluster is currently absolutely unusable. Should existing consensus be adapted, or is a new consensus needed?

hsanjuan:
Fix #787: Connectivity fixes

Currently, unless doing Join() (--bootstrap), we do not connect to any peers on startup.

We however loaded up the peerstore file and Raft would automatically connect
older peers in its peerset to figure out who is the leader, catch up etc. DHT bootstrap, after Raft
was working, did the rest. This worked more or less well for Raft.

For CRDTs we need to connect to people on a normal boot as otherwise, unless
bootstrapping, this does not happen, even if the peerstore contains known peers.

This introduces a number of changes:

* Move peerstore file management back inside the Cluster component, which was
already in charge of saving the peerstore file.
* We keep saving all "known addresses" but we load them with a non permanent
TTL, so that there will be clean up of peers we're not connected to for long.
* "Bootstrap" (connect) to a small number of peers during Cluster component creation.
* Bootstrap the DHT asap after this, so that other cluster components can
initialize with a working peer discovery mechanism.
* CRDT Trust() method will now:
  * Protect the trusted Peer ID in the conn manager
  * Give top priority in the PeerManager to that Peer (see below)
  * Mark addresses as permanent in the Peerstore

The PeerManager now attaches priorities to peers when importing them and is
able to order them according to that priority. The result is that peers with
high priority are saved first in the peerstore file. When we load the peerstore
file, the first entries in it are given the highest priority.

This means that during startup we will connect to "trusted peers" first
(because they have been tagged with priority in the previous run and saved at
the top of the list). Once connected to a small number of peers, we let the
DHT bootstrap process in the background do the rest and discover the network.

All this makes the peerstore file a "bootstrap" list for CRDTs and we will attempt
to connect to peers on that list until some of those connections succeed.

### DIFF
--- a/api/ipfsproxy/config.go
+++ b/api/ipfsproxy/config.go
@@ -206,20 +206,23 @@ func (cfg *Config) LoadJSON(raw []byte) error {
 }
 
 func (cfg *Config) applyJSONConfig(jcfg *jsonConfig) error {
-	proxyAddr, err := ma.NewMultiaddr(jcfg.ListenMultiaddress)
-	if err != nil {
-		return fmt.Errorf("error parsing proxy listen_multiaddress: %s", err)
+	if jcfg.ListenMultiaddress != "" {
+		proxyAddr, err := ma.NewMultiaddr(jcfg.ListenMultiaddress)
+		if err != nil {
+			return fmt.Errorf("error parsing proxy listen_multiaddress: %s", err)
+		}
+		cfg.ListenAddr = proxyAddr
 	}
-	nodeAddr, err := ma.NewMultiaddr(jcfg.NodeMultiaddress)
-	if err != nil {
-		return fmt.Errorf("error parsing ipfs node_multiaddress: %s", err)
+	if jcfg.NodeMultiaddress != "" {
+		nodeAddr, err := ma.NewMultiaddr(jcfg.NodeMultiaddress)
+		if err != nil {
+			return fmt.Errorf("error parsing ipfs node_multiaddress: %s", err)
+		}
+		cfg.NodeAddr = nodeAddr
 	}
-
-	cfg.ListenAddr = proxyAddr
-	cfg.NodeAddr = nodeAddr
 	config.SetIfNotDefault(jcfg.NodeHTTPS, &cfg.NodeHTTPS)
 
-	err = config.ParseDurations(
+	err := config.ParseDurations(
 		"ipfsproxy",
 		&config.DurationOpt{Duration: jcfg.ReadTimeout, Dst: &cfg.ReadTimeout, Name: "read_timeout"},
 		&config.DurationOpt{Duration: jcfg.ReadHeaderTimeout, Dst: &cfg.ReadHeaderTimeout, Name: "read_header_timeout"},
--- a/api/rest/client/transports.go
+++ b/api/rest/client/transports.go
@@ -8,11 +8,10 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/ipfs/ipfs-cluster/api"
-
 	p2phttp "github.com/hsanjuan/go-libp2p-http"
 	libp2p "github.com/libp2p/go-libp2p"
 	ipnet "github.com/libp2p/go-libp2p-interface-pnet"
+	peer "github.com/libp2p/go-libp2p-peer"
 	peerstore "github.com/libp2p/go-libp2p-peerstore"
 	pnet "github.com/libp2p/go-libp2p-pnet"
 	madns "github.com/multiformats/go-multiaddr-dns"
@@ -41,9 +40,13 @@ func (c *defaultClient) defaultTransport() {
 func (c *defaultClient) enableLibp2p() error {
 	c.defaultTransport()
 
-	pid, addr, err := api.Libp2pMultiaddrSplit(c.config.APIAddr)
+	pinfo, err := peerstore.InfoFromP2pAddr(c.config.APIAddr)
 	if err != nil {
 		return err
+	}
+
+	if len(pinfo.Addrs) == 0 {
+		return errors.New("APIAddr only includes a Peer ID")
 	}
 
 	var prot ipnet.Protector
@@ -67,16 +70,16 @@ func (c *defaultClient) enableLibp2p() error {
 
 	ctx, cancel := context.WithTimeout(c.ctx, ResolveTimeout)
 	defer cancel()
-	resolvedAddrs, err := madns.Resolve(ctx, addr)
+	resolvedAddrs, err := madns.Resolve(ctx, pinfo.Addrs[0])
 	if err != nil {
 		return err
 	}
 
-	h.Peerstore().AddAddrs(pid, resolvedAddrs, peerstore.PermanentAddrTTL)
+	h.Peerstore().AddAddrs(pinfo.ID, resolvedAddrs, peerstore.PermanentAddrTTL)
 	c.transport.RegisterProtocol("libp2p", p2phttp.NewTransport(h))
 	c.net = "libp2p"
 	c.p2p = h
-	c.hostname = pid.Pretty()
+	c.hostname = peer.IDB58Encode(pinfo.ID)
 	return nil
 }
 
--- a/cluster.go
+++ b/cluster.go
@@ -24,6 +24,7 @@ import (
 	host "github.com/libp2p/go-libp2p-host"
 	dht "github.com/libp2p/go-libp2p-kad-dht"
 	peer "github.com/libp2p/go-libp2p-peer"
+	peerstore "github.com/libp2p/go-libp2p-peerstore"
 	ma "github.com/multiformats/go-multiaddr"
 
 	ocgorpc "github.com/lanzafame/go-libp2p-ocgorpc"
@@ -36,7 +37,10 @@ import (
 // consensus layer.
 var ReadyTimeout = 30 * time.Second
 
-var pingMetricName = "ping"
+const (
+	pingMetricName = "ping"
+	bootstrapCount = 3
+)
 
 // Cluster is the main IPFS cluster component. It provides
 // the go-API for it and orchestrates the components that make up the system.
@@ -116,9 +120,7 @@ func NewCluster(
 
 	logger.Infof("IPFS Cluster v%s listening on:\n%s\n", version.Version, listenAddrs)
 
-	// Note, we already loaded peers from peerstore into the host
-	// in daemon.go.
-	peerManager := pstoremgr.New(host, cfg.GetPeerstorePath())
+	peerManager := pstoremgr.New(ctx, host, cfg.GetPeerstorePath())
 
 	c := &Cluster{
 		ctx:         ctx,
@@ -144,6 +146,18 @@ func NewCluster(
 		readyB:      false,
 	}
 
+	// Import known cluster peers from peerstore file. Set
+	// a non permanent TTL.
+	c.peerManager.ImportPeersFromPeerstore(false, peerstore.AddressTTL)
+	// Attempt to connect to some peers (up to bootstrapCount)
+	actualCount := c.peerManager.Bootstrap(bootstrapCount)
+	// We cannot warn about this as this is normal if going to Join() later
+	logger.Debugf("bootstrap count %d", actualCount)
+	// Bootstrap the DHT now that we possibly have some connections
+	c.dht.Bootstrap(c.ctx)
+
+	// After setupRPC components can do their tasks with a fully operative
+	// routed libp2p host with some connections and a working DHT (hopefully).
 	err = c.setupRPC()
 	if err != nil {
 		c.Shutdown(ctx)
@@ -465,9 +479,6 @@ This might be due to one or several causes:
 
 	// Cluster is ready.
 
-	// Bootstrap the DHT now that we possibly have some connections
-	c.dht.Bootstrap(c.ctx)
-
 	peers, err := c.consensus.Peers(ctx)
 	if err != nil {
 		logger.Error(err)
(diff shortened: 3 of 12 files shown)


api/ipfsproxy/config.go
```diff
@@ -206,20 +206,23 @@ func (cfg *Config) LoadJSON(raw []byte) error {
...
+		proxyAddr, err := ma.NewMultiaddr(jcfg.ListenMultiaddress)
+		if err != nil {
+			return fmt.Errorf("error parsing proxy listen_multiaddress: %s", err)
+		}
+		cfg.ListenAddr = proxyAddr
 	}
-	nodeAddr, err := ma.NewMultiaddr(jcfg.NodeMultiaddress)
-	if err != nil {
-		return fmt.Errorf("error parsing ipfs node_multiaddress: %s", err)
+	if jcfg.NodeMultiaddress != "" {
```
hsanjuan: These two changes are two fix a broken tests that I must have missed when updating deps.

api/rest/client/transports.go
```diff
@@ -41,11 +40,15 @@ func (c *defaultClient) defaultTransport() {
 func (c *defaultClient) enableLibp2p() error {
 	c.defaultTransport()
 
-	pid, addr, err := api.Libp2pMultiaddrSplit(c.config.APIAddr)
+	pinfo, err := peerstore.InfoFromP2pAddr(c.config.APIAddr)
```
hsanjuan: A function in `peerstore` does mostly the same as a function we had, so started using that one.

cluster.go
```diff
@@ -36,7 +37,10 @@ import (
 // consensus layer.
 var ReadyTimeout = 30 * time.Second
 
-var pingMetricName = "ping"
+const (
+	pingMetricName = "ping"
+	bootstrapCount = 3
```
hsanjuan: This is the number of peers we attempt to connect. We block until 3 connections succeed or we run out of addresses to try. After that the DHT will take care of discovering the swarm in the background.

cluster.go
```diff
@@ -633,12 +644,24 @@ func (c *Cluster) ID(ctx context.Context) *api.ID {
 		peers, _ = c.consensus.Peers(ctx)
 	}
 
+	clusterPeerInfos := c.peerManager.PeerInfos(peers)
```
hsanjuan: This change block is because the PeerManager does not return `api.Multiaddr` type anymore (which is our serializable wrapper for multiaddresses). Instead we need to build that address list by hand.

cluster.go
```diff
@@ -721,20 +744,15 @@ func (c *Cluster) Join(ctx context.Context, addr ma.Multiaddr) error {
 
 	logger.Debugf("Join(%s)", addr)
 
-	pid, _, err := api.Libp2pMultiaddrSplit(addr)
+	// Add peer to peerstore so we can talk to it (and connect)
+	pid, err := c.peerManager.ImportPeer(addr, true, peerstore.PermanentAddrTTL)
```
hsanjuan: returning the pid from that function allowed some simplication here, but it does not change behaviour.

consensus/crdt/consensus.go
```diff
@@ -22,13 +23,15 @@ import (
...
+	peerstore "github.com/libp2p/go-libp2p-peerstore"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 )
 
 var logger = logging.Logger("crdt")
 
 var (
-	blocksNs = "b" // blockstore namespace
+	blocksNs   = "b" // blockstore namespace
+	connMgrTag = "crdt"
```
hsanjuan: The connmanager needs a tag when we tag something for Protect().

consensus/crdt/consensus.go
```diff
@@ -85,19 +89,21 @@ func New(
...
-		store:     store,
-		namespace: ds.NewKey(cfg.DatastoreNamespace),
-		pubsub:    pubsub,
-		rpcReady:  make(chan struct{}, 1),
-		readyCh:   make(chan struct{}, 1),
+		ctx:         ctx,
+		cancel:      cancel,
+		config:      cfg,
+		host:        host,
+		peerManager: pstoremgr.New(ctx, host, ""),
```
hsanjuan: We keep a peerManager around to set higher priority to trusted peers

consensus/crdt/consensus.go
```diff
@@ -296,9 +302,18 @@ func (css *Consensus) IsTrustedPeer(ctx context.Context, pid peer.ID) bool {
...
+// Trust marks a peer as "trusted". It makes sure it is trusted as issuer
+// for pubsub updates, it is protected in the connection manager, it
+// has the highest priority when the peerstore is saved, and it's addresses
+// are always remembered.
 func (css *Consensus) Trust(ctx context.Context, pid peer.ID) error {
 	css.trustedPeers.Store(pid, struct{}{})
+	if conman := css.host.ConnManager(); conman != nil {
+		conman.Protect(pid, connMgrTag)
+	}
+	css.peerManager.SetPriority(pid, 0)
```
hsanjuan: 0 means highest priority.

pstoremgr/pstoremgr.go
```diff
@@ -54,38 +57,33 @@ func New(h host.Host, peerstorePath string) *Manager {
...
+	pinfo, err := peerstore.InfoFromP2pAddr(addr)
 	if err != nil {
-		return err
+		return "", err
 	}
-	pm.host.Peerstore().AddAddr(pid, decapAddr, peerstore.PermanentAddrTTL)
 
-	// dns multiaddresses need to be resolved because libp2p only does that
-	// on explicit bhost.Connect().
-	if madns.Matches(addr) {
```
hsanjuan: By using `Connect()` below we can get rid of resolving here. `Connect` will resolve. And also use the DHT if no addresses are known for a peer. This was a bug, because `Network().DialPeer()` does not use the DHT when no addresses are known.

pstoremgr/pstoremgr.go
```diff
@@ -123,11 +119,12 @@ func (pm *Manager) filteredPeerAddrs(p peer.ID) []api.Multiaddr {
...
-// /ipfs/<peerID> part) for the given set of peers. For peers for which
-// we know DNS multiaddresses, we only return those. Otherwise, we return
-// all the multiaddresses known for that peer.
-func (pm *Manager) PeersAddresses(peers []peer.ID) []api.Multiaddr {
+// PeerInfos returns a slice of peerinfos for the given set of peers in order
+// of priority. For peers for which we know DNS
+// multiaddresses, we only include those. Otherwise, the PeerInfo includes all
+// the multiaddresses known for that peer. Peers without addresses are not
+// included.
+func (pm *Manager) PeerInfos(peers []peer.ID) []peerstore.PeerInfo {
```
hsanjuan: Easier to work with PeerInfos so changing this method. Note that returned slice is ordered by priority.

pstoremgr/pstoremgr.go
```diff
@@ -136,29 +133,47 @@ func (pm *Manager) PeersAddresses(peers []peer.ID) []api.Multiaddr {
...
-func (pm *Manager) ImportPeers(addrs []ma.Multiaddr, connect bool) error {
-	for _, a := range addrs {
-		pm.ImportPeer(a, connect)
+// given connect parameter. Peers are tagged with priority as given
+// by their position in the list.
+func (pm *Manager) ImportPeers(addrs []ma.Multiaddr, connect bool, ttl time.Duration) error {
+	for i, a := range addrs {
+		pid, err := pm.ImportPeer(a, connect, ttl)
+		if err == nil {
+			pm.SetPriority(pid, i)
```
hsanjuan: The first peer in the slice gets priority 0 (highest). Loading the peerstore uses this so the addresses at the top of the peerstore get the highest priorities and will be saved in the same order. Bootstrap() also uses these priorities to contact peers starting from the highest priority (0).

pstoremgr/pstoremgr.go
```diff
@@ -221,13 +236,95 @@ func (pm *Manager) SavePeerstore(addrs []api.Multiaddr) {
...
+
+	prio1iface, err := ps.pstore.Get(pinfo1.ID, PriorityTag)
+	if err == nil {
+		prio1 = prio1iface.(int)
+	}
+	prio2iface, err := ps.pstore.Get(pinfo2.ID, PriorityTag)
+	if err == nil {
+		prio2 = prio2iface.(int)
+	}
+	return prio1 < prio2
```
hsanjuan: A peer without priority has the highest priority (0). This is on purpose.

pstoremgr/pstoremgr_test.go
```diff
@@ -4,21 +4,21 @@ import (
...
 
-	"github.com/ipfs/ipfs-cluster/api"
+	"github.com/ipfs/ipfs-cluster/test"
 
 	libp2p "github.com/libp2p/go-libp2p"
+	peer "github.com/libp2p/go-libp2p-peer"
 	ma "github.com/multiformats/go-multiaddr"
 )
 
-var pid = "QmXZrtE5jQwXNqCJMfHUTQkvhQ4ZAnqMnmzFMJfLewuabc"
```
hsanjuan: Cleaned up these tests a bit to not work with strings. Only added one tests at the end.

pstoremgr/pstoremgr_test.go
```diff
@@ -27,79 +27,169 @@ func clean(pm *Manager) {
...
+	if len(pinfos) != 1 {
+		t.Fatal("expected 1 peer in the peerstore")
+	}
+
+	if len(pinfos[0].Addrs) != 2 {
+		t.Error("expected 2 addresses")
+	}
+}
+
+func TestPriority(t *testing.T) {
```
hsanjuan: This is the new test.

pstoremgr/pstoremgr.go
lanzafame: This should probably be a comment in the code

hsanjuan:
There is a part missing here about keeping connections open to trusted peers... For small scale clusters, I think the DHT bootstrapping should make that happen (pretty much everyone will connect)... but we can manually `Connect` to trusted peers in a different PR (although I don't like it as a general approach, we should be ok until we hit many thousands of peers per cluster - IPFS bootstrappers deal with this issue just fine).

pstoremgr/pstoremgr.go
```diff
@@ -221,13 +236,95 @@ func (pm *Manager) SavePeerstore(addrs []api.Multiaddr) {
...
+// SavePeerstoreForPeers calls PeerInfos and then saves the peerstore
 // file using the result.
 func (pm *Manager) SavePeerstoreForPeers(peers []peer.ID) {
-	pm.SavePeerstore(pm.PeersAddresses(peers))
+	pm.SavePeerstore(pm.PeerInfos(peers))
+}
+
+// Bootstrap attempts to get as much as count connected peers by selecting
+// randomly from those in the libp2p host peerstore. It returns the number
+// if peers it sucessfully connected to.
```
kishansagathiya:
```suggestion
// of peers it successfully connected to.
```

pstoremgr/pstoremgr.go
```diff
@@ -221,13 +236,95 @@ func (pm *Manager) SavePeerstore(addrs []api.Multiaddr) {
...
+	for i := 0; i < lenKnown && totalConns < count; i++ {
+		pinfo := pinfos[i]
+		ctx, cancel := context.WithTimeout(pm.ctx, ConnectTimeout)
+		defer cancel()
+
+		logger.Infof("connecting to %s", pinfo.ID)
+		err := pm.host.Connect(ctx, pinfo)
+		if err != nil {
+			logger.Warning(err)
+			pm.SetPriority(pinfo.ID, 9999)
```
kishansagathiya: Say, we got an error while connecting to a trusted peer. This would send it down the list. Isn't it more important to stay connected to trusted peers?

pstoremgr/pstoremgr.go
```diff
@@ -25,10 +24,14 @@ import (
...
-// Timeouts for network operations triggered by the Manager
+// PriorityTag is used to attach metadata to peers in the peerstore
+// so they can be sorted.
+var PriorityTag = "cluster"
+
+// Timeouts for network operations triggered by the Manager.
 var (
 	DNSTimeout     = 5 * time.Second
-	ConnectTimeout = 10 * time.Second
+	ConnectTimeout = 5 * time.Second
```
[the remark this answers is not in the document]
hsanjuan: Because we now block while connecting to peers in the `Bootstrap()` function and we should skip over non-responding peers a bit more quickly.

pstoremgr/pstoremgr.go
hsanjuan: Hmm  yeah, I will move the code that SetsPriority=0 in crdt consensus to the `setup()` function. This will ensure that even it was demoted on bootstrapping, it will go up the list soon after. Right now, crdt is setting priorities even before the peers are imported and I'm not sure how well that works (and it's not correctly setting the TTL on the addresses anyway because there will be no addresses, so that's buggy).

Good catch.

PR #806, opened 2019-06-04
Fix: api: replace two regexes with strings.HasPrefix()

MichaelMure:
Those regexes were compiled with each call to the function. As it's
called by PinLs, this resulted in a significant amount of memory used,
500MB in my case after a single call.

License: MIT
Signed-off-by: Michael Muré <batolettre@gmail.com>

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -12,7 +12,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


MichaelMure:
As another data point, this two regexes amount to 62% of the allocated memory of a long running instance.

lanzafame:
@MichaelMure Thanks very much for this PR. I quickly benchmarked it and it is a huge improvement :+1: 

```
$ go test -run none -bench IPFSPinStatus -benchmem -benchtime 3s
goos: linux
goarch: amd64
pkg: github.com/ipfs/ipfs-cluster/api
BenchmarkIPFSPinStatusFromString-8   	 1000000	      3877 ns/op	    4378 B/op	      39 allocs/op
PASS
ok  	github.com/ipfs/ipfs-cluster/api	3.921s
$ go test -run none -bench IPFSPinStatus -benchmem -benchtime 3s
goos: linux
goarch: amd64
pkg: github.com/ipfs/ipfs-cluster/api
BenchmarkIPFSPinStatusFromString-8   	5000000000	         1.56 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/ipfs/ipfs-cluster/api	7.963s
```

hsanjuan:
ouch

PR #807, opened 2019-06-05
add IPFSPinStatusFromString benchmark

lanzafame:
Follow on from #806.

### DIFF
--- a/api/types_test.go
+++ b/api/types_test.go
@@ -53,6 +53,12 @@ func TestIPFSPinStatusFromString(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


hsanjuan:
I'm not sure this code will be super useful as part of the codebase (good to test/verify the related fix), but merge at will :)

lanzafame:
> I'm not sure this code will be super useful as part of the codebase

Yeah, I know what you mean but I do want to set up some benchmarking infrastructure so that we can record the diff in perf on PR basis, so once that is in place having benchmarks already in place will be beneficial.

PR #808, opened 2019-06-05
Fix: default base32 in IPFS breaks cluster pin-ls

hsanjuan:
The switch to base32 as default output format for cidv1 keys in IPFS pin/ls
responses causes an error in PinLsCid as the base32 string does not correspond
to the base58 string expected by cluster when using an old peer.

This affects old Cluster peers running with new IPFS versions and new cluster
peers running with old IPFS versions for v1 CIDs.

Since Current master uses an updated cid dependency which also uses base32 by
default, master would already work with the latest IPFS daemon, so this is
just allowing to use cluster peers with older ipfs daemons, and preventing a
similar breakage in the future.


hsanjuan:
@lanzafame added a test. Verified that the test fails without the fix and passes with the fix.

Issue #805, opened 2019-06-04
Bootstrap from loaded peerstore fails to connect to peers


lanzafame:
https://github.com/ipfs/ipfs-cluster/blob/master/cluster.go#L153 this fails to actually 
"connect". As in it says it connects but `peers ls` doesn't show the supposedly connected peers.

hsanjuan:
no, because `peers ls` only shows the peers once you receive `ping` metrics from them, even if you are connected before that. This is bad UI but workaroundable by inserting a metric when connection succeeds (like on `Join`)

Issue #798, opened 2019-05-28
[CRDT] Reconnecting a peer causes issues receiving deltas and sending out pins


lanzafame:
**Additional information:**
 - OS: Linux
 - IPFS Cluster version: master
 - Installation method: built from source

**Describe the bug:**
Three peer cluster, two peers bootstrapped to the first peer. Shutdown 2nd peer and then start again without `--bootstrap` as the trusted peers `PeerInfo` should now be in the peerstore and it will connect to those peers when it starts back up. This worked fine, confirmed with a `peers ls` against the restarted peer. The issue occurred when pin operations against the first peer started. 

The following is an excerpt of the logs from the restarted peer:
```
15:57:00.335 ERROR       crdt: error getting root delta priority: %s failed to get block for QmVGD6aZQsbjpLpNTWEnK57kYLLtBaFSn4k39hGMw1nzmq: context deadline exceeded crdt.go:413
15:57:00.336 ERROR       crdt: error getting delta: context deadline exceeded crdt.go:420
15:57:01.456 ERROR       crdt: error getting root delta priority: %s failed to get block for Qmaz5xpm5wgMg4ULiE9WtpQz6xdURK2Wuafyi88F1dLAVS: context deadline exceeded crdt.go:413
15:57:01.456 ERROR       crdt: error getting delta: context deadline exceeded crdt.go:420
15:57:02.559 ERROR       crdt: error getting root delta priority: %s failed to get block for QmdfJ7ouWeCa88BYRgrXkGk6kjtkCyC1PHRyeHdYiGitqY: context deadline exceeded crdt.go:413
15:57:02.559 ERROR       crdt: error getting delta: context deadline exceeded crdt.go:420
```

If you `ipfs-cluster-ctl add <file>` from the out-of-sync peer, it will attempt to pin on all the peers but at the `height` of the out-of-sync peer, though it doesn't succed:
```
### out-of-sync peer
16:10:27.023  INFO    cluster: pinning QmZkDEwrfwHKfoqWg3jnqSVwEWiFCU9TRnnRM2Xe1xYNqh everywhere: cluster.go:1227
16:10:27.026  INFO       crdt: new pin added: QmZkDEwrfwHKfoqWg3jnqSVwEWiFCU9TRnnRM2Xe1xYNqh consensus.go:199
[...]

PR #811, opened 2019-06-07
Multiple fixes to CRDTs

hsanjuan:
As I'm debugging I'm fixing things and I'm putting them in this branch.

This is likely eventually going to address the last comments in #787, #798 , #805


cluster.go
```diff
@@ -150,9 +151,18 @@ func NewCluster(
...
-	// We cannot warn about this as this is normal if going to Join() later
-	logger.Debugf("bootstrap count %d", actualCount)
+	connectedPeers := c.peerManager.Bootstrap(bootstrapCount)
+	// We cannot warn when count is low as this as this is normal if going
+	// to Join() later.
+	logger.Debugf("bootstrap count %d", len(connectedPeers))
+	// Log a ping metric for every connected peer. This will make them
+	// visible as peers without having to wait for them to send one.
+	for _, p := range connectedPeers {
+		if err := c.logPingMetric(ctx, p); err != nil {
```
hsanjuan: peers we connected to will appear in the `current cluster peers` log message when cluster is Ready().

cluster.go
```diff
@@ -434,11 +491,35 @@ func (c *Cluster) repinFromPeer(ctx context.Context, p peer.ID) {
...
 // run launches some go-routines which live throughout the cluster's life
 func (c *Cluster) run() {
-	go c.syncWatcher()
-	go c.pushPingMetrics(c.ctx)
-	go c.pushInformerMetrics(c.ctx)
-	go c.watchPeers()
-	go c.alertsHandler()
+	c.wg.Add(1)
+	go func() {
+		defer c.wg.Done()
```
hsanjuan: I thought better to have all the `Add`s and `Done`s in the same place.

consensus/crdt/consensus.go
```diff
@@ -92,6 +93,24 @@ func New(
 
 	ctx, cancel := context.WithCancel(context.Background())
 
+	var blocksDatastore ds.Batching
+	ns := ds.NewKey(cfg.DatastoreNamespace)
+	blocksDatastore = namespace.Wrap(store, ns.ChildString(blocksNs))
+
+	ipfs, err := ipfslite.New(
```
hsanjuan: ipfslite (bitswap) needs to be created before connecting (bootstrapping) to any other peers. Otherwise nothing works. Thus ipfslite Peer creation is moved here.

PR #814, opened 2019-06-10
alerting for peers stops after one alert

### LINKED ISSUE #812: Monitor: alerts never stop
**Additional information:**
 - OS: Linux
 - IPFS Cluster version: master
 - Installation method: source

**Describe the bug:**

When the new monitor, alerts are triggered constantly whenever a peer dissapears. This causes re-pinning attemps constantly. We should probably not alert all the time, as some peers may as well never come back again. Possible related to #410 

@lanzafame maybe there's a simple fix for this?

lanzafame:
Resolves #812.

This PR still requires clarification of whether the alerting for a peer should stop after one alert or several.

### DIFF
--- a/monitor/metrics/checker.go
+++ b/monitor/metrics/checker.go
@@ -17,6 +17,10 @@ import (
 // AlertChannelCap specifies how much buffer the alerts channel has.
 var AlertChannelCap = 256
 
+// MaxAlertThreshold specifies how many alerts will occur per a peer is
+// removed the list of monitored peers.
+var MaxAlertThreshold = 5
+
 // ErrAlertChannelFull is returned if the alert channel is full.
 var ErrAlertChannelFull = errors.New("alert channel is full")
 
@@ -28,8 +32,10 @@ type Checker struct {
 	metrics   *Store
 	threshold float64
 
+	alertThreshold int
+
 	failedPeersMu sync.Mutex
-	failedPeers   map[peer.ID]bool
+	failedPeers   map[peer.ID]int
 }
 
 // NewChecker creates a Checker using the given
@@ -44,7 +50,7 @@ func NewChecker(ctx context.Context, metrics *Store, threshold float64) *Checker
 		alertCh:     make(chan *api.Alert, AlertChannelCap),
 		metrics:     metrics,
 		threshold:   threshold,
-		failedPeers: make(map[peer.ID]bool),
+		failedPeers: make(map[peer.ID]int),
 	}
 }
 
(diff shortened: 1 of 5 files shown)


monitor/metrics/checker.go
```diff
@@ -26,6 +27,9 @@ type Checker struct {
 	alertCh   chan *api.Alert
 	metrics   *Store
 	threshold float64
+
+	failedPeersMu sync.Mutex
+	failedPeers   map[peer.ID]bool
```
hsanjuan: `map[peer.ID]struct{}` ?

monitor/metrics/checker.go
```diff
@@ -89,6 +94,15 @@ func (mc *Checker) alertIfExpired(metric *api.Metric) error {
 }
 
 func (mc *Checker) alert(pid peer.ID, metricName string) error {
+	mc.failedPeersMu.Lock()
+	if mc.failedPeers[pid] {
+		mc.metrics.RemovePeer(pid)
+		delete(mc.failedPeers, pid)
+		return nil
```
hsanjuan: returning without `Unlock()`.

monitor/metrics/store.go
```diff
@@ -47,6 +47,15 @@ func (mtrs *Store) Add(m *api.Metric) {
 	window.Add(m)
 }
 
+// RemovePeer removes all metrics related to a peer from the Store.
+func (mtrs *Store) RemovePeer(pid peer.ID) {
+	mtrs.mux.Lock()
+	for _, mtrs := range mtrs.byName {
+		delete(mtrs, pid)
+	}
+	mtrs.mux.Unlock()
```
hsanjuan: `defer Unlock` above

hsanjuan:
Also, mind the codeclimate warning!

monitor/metrics/checker.go
hsanjuan: actually, `map[peer.ID]int` and record how many times something has alerted, and then make the threshold of number of times something should alert at most a variable at the top?

monitor/metrics/store.go
lanzafame: Overhead of defer is non-trivial, so since this doesn't have multiple returns, I preferred to leave as is.

lanzafame:
> This means that a peer re-appearing will need to fill-in the window up to 5 for the distribution to kick in and eventually trigger alerts (which is good).

This is updated in #815.

lanzafame:
@hsanjuan We should probably take the `-failfast` out of the travis config so that we know if just a single flaky test failed or if there were several other test failures.

hsanjuan:
The problem is that if the AlertThreshold is set to 1 we only send 1 alert regardless of the metrics alerting. So we end up sending only a `freespace` alert and not a `freespace` and a `ping` alert.

PR #832, opened 2019-07-09
Fix #793: Allow to fully disable unpinning

### LINKED ISSUE #793: REST API: Allow to disable unpin (and other endpoints)
Question is related to the #750 
We have 10 nodes cluster. If someone get access to the node - we need to be sure that data will not deleted from the all nodes.
Could we prevent unpin operation on a cluster configuration level? In this case even if someone get access to the node - he will not be able to force file deletion form all nodes.
As alternative - maybe cluster secret could contain information about allowed operation in the cluster.

hsanjuan:
As raised in #793, sometimes the user wants to simply disallow unpinning
altogether in a Cluster peer so that no content can be removed from IPFS
through cluster, even when crendentials are compromised (including access to
RPC endpoint).

This introduces a disable_unpinning parameter for the ipfshttp
connector. There is no way that a cluster will effecitvely unpin something if
the connector refuses to ask ipfs.

### DIFF
--- a/ipfsconn/ipfshttp/config.go
+++ b/ipfsconn/ipfshttp/config.go
@@ -24,6 +24,7 @@ const (
 	DefaultIPFSRequestTimeout = 5 * time.Minute
 	DefaultPinTimeout         = 24 * time.Hour
 	DefaultUnpinTimeout       = 3 * time.Hour
+	DefaultUnpinDisable       = false
 )
 
 // Config is used to initialize a Connector and allows to customize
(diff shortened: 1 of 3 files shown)


hsanjuan:
> This looks good.
> But shouldn't we achieve this result by blocking rpc calls for unpin instead? To me that seems like the natural place to edit to provide a limited/different set of cluster functionality.

We could but if the method is ever called the method directly (by something in the Cluster component), it won't help. I have mixed feelings about it though, since it's only called via RPC from the PinTracker anyways. Maybe I'll re-do that way.

volodyad:
Hello, Is it going to be merged in the near future?

hsanjuan:
Slept on this. Added `unpin_disable` parameter to `ipfshttp`. It is hidden (won't appear in default config).

hsanjuan:
Main reason is simplicity. It's very straightforward how this feature works and where. Remembering that the rpc endpoint has as disable unpinning option and that no ipfsconn.Unpin can be used directly from the code, or it needs to take this option into account too is more complex.

PR #836, opened 2019-07-15
Do not load API components removed from the config

### LINKED ISSUE #452: [Daemon] Do not load API components which have been removed from the config
#### Basic information

  * [ ] Bug
  * [x] Feature request
  * [ ] Enhancement

#### Description

We should be able to initialize Cluster with multiple API components.

As a nice-to-have, the `config` module should tell us when no configuration exists in the JSON for a certain component, and use that as an indication to decide which components are used when creating the `Cluster` object from `ipfs-cluster-service`.

kishansagathiya:
This commit introduces a map that would keep track of whether components
for a component were missing or not from the JSON config file. This map
can be check while creating cluster to avoid loading a component.

It would consider component only if the component is fully missing from the
config.
Say the component in question is `ipfsproxy` which is under `api`
section.
This would use defaults for `ipfsproxy` and load IPFS proxy.

```
{
"api":{
	"ipfsproxy": {}
	}
}
```

However, this would not load IPFS proxy
```
{
"api":{}
}
```

Fixes #452

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -65,33 +65,6 @@ const (
 // SectionType specifies to which section a component configuration belongs.
 type SectionType int
 
-func (s *SectionType) String() string {
-	switch *s {
-	case Consensus:
-		return "consensus"
-	case API:
-		return "api"
-	case IPFSConn:
-		return "ipfs_connector"
-	case State:
-		return "state"
-	case PinTracker:
-		return "pin_tracker"
-	case Monitor:
-		return "monitor"
-	case Allocator:
-		return "allocator"
-	case Informer:
-		return "informer"
-	case Observations:
-		return "observations"
-	case Datastore:
-		return "datastore"
-	default:
-		return ""
-	}
-}
-
 // SectionTypes returns the list of supported SectionTypes
 func SectionTypes() []SectionType {
 	var l []SectionType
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -483,6 +488,12 @@ func (cfg *Manager) ToJSON() ([]byte, error) {
 	return DefaultJSONMarshal(jcfg)
 }
 
+// EmptyComponents returns a map of component that were found
+// to have no config in JSON file.
+func (cfg *Manager) EmptyComponents() map[string]bool {
```
hsanjuan: This is not very from elegant, and assumes two components in different sections cannot have the same name...

Rather than this, create a `func IsDefined(SectionType, SectionName) bool`. Carry the cfgMgr into `CreateCluster`

config/config.go
```diff
@@ -403,7 +376,7 @@ func (cfg *Manager) LoadJSON(bs []byte) error {
 			}
 			logger.Debugf("%s component configuration loaded", name)
 		} else {
-			cfg.undefinedComps[t.String()+name] = true
+			cfg.undefinedComps[t] = map[string]bool{name: true}
```
hsanjuan: This will re-assign a new map every time, overwriting any existing and thus losing information for other components in the same section.

I suggest you create an empty map everytime a section map is created (`RegisterComponent`).

hsanjuan:
(@kishansagathiya the `signed-off` trails in every commit message are not necessary anymore)

PR #837, opened 2019-07-15
JSON Config object key should match JSON tags

### LINKED ISSUE #778: Config env variables: someconfig object keys do not match the JSON tags
**Additional information:**

 - IPFS Cluster version:  master

**Describe the bug:**

For example, `CLUSTER_RESTAPI_BASICAUTHCREDENTIALS` should work as described by the documentation, but instead it must be `CLUSTER_RESTAPI_BASICAUTHCREDS` because the `jsonConfig` object's key in `api/rest/config.go`  is `BasicAuthCreds`. We should rename this and others to match the json tag (`BasicAuthCredentials`).

This involves checking all the `config.go` files and finding similar occurences of this.

kishansagathiya:
Went through all the `config.go` files. There was only one extra inconsistency that what was mentioned in the issue.

Fixes #778

### DIFF
--- a/api/rest/config.go
+++ b/api/rest/config.go
@@ -104,9 +104,9 @@ type Config struct {
 	ID         peer.ID
 	PrivateKey crypto.PrivKey
 
-	// BasicAuthCreds is a map of username-password pairs
+	// BasicAuthCredentials is a map of username-password pairs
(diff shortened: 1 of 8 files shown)


hsanjuan:
Thanks!

Issue #835, opened 2019-07-11
[Daemon] Init should take a list of peers


hsanjuan:
**Describe the feature you are proposing**

`ipfs-cluster-service init --peers <multiaddress,multiaddress>`

It should:

1) Add and write the given peers to the peerstore file
2) For raft config section, add the peer IDs to the init_raft_peerset
3) For crdt config section, add the peer IDs to the trusted_peers slice 

**Additional context**

Mostly useful for CRDTs:

`ipfs-cluster-service init --peers /dnsaddr/mycluster.myclyster.io` should allow to run `ipfs-cluster-service daemon --consensus crdt` without needing the bootstrap flag.

lanzafame:
@hsanjuan couldn't the bootstrap flag do the same things as the proposed `--peers` flag?

hsanjuan:
> @hsanjuan couldn't the bootstrap flag do the same things as the proposed `--peers` flag?

`--bootstrap` should probably not touch the `init_raft_peerset`

Regarding adding to peerstore, I think it does (but on a different workflow, only when the peer is shutdown and the peerstore written).

Regarding trusting, yes (but I don't know if permanently)

I don't feel super comfortable with having side effects to the configuration file (re-writing it) when running the daemon.

lanzafame:
> I don't feel super comfortable with having side effects to the configuration file (re-writing it) when running the daemon.

Actually, as long as `trusted_peers` has a env var, I think the use case I had in mind is still already addressed, cloud deployment method is just inconsistent with manual deployment. Which can be solved via documentation for the time being.

PR #839, opened 2019-07-17
Init should take a list of peers

kishansagathiya:
This commit adds `--peers` option to `ipfs-cluster-service init`

`ipfs-cluster-service init --peers <multiaddress,multiaddress>`

- Adds and writes the given peers to the peerstore file
- For raft config section, adds the peer IDs to the `init_peerset`
- For crdt config section, add the peer IDs to the `trusted_peers`

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -620,7 +620,10 @@ func (c *Cluster) Shutdown(ctx context.Context) error {
 	// Try to store peerset file for all known peers whatsoever
 	// if we got ready (otherwise, don't overwrite anything)
 	if c.readyB {
-		c.peerManager.SavePeerstoreForPeers(c.host.Peerstore().Peers())
+		err := c.peerManager.SavePeerstoreForPeers(c.host.Peerstore().Peers())
+		if err != nil {
+			logger.Warningf("could not save peers to peerstore: %s", err)
+		}
 	}
 
 	// Only attempt to leave if:
--- a/cluster.go
+++ b/cluster.go
@@ -620,10 +620,8 @@ func (c *Cluster) Shutdown(ctx context.Context) error {
 	// Try to store peerset file for all known peers whatsoever
 	// if we got ready (otherwise, don't overwrite anything)
 	if c.readyB {
-		err := c.peerManager.SavePeerstoreForPeers(c.host.Peerstore().Peers())
-		if err != nil {
-			logger.Warningf("could not save peers to peerstore: %s", err)
-		}
+		// Ignoring error since it's a best-effort
+		c.peerManager.SavePeerstoreForPeers(c.host.Peerstore().Peers())
 	}
 
 	// Only attempt to leave if:
--- a/cmd/ipfs-cluster-service/main.go
+++ b/cmd/ipfs-cluster-service/main.go
@@ -310,11 +310,9 @@ remove the %s file first and clean any Raft state.
 						multiAddrs = append(multiAddrs, multiAddr)
 					}
 
-					if len(addrs) > 0 {
-						peers := ipfscluster.PeersFromMultiaddrs(multiAddrs)
-						cfgs.crdtCfg.TrustedPeers = peers
-						cfgs.raftCfg.InitPeerset = peers
-					}
+					peers := ipfscluster.PeersFromMultiaddrs(multiAddrs)
+					cfgs.crdtCfg.TrustedPeers = peers
+					cfgs.raftCfg.InitPeerset = peers
 				}
 
 				// Save
--- a/cmd/ipfs-cluster-service/main.go
+++ b/cmd/ipfs-cluster-service/main.go
@@ -303,14 +303,14 @@ remove the %s file first and clean any Raft state.
 				if peersOpt != "" {
 					addrs := strings.Split(peersOpt, ",")
 
-					if len(addrs) > 0 {
-						for _, addr := range addrs {
-							addr = strings.TrimSpace(addr)
-							multiAddr, err := ma.NewMultiaddr(addr)
-							checkErr("parsing peer multiaddress: "+addr, err)
-							multiAddrs = append(multiAddrs, multiAddr)
-						}
+					for _, addr := range addrs {
+						addr = strings.TrimSpace(addr)
+						multiAddr, err := ma.NewMultiaddr(addr)
+						checkErr("parsing peer multiaddress: "+addr, err)
+						multiAddrs = append(multiAddrs, multiAddr)
+					}
 
+					if len(addrs) > 0 {
 						peers := ipfscluster.PeersFromMultiaddrs(multiAddrs)
 						cfgs.crdtCfg.TrustedPeers = peers
 						cfgs.raftCfg.InitPeerset = peers
--- a/cmd/ipfs-cluster-service/main.go
+++ b/cmd/ipfs-cluster-service/main.go
@@ -16,7 +16,6 @@ import (
 	"github.com/ipfs/ipfs-cluster/pstoremgr"
 	"github.com/ipfs/ipfs-cluster/version"
 	peer "github.com/libp2p/go-libp2p-core/peer"
-	peerstore "github.com/libp2p/go-libp2p-core/peerstore"
 	ma "github.com/multiformats/go-multiaddr"
 
 	semver "github.com/blang/semver"
@@ -246,7 +245,7 @@ remove the %s file first and clean any Raft state.
 					Usage: "prompt for the cluster secret",
 				},
 				cli.StringFlag{
-					Name:  "peers, p",
+					Name:  "peers",
 					Usage: "comma-separated list of multiaddresses to init with",
 				},
 			},
@@ -301,18 +300,18 @@ remove the %s file first and clean any Raft state.
 
 				peersOpt := c.String("peers")
 				multiAddrs := []ma.Multiaddr{}
-				var peers []peer.ID
 				if peersOpt != "" {
 					addrs := strings.Split(peersOpt, ",")
 
 					if len(addrs) > 0 {
 						for _, addr := range addrs {
-							multiAddr, err := ma.NewMultiaddr(strings.TrimSpace(addr))
-							checkErr("parsing multiaddress", err)
+							addr = strings.TrimSpace(addr)
+							multiAddr, err := ma.NewMultiaddr(addr)
+							checkErr("parsing peer multiaddress: "+addr, err)
 							multiAddrs = append(multiAddrs, multiAddr)
 						}
 
-						peers = ipfscluster.PeersFromMultiaddrs(multiAddrs)
+						peers := ipfscluster.PeersFromMultiaddrs(multiAddrs)
 						cfgs.crdtCfg.TrustedPeers = peers
 						cfgs.raftCfg.InitPeerset = peers
 					}
@@ -334,18 +333,14 @@ remove the %s file first and clean any Raft state.
 					out("new identity written to %s\n", identityPath)
 				}
 				if peersOpt != "" {
-					_, ident, cfgs := makeAndLoadConfigs()
-
-					ctx := context.Background()
-					host, _, _, err := ipfscluster.NewClusterHost(ctx, ident, cfgs.clusterCfg)
-					checkErr("creating libp2p host", err)
+					_, _, cfgs := makeAndLoadConfigs()
 
 					peerstorePath := cfgs.clusterCfg.GetPeerstorePath()
-					peerManager := pstoremgr.New(ctx, host, peerstorePath)
+					peerManager := pstoremgr.New(context.Background(), nil, peerstorePath)
 
-					err = peerManager.ImportPeers(multiAddrs, false, peerstore.AddressTTL)
-					checkErr("importing peers into peerstore", err)
-					peerManager.SavePeerstoreForPeers(peers)
+					addrInfos, err := peer.AddrInfosFromP2pAddrs(multiAddrs...)
+					checkErr("getting address infos from peer multiaddresses", err)
+					peerManager.SavePeerstore(addrInfos)
 					out("peerstore written to %s\n", peerstorePath)
 				}
 
--- a/cmd/ipfs-cluster-service/main.go
+++ b/cmd/ipfs-cluster-service/main.go
@@ -299,9 +299,9 @@ remove the %s file first and clean any Raft state.
 				}
 
 				peersOpt := c.String("peers")
+				var multiAddrs []ma.Multiaddr
 				if peersOpt != "" {
 					addrs := strings.Split(peersOpt, ",")
-					multiAddrs := []ma.Multiaddr{}
 
 					for _, addr := range addrs {
 						addr = strings.TrimSpace(addr)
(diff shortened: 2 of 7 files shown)


cmd/ipfs-cluster-service/main.go
```diff
@@ -289,6 +299,25 @@ remove the %s file first and clean any Raft state.
...
 
+				peersOpt := c.String("peers")
+				multiAddrs := []ma.Multiaddr{}
+				var peers []peer.ID
+				if peersOpt != "" {
+					addrs := strings.Split(peersOpt, ",")
+
+					if len(addrs) > 0 {
+						for _, addr := range addrs {
+							multiAddr, err := ma.NewMultiaddr(strings.TrimSpace(addr))
```
kishansagathiya: I can do without trimming spaces, NewMultiaddr works well even if string contains spaces

lanzafame:
@kishansagathiya how is this different from passing several peers to the `ipfs-cluster-service daemon --bootstrap` flag?

kishansagathiya:
> how is this different from passing several peers to the ipfs-cluster-service daemon --bootstrap flag?

They are similar in some sense.

`init --peers` lets you add peers at the initialization stage without starting a daemon. And then without starting the daemon, you won't need to pass them with bootstrap.

cmd/ipfs-cluster-service/main.go
```diff
@@ -239,6 +245,10 @@ remove the %s file first and clean any Raft state.
 					Name:  "custom-secret, s",
 					Usage: "prompt for the cluster secret",
 				},
+				cli.StringFlag{
+					Name:  "peers, p",
```
hsanjuan: I think we can live without the short-form `p`.

cmd/ipfs-cluster-service/main.go
```diff
@@ -239,6 +245,10 @@ remove the %s file first and clean any Raft state.
 					Name:  "custom-secret, s",
 					Usage: "prompt for the cluster secret",
 				},
+				cli.StringFlag{
```
hsanjuan: There is a `SliceString` type for this. Saves you having to split things

cmd/ipfs-cluster-service/main.go
hsanjuan: then do without

cmd/ipfs-cluster-service/main.go
```diff
@@ -289,6 +299,25 @@ remove the %s file first and clean any Raft state.
...
+				peersOpt := c.String("peers")
+				multiAddrs := []ma.Multiaddr{}
+				var peers []peer.ID
+				if peersOpt != "" {
+					addrs := strings.Split(peersOpt, ",")
+
+					if len(addrs) > 0 {
+						for _, addr := range addrs {
+							multiAddr, err := ma.NewMultiaddr(strings.TrimSpace(addr))
+							checkErr("parsing multiaddress", err)
```
hsanjuan: `"parsing peer multiaddress: " + addr`

cmd/ipfs-cluster-service/main.go
```diff
@@ -304,6 +333,22 @@ remove the %s file first and clean any Raft state.
 					checkErr("saving "+DefaultIdentityFile, err)
 					out("new identity written to %s\n", identityPath)
 				}
+				if peersOpt != "" {
+					_, ident, cfgs := makeAndLoadConfigs()
+
+					ctx := context.Background()
+					host, _, _, err := ipfscluster.NewClusterHost(ctx, ident, cfgs.clusterCfg)
```
hsanjuan: This is an overkill approach (creating a fully configured libp2p peer).

You can just create a peerManager with a nil host and call `SavePeerstore` with the right `AddrInfo` objects. And you have this to help you: https://godoc.org/github.com/libp2p/go-libp2p-core/peer#AddrInfosFromP2pAddrs

cmd/ipfs-cluster-service/main.go
kishansagathiya: Yes, but it is not a user friendly option.

You have to use it like `ipfs-cluster init --peers <multiaddress1> --peers <multiaddress2>`
If you try `ipfs-cluster init --peers <multiaddress1>,<multiaddress2>`, it would assume `<multiaddress1>,<multiaddress2>` to be the first entry in the string slice

hsanjuan: I had completely missed that. Can you open an issue about this in the places that we use StringSlice?

kishansagathiya: Sure

cmd/ipfs-cluster-service/main.go
kishansagathiya: my bad, I was wrong. 
tried like this `ipfs-cluster-service init --peers "   <multiaddress1>,   <multiaddress>"`, which failed because it doesn't start with `/`.

cmd/ipfs-cluster-service/main.go
```diff
@@ -289,6 +298,25 @@ remove the %s file first and clean any Raft state.
 					cfgs.clusterCfg.Secret = userSecret
 				}
 
+				peersOpt := c.String("peers")
+				multiAddrs := []ma.Multiaddr{}
+				if peersOpt != "" {
+					addrs := strings.Split(peersOpt, ",")
```
hsanjuan: there is an if inside an if here, but you could do ` addrs := strings.Split(c.String("peers"), " ,")`  directly and have a single `if len(addrs) > 0` .

cmd/ipfs-cluster-service/main.go
```diff
@@ -304,6 +332,18 @@ remove the %s file first and clean any Raft state.
 					checkErr("saving "+DefaultIdentityFile, err)
 					out("new identity written to %s\n", identityPath)
 				}
+				if peersOpt != "" {
+					_, _, cfgs := makeAndLoadConfigs()
```
hsanjuan: you have a `cfgs` object already above with all the default things you need, why creating it again by forcing a re-read of all the config file?

In fact, this block can be moved up to the place the peers addresses and handled.

cmd/ipfs-cluster-service/main.go
hsanjuan: In fact you don't even need an `if len(addrs) > 0` at that point since the loop won't even run with 0 elements. You can put an if `len(peers) > 0` below and do the trusted_peer, init_peerset assignments and peerstore writing inside it.

cmd/ipfs-cluster-service/main.go
kishansagathiya: `cfgs` object that we have already would result `cfgs.clusterCfg.GetPeerstorePath()` in an empty string (because clusterConfig.BaseDir would be empty)

clusterConfig.BaseDir will be set in cfgMgr.LoadJSON

So, once the json file is saved and then when we load the configs again, `cfgs.clusterCfg.GetPeerstorePath()` would return the right path.

`makeAndLoadConfigs` also loads identity, so I think it is safe to keep it after we have save identity as well.

cmd/ipfs-cluster-service/main.go
kishansagathiya: but I can still remove that if condition

hsanjuan: true

cmd/ipfs-cluster-service/main.go
```diff
@@ -303,14 +303,14 @@ remove the %s file first and clean any Raft state.
...
-							multiAddrs = append(multiAddrs, multiAddr)
-						}
+					for _, addr := range addrs {
+						addr = strings.TrimSpace(addr)
+						multiAddr, err := ma.NewMultiaddr(addr)
+						checkErr("parsing peer multiaddress: "+addr, err)
+						multiAddrs = append(multiAddrs, multiAddr)
+					}
 
+					if len(addrs) > 0 {
```
hsanjuan: this will always be true then?

kishansagathiya: yes

cmd/ipfs-cluster-service/main.go
hsanjuan: then do a `cfgs.clusterCfg.SetBaseDir(filepath.Dir(configPath))`. Re-reading the configuration file and the identity from disk and re-parsing it is an overkill operation that you don't need.

hsanjuan:
I have rebased and pushed some changes (have a look). There are 3 problems remaining:

* Returning errors in SavePeerstore when they happen (I think they are warnings now because that function was only used during cluster shutdown which meant it was a best effort thing and in some cases like tests saving the peerstore was expected to fail).
* Erroring when the addrInfos have nothing but a peer id (`init -peers /ipfs/Qm...`  works and it should error because there is no addresses associated to the given peer ID)
* A sharness test checking the peerstore file was written with the expected contents.

kishansagathiya:
`pstoremgr.go` has few other functions as well that errors, where errors are logged but not returned(LoadPeerstore, Bootstrap). I think it would be safer to return errors here and ignore them during output (cmd and rest response)

hsanjuan:
> `pstoremgr.go` has few other functions as well that errors, where errors are logged but not returned(LoadPeerstore, Bootstrap). I think it would be safer to return errors here and ignore them during output (cmd and rest response)

For the moment just change the SavePeerstore* ones which are relevant to this PR.

sharness/t0021-service-init.sh
```diff
@@ -0,0 +1,32 @@
...
+'
+
+PEERS=/ip4/192.168.0.129/tcp/9196/ipfs/12D3KooWRN8KRjpyg9rsW2w7StbBRGper65psTZm68cjud9KAkaW,/ip4/192.168.0.129/tcp/9196/ipfs/12D3KooWPwrYNj7VficHw5qYidepMGA85756kYgMdNmRM9A1ZHjN
+
+test_expect_success "cluster-service init with --peers succeeds and fills peerstore" '
+    ipfs-cluster-service --config "test-config" init --peers $PEERS
+    [ -f "test-config/peerstore" ]
+    IFS=','
+    read -ra P <<< $PEERS
+    grep -q ${P[0]} "test-config/peerstore"
```
hsanjuan: This line is naught. The test result depends on the last command so whether this one passes or fails doesn't batter. Same on next test. Please create a file with the expected results and use `test_cmp` to compare the resulting peerstore with the pre-created file. `t0031-ct-add.sh` has examples.

Same problem in next test, in anything you need to `&&` tests.

sharness/t0021-service-init.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+test_description="Test init functionality"
+
+. lib/test-lib.sh
+test_ipfs_init
+
+test_expect_success "prerequisites" '
+    test_have_prereq IPFS
```
hsanjuan: Given these test do not start cluster, do we need to start ipfs at all?

pstoremgr/pstoremgr.go
```diff
@@ -243,41 +244,50 @@ func (pm *Manager) LoadPeerstore() (addrs []ma.Multiaddr) {
...
+			logger.Error(err)
+			return err
+		}
+
 		addrs, err := peer.AddrInfoToP2pAddrs(&pinfo)
 		if err != nil {
-			logger.Warning(err)
-			continue
+			logger.Error(err)
+			return err
```
hsanjuan: keep this as it was

pstoremgr/pstoremgr.go
```diff
@@ -243,41 +244,50 @@ func (pm *Manager) LoadPeerstore() (addrs []ma.Multiaddr) {
...
-		return
+		return err
 	}
 	defer f.Close()
 
 	for _, pinfo := range pinfos {
+		if len(pinfo.Addrs) == 0 {
+			err = errors.New("address info does not have any multiaddresses")
+			logger.Error(err)
+			return err
```
hsanjuan: don't abort the whole peerstore-writing operation because you found one peer without any multiaddresses, warn about it and move on. This code is used on different places and might be called with things straight from the peerstore which may have no addresses (because maybe they have expired).

You do want to error when reading the list from the flag specifically.

pstoremgr/pstoremgr.go
```diff
@@ -243,41 +244,50 @@ func (pm *Manager) LoadPeerstore() (addrs []ma.Multiaddr) {
 
 // SavePeerstore stores a slice of multiaddresses in the peerstore file, one
 // per line.
-func (pm *Manager) SavePeerstore(pinfos []peer.AddrInfo) {
+func (pm *Manager) SavePeerstore(pinfos []peer.AddrInfo) error {
 	if pm.peerstorePath == "" {
-		return
+		err := errors.New("empty peerstore path")
+		logger.Error(err)
+		return err
```
hsanjuan: No, why? This code is clearly saying that using an empty path in the peerstore is a no-op for Save. Don't change behaviour.

pstoremgr/pstoremgr.go
```diff
@@ -243,41 +244,50 @@ func (pm *Manager) LoadPeerstore() (addrs []ma.Multiaddr) {
...
+
 		addrs, err := peer.AddrInfoToP2pAddrs(&pinfo)
 		if err != nil {
-			logger.Warning(err)
-			continue
+			logger.Error(err)
+			return err
 		}
 		for _, a := range addrs {
 			f.Write([]byte(fmt.Sprintf("%s\n", a.String())))
```
hsanjuan: HERE, here is where you want to return an error if writing to the file failed

cluster.go
```diff
@@ -620,7 +620,10 @@ func (c *Cluster) Shutdown(ctx context.Context) error {
 	// Try to store peerset file for all known peers whatsoever
 	// if we got ready (otherwise, don't overwrite anything)
 	if c.readyB {
-		c.peerManager.SavePeerstoreForPeers(c.host.Peerstore().Peers())
+		err := c.peerManager.SavePeerstoreForPeers(c.host.Peerstore().Peers())
+		if err != nil {
+			logger.Warningf("could not save peers to peerstore: %s", err)
```
hsanjuan: double-logging, since you are logging a similar message already in the other place. leave unchanged, maybe with a comment saying we are ignoring errors because it's a best-effott.

hsanjuan:
>     * Returning errors in SavePeerstore when they happen (I think they are warnings now because that function was only used during cluster shutdown which meant it was a best effort thing and in some cases like tests saving the peerstore was expected to fail).

I should have been more specific: returning errors when writing the file errors.

hsanjuan:
#835

sharness/t0021-service-init.sh
```diff
@@ -3,30 +3,24 @@
...
-    grep -q ${P[0]} "test-config/peerstore"
-    grep -q ${P[1]} "test-config/peerstore"
+    PEER1=/ip4/192.168.0.129/tcp/9196/ipfs/12D3KooWRN8KRjpyg9rsW2w7StbBRGper65psTZm68cjud9KAkaW
+    PEER2=/ip4/192.168.0.129/tcp/9196/ipfs/12D3KooWPwrYNj7VficHw5qYidepMGA85756kYgMdNmRM9A1ZHjN
+    echo $PEER1 >> testPeerstore
+    echo $PEER2 >> testPeerstore
+    ipfs-cluster-service --config "test-config" init --peers $PEER1,$PEER2 &&
+    [ -f "test-config/peerstore" ] &&
+    test_cmp testPeerstore test-config/peerstore &&
+    rm testPeerstore
```
hsanjuan: `rm` not needed. sharness removes files when tests pass and leaves them for inspection otherwise. `[-f ".." ]` also not needed. concatenating with `&&` also not needed as long as the last line is test_cmp.

sharness/t0021-service-init.sh
```diff
@@ -3,30 +3,24 @@
...
 '
 
 test_expect_success "cluster-service init without --peers succeeds and creates empty peerstore" '
-    ipfs-cluster-service -f --config "test-config" init
-    [ -f "test-config/peerstore" ]
+    ipfs-cluster-service -f --config "test-config" init &&
+    [ -f "test-config/peerstore" ] &&
     [ -z `cat "test-config/peerstore"` ]
```
hsanjuan: https://stackoverflow.com/questions/9964823/how-to-check-if-a-file-is-empty-in-bash

`[ ! -s "test-config/peerstore" ]`

PR #843, opened 2019-07-19
Improve pin/unpin method signatures

hsanjuan:
These changes the following Cluster Go API methods:
    
    * -> Cluster.Pin(ctx, cid, options) (pin, error)
    * -> Cluster.Unpin(ctx, cid) (pin, error)
    * -> Cluster.PinPath(ctx, path, opts) (pin,error)
    
    Pin and Unpin now return the pinned object.
    
    The signature of the methods now matches that of the API Client, is clearer as
    to what options the user can set and is aligned with PinPath, UnpinPath, which
    returned pin objects.
    
    The REST API now returns the Pinned/Unpinned object rather than 204-Accepted.
    
    This was necessary for a cleaner pin/update approach, which I'm working on in
    another branch.
    
    Most of the changes here are updating tests to the new signatures

### DIFF
--- a/api/rest/client/lbclient_test.go
+++ b/api/rest/client/lbclient_test.go
@@ -34,9 +34,9 @@ func TestLBClient(t *testing.T) {
 
 	// With Failover strategy, it would go through first 5 empty clients
 	// and then 6th working client. Thus, all requests should always succeed.
-	testRunManyRequestsConcurrently(t, cfgs, &Failover{}, 1000, 6, true)
+	testRunManyRequestsConcurrently(t, cfgs, &Failover{}, 200, 6, true)
 	// First 5 clients are empty. Thus, all requests should fail.
-	testRunManyRequestsConcurrently(t, cfgs, &Failover{}, 1000, 5, false)
+	testRunManyRequestsConcurrently(t, cfgs, &Failover{}, 200, 5, false)
 }
 
 func testRunManyRequestsConcurrently(t *testing.T, cfgs []*Config, strategy LBStrategy, requests int, retries int, pass bool) {
--- a/rpc_api.go
+++ b/rpc_api.go
@@ -154,19 +154,32 @@ func (rpcapi *ClusterRPCAPI) ID(ctx context.Context, in struct{}, out *api.ID) e
 	return nil
 }
 
-// Pin runs Cluster.Pin().
-func (rpcapi *ClusterRPCAPI) Pin(ctx context.Context, in *api.Pin, out *struct{}) error {
-	return rpcapi.c.Pin(ctx, in)
+// Pin runs Cluster.pin().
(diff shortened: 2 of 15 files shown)


rpc_api.go
```diff
@@ -154,19 +154,32 @@ func (rpcapi *ClusterRPCAPI) ID(ctx context.Context, in struct{}, out *api.ID) e
...
 
-// Pin runs Cluster.Pin().
-func (rpcapi *ClusterRPCAPI) Pin(ctx context.Context, in *api.Pin, out *struct{}) error {
-	return rpcapi.c.Pin(ctx, in)
+// Pin runs Cluster.pin().
+func (rpcapi *ClusterRPCAPI) Pin(ctx context.Context, in *api.Pin, out *api.Pin) error {
+	// we do not call the Pin method directly since that method does not
+	// allow to pin other than regular DataType pins. The adder will
+	// however send Meta, Shard and ClusterDAG pins.
+	pin, _, err := rpcapi.c.pin(ctx, in, []peer.ID{})
```
kishansagathiya: Most cluster method names match with rpc method names. This confuses, because there is a `Pin` method in cluster.

Instead of this we can support all types in `cluster.Pin` method. cluster.Pin is used by cluster.PinPath. So we would have to set pin options ( and thus pin type) in pinPathHandler(rest api). It already happens for pinHandler.

api/rest/client/lbclient_test.go
```diff
@@ -34,9 +34,9 @@ func TestLBClient(t *testing.T) {
 
 	// With Failover strategy, it would go through first 5 empty clients
 	// and then 6th working client. Thus, all requests should always succeed.
-	testRunManyRequestsConcurrently(t, cfgs, &Failover{}, 1000, 6, true)
+	testRunManyRequestsConcurrently(t, cfgs, &Failover{}, 200, 6, true)
```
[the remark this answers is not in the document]
hsanjuan: yes

rpc_api.go
hsanjuan: pinType is not an option that should be set by the user. This is a good example where our internal API does not need to match our external one.

kishansagathiya: sorry, I didn't mean it to be set by user.
I meant like this
https://github.com/ipfs/ipfs-cluster/blob/b6ba67804fc261446d32807660751c93df516377/api/rest/restapi.go#L1032

hsanjuan: pinType is not a `PinOptions`. I'm not sure what you mean.

kishansagathiya: What I was proposing would require significant changes, PinPath rpc method accepting `types.Pin` instead of `types.PinPath` (in order to pass pin Type)

 We can let this be.

PR #846, opened 2019-07-23
When a peer is removed, clear it from peerstore 

### LINKED ISSUE #840: [Raft] peers rm doesn't actually remove the peer
**Additional information:**
 - OS: Linux
 - IPFS Cluster version: 0.11.0-rc4
 - Installation method: docker

**Describe the bug:**

Running `ipfs-cluster-ctl peers rm <peerid>` logs that it removes the peer but then attempts to connect that peer again. Logs:
```
06:33:28.758  INFO  pstoremgr: connecting to QmPu66JRV1CWHXVz7XJDXdkobjrPoyecs6fA2DhiswP2to pstoremgr.go:315
06:33:33.759 WARNI  pstoremgr: context deadline exceeded pstoremgr.go:318
06:33:45.067  INFO    cluster: re-allocating all CIDs directly associated to QmPu66JRV1CWHXVz7XJDXdkobjrPoyecs6fA2DhiswP2to cluster.go:810
06:33:45.073  INFO    cluster: Peer removed  QmPu66JRV1CWHXVz7XJDXdkobjrPoyecs6fA2DhiswP2to cluster.go:818
06:33:58.758  INFO  pstoremgr: connecting to QmPu66JRV1CWHXVz7XJDXdkobjrPoyecs6fA2DhiswP2to pstoremgr.go:315
06:34:03.760 WARNI  pstoremgr: context deadline exceeded pstoremgr.go:318
```

Note this peer doesn't exist anymore, and no other peer in the cluster thinks it is still around.

Normally, removing the peerstore after shutting the peer down would be a workaround to this, but I don't have access to the volume once the peer has been shutdown.

kishansagathiya:
With this PR, cluster peer will observe on events of peer removal
from cluster. On occurence of the event, the cluster peer will clear
the removed peer from its peerstore.

Fixes #840

### DIFF
--- a/consensus/raft/raft.go
+++ b/consensus/raft/raft.go
@@ -698,7 +698,7 @@ func (rw *raftWrapper) observePeers() {
 	obsCh := make(chan hraft.Observation, 1)
 	defer close(obsCh)
 
-	observer := hraft.NewObserver(obsCh, false, func(o *hraft.Observation) bool {
+	observer := hraft.NewObserver(obsCh, true, func(o *hraft.Observation) bool {
 		po, ok := o.Data.(hraft.PeerObservation)
 		return ok && po.Removed
 	})
--- a/consensus/raft/raft.go
+++ b/consensus/raft/raft.go
@@ -55,7 +55,7 @@ var maxShutdownSnapshotRetries = 5
 // Its methods provide functionality for working with Raft.
 type raftWrapper struct {
 	ctx           context.Context
-	cancel        func()
+	cancel        context.CancelFunc
 	raft          *hraft.Raft
 	config        *Config
 	host          host.Host
@@ -699,12 +699,8 @@ func (rw *raftWrapper) observePeers() {
 	defer close(obsCh)
 
 	observer := hraft.NewObserver(obsCh, false, func(o *hraft.Observation) bool {
-		switch o.Data.(type) {
-		case hraft.PeerObservation:
-			return true
-		default:
-			return false
-		}
+		po, ok := o.Data.(hraft.PeerObservation)
+		return ok && po.Removed
 	})
 
 	rw.raft.RegisterObserver(observer)
@@ -714,16 +710,13 @@ func (rw *raftWrapper) observePeers() {
 		select {
 		case obs := <-obsCh:
 			pObs := obs.Data.(hraft.PeerObservation)
-			if pObs.Removed {
-				logger.Infof("raft peer departed. Removing from peerstore: %s",
-					pObs.Peer.ID)
-				pID, err := peer.IDB58Decode(string(pObs.Peer.ID))
-				if err != nil {
-					logger.Error(err)
-					continue
-				}
-				rw.host.Peerstore().ClearAddrs(pID)
+			logger.Info("raft peer departed. Removing from peerstore: ", pObs.Peer.ID)
+			pID, err := peer.IDB58Decode(string(pObs.Peer.ID))
+			if err != nil {
+				logger.Error(err)
+				continue
 			}
+			rw.host.Peerstore().ClearAddrs(pID)
 		case <-rw.ctx.Done():
 			logger.Debug("stopped observing raft peers")
 			return
--- a/consensus/raft/raft.go
+++ b/consensus/raft/raft.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/ipfs/ipfs-cluster/pstoremgr"
 	"github.com/ipfs/ipfs-cluster/state"
 
 	host "github.com/libp2p/go-libp2p-core/host"
@@ -57,6 +58,7 @@ type raftWrapper struct {
 	raft          *hraft.Raft
 	config        *Config
 	host          host.Host
+	peerManager   *pstoremgr.Manager
 	serverConfig  hraft.Configuration
 	transport     *hraft.NetworkTransport
 	snapshotStore hraft.SnapshotStore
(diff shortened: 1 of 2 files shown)


consensus/raft/raft.go
```diff
@@ -686,3 +695,47 @@ func find(s []string, elem string) bool {
 	}
 	return false
 }
+
+func (rw *raftWrapper) observePeers() {
+	obsCh := make(chan hraft.Observation, 1)
+	defer close(obsCh)
+
+	if !sixtyfour {
+		// 32-bit systems don't support observers
```
kishansagathiya: I am not sure what was the issue here.

https://github.com/hashicorp/raft/blob/dcbf3052ba602706d2a0baa7c51a4f23c597dd7f/observer.go#L40-L43 makes me think that it is fixed
There is a PR with similar problem containing more details.

consensus/raft/raft.go
```diff
@@ -686,3 +695,47 @@ func find(s []string, elem string) bool {
...
+		// TODO: Check if this is still the case?
+		logger.Warning("raft Observers panic on 32-bit architecture")
+		return
+	}
+
+	observer := hraft.NewObserver(obsCh, false, nil)
```
hsanjuan: why not blocking?

hsanjuan: and why not filtering?

consensus/raft/raft.go
```diff
@@ -686,3 +695,47 @@ func find(s []string, elem string) bool {
...
+	defer rw.raft.DeregisterObserver(observer)
+
+	for {
+		select {
+		case obs := <-obsCh:
+			switch obs.Data.(type) {
+			case hraft.PeerObservation:
+				pObs := obs.Data.(hraft.PeerObservation)
+				if pObs.Removed {
+					logger.Infof("Observed removal of raft peer: %s",
```
hsanjuan: we have a tendency to start log messages with lowercase. (golint complains for Printlns). Message could be a bit more informative: `raft peer departed. Removing from peerstore`.

consensus/raft/raft.go
```diff
@@ -686,3 +695,47 @@ func find(s []string, elem string) bool {
...
+					}
+					err = rw.peerManager.RmPeer(pID)
+					if err != nil {
+						logger.Error(err)
+					}
+				}
+			default:
+			}
+		case <-rw.finishObs:
+			close(rw.finishObs)
```
hsanjuan: channels should be closed by the senders, never by the receivers. But you want to use a context anyways. Have the raftWrapper keep a context that can be cancelled on shutdown.

consensus/raft/raft.go
```diff
@@ -79,6 +82,7 @@ func newRaftWrapper(
 	raftW := &raftWrapper{}
 	raftW.config = cfg
 	raftW.host = host
+	raftW.peerManager = pstoremgr.New(context.Background(), host, "")
```
hsanjuan: unsure we need to carry this, if you just need a `host.Peerstore().ClearAddrs(pid)` inside the `observePeers()` function.

consensus/raft/raft.go
hsanjuan: Do you have a raspberry Pi you can test with?

kishansagathiya: No, but I can order one

hsanjuan: I think you can just build for 32 bit architecture and run it

kishansagathiya: I think it works. 

I commented this sixtyfour check.
installed cluster like this `env GOOS=linux GOARCH=386 make install`
ran a cluster with two peers, remove one of them with `peers rm`
other peer observed it successfully and removed it from peerstore, just like 64 bit.

consensus/raft/raft.go
kishansagathiya: filtering I have done after receiving observations, but I can change it to here.

Making blocking true will cause raft to block when sending observation. I dont see a strong reason to do so. Generally it should be false.
https://github.com/hashicorp/raft/blob/ff523e1c1515a5b33cc3153a34c984093fed06bb/observer.go#L50-L51

hsanjuan: > I dont see a strong reason to do so

Think again. If several peers are removed at once and Raft cannot put them in the channel, you will never get notified. Yes, there are strong reasons to block so you don't miss any observation. There are mild reasons to filter. I'm ok with no filtering.

consensus/raft/raft.go
```diff
@@ -700,41 +698,34 @@ func (rw *raftWrapper) observePeers() {
...
 	defer close(obsCh)
 
-	if !sixtyfour {
-		// 32-bit systems don't support observers
-		// TODO: Check if this is still the case?
-		logger.Warning("raft Observers panic on 32-bit architecture")
-		return
-	}
+	observer := hraft.NewObserver(obsCh, false, func(o *hraft.Observation) bool {
+		switch o.Data.(type) {
```
hsanjuan:
```go
po,ok := o.Data.(*hraft.Observation)
return ok && po.Removed
```

pstoremgr/pstoremgr.go
```diff
@@ -312,14 +312,14 @@ func (pm *Manager) Bootstrap(count int) []peer.ID {
 			continue
 		}
 
-		logger.Debugf("connecting to %s", pinfo.ID)
```
hsanjuan: No reason to change these and below. They were more verbose in the past and I had to tune it down because it seriously spams the logs on big clusters (and not being able to connect some peer during bootstrap is not a big deal).

Issue #841, opened 2019-07-18
Use `StringFlag` instead of `StringSliceFlag`


kishansagathiya:
`StringSliceFlag`  is not an user friendly option.

You have to use it like `ipfs-cluster init --peers <multiaddress1> --peers <multiaddress2>`
If you try `ipfs-cluster init --peers <multiaddress1>,<multiaddress2>`, it would assume `<multiaddress1>,<multiaddress2>` to be the first entry in the string slice

Use  `StringFlag` instead and split string with `,`

There are two occurrences of this flag https://github.com/ipfs/ipfs-cluster/search?q=StringSliceFlag&unscoped_q=StringSliceFlag

PR #851, opened 2019-07-24
refer #841: change StringSliceFlag to StringFlag

chenminjian:
License: MIT
Signed-off-by: chenminjian <727180553@qq.com>

### DIFF
--- a/cmd/ipfs-cluster-service/daemon.go
+++ b/cmd/ipfs-cluster-service/daemon.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cmd/ipfs-cluster-service/daemon.go
```diff
@@ -36,8 +37,9 @@ import (
 	cli "github.com/urfave/cli"
 )
 
-func parseBootstraps(flagVal []string) (bootstraps []ma.Multiaddr) {
-	for _, a := range flagVal {
+func parseBootstraps(flagVal string) (bootstraps []ma.Multiaddr) {
+	addrs := strings.Split(flagVal, ",")
+	for _, a := range addrs {
```
hsanjuan: If the flag is empty, addrs will always be an slice of length 1, with an empty string. Thus you need to skip empty or check `flagVal` as more than 0 length.

hsanjuan:
Fixed in #876

PR #856, opened 2019-07-26
service: Make --consensus a mandatory flag

hsanjuan:
It has a few implications to launch a raft peer when you wanted to do crdt and vice-versa.

Same when exporting and exporting states.

Users starting cluster peers should be explicit about their consensus choice.

Also, if we ever want to make `crdt` the default, we can't do that before
making `raft` non-default first. I don't like to break things but otherwise
the experience for new users wanting to try crdts might be aweful.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,4 +56,4 @@ VOLUME $IPFS_CLUSTER_PATH
 ENTRYPOINT ["/sbin/tini", "--", "/usr/local/bin/entrypoint.sh"]
 
 # Defaults for ipfs-cluster-service go here
-CMD ["daemon", "--upgrade"]
+CMD ["daemon", "--consensus raft"]
--- a/cmd/ipfs-cluster-service/daemon.go
+++ b/cmd/ipfs-cluster-service/daemon.go
@@ -47,6 +47,10 @@ func parseBootstraps(flagVal []string) (bootstraps []ma.Multiaddr) {
 
 // Runs the cluster peer
 func daemon(c *cli.Context) error {
+	if c.String("consensus") == "" {
+		checkErr("starting daemon", errors.New("--consensus flag must be set to \"raft\" or \"crdt\""))
+	}
+
 	logger.Info("Initializing. For verbose output run with \"-l debug\". Please wait...")
 
 	ctx, cancel := context.WithCancel(context.Background())
--- a/cmd/ipfs-cluster-service/main.go
+++ b/cmd/ipfs-cluster-service/main.go
@@ -28,7 +28,6 @@ const programName = `ipfs-cluster-service`
 
 // flag defaults
 const (
-	defaultConsensus  = "raft"
 	defaultAllocation = "disk-freespace"
 	defaultPinTracker = "map"
 	defaultLogLevel   = "info"
@@ -369,7 +368,6 @@ multiaddresses.
 				},
 				cli.StringFlag{
 					Name:  "consensus",
-					Value: defaultConsensus,
 					Usage: "shared state management provider [raft,crdt]",
 				},
 				cli.StringFlag{
@@ -414,13 +412,16 @@ By default, the state will be printed to stdout.
 						},
 						cli.StringFlag{
 							Name:  "consensus",
-							Value: "raft",
 							Usage: "consensus component to export data from [raft, crdt]",
 						},
 					},
 					Action: func(c *cli.Context) error {
 						locker.lock()
 						defer locker.tryUnlock()
+
+						cfgMgr, ident, cfgs := makeAndLoadConfigs()
+						defer cfgMgr.Shutdown()
+						mgr := newStateManager(c.String("consensus"), ident, cfgs)
 
 						var w io.WriteCloser
 						var err error
@@ -435,9 +436,6 @@ By default, the state will be printed to stdout.
 						}
 						defer w.Close()
 
-						cfgMgr, ident, cfgs := makeAndLoadConfigs()
-						defer cfgMgr.Shutdown()
-						mgr := newStateManager(c.String("consensus"), ident, cfgs)
 						checkErr("exporting state", mgr.ExportState(w))
 						logger.Info("state successfully exported")
 						return nil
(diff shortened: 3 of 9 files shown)


Dockerfile
```diff
@@ -56,4 +56,4 @@ VOLUME $IPFS_CLUSTER_PATH
 ENTRYPOINT ["/sbin/tini", "--", "/usr/local/bin/entrypoint.sh"]
 
 # Defaults for ipfs-cluster-service go here
-CMD ["daemon", "--upgrade"]
+CMD ["daemon", "--consensus raft"]
```
hsanjuan: `--upgrade` was deprecated btw.

cmd/ipfs-cluster-service/daemon.go
```diff
@@ -47,6 +47,10 @@ func parseBootstraps(flagVal []string) (bootstraps []ma.Multiaddr) {
 
 // Runs the cluster peer
 func daemon(c *cli.Context) error {
+	if c.String("consensus") == "" {
+		checkErr("starting daemon", errors.New("--consensus flag must be set to \"raft\" or \"crdt\""))
```
hsanjuan: This would fail anyways later down, but I prefer to give a very clear error message.

cmd/ipfs-cluster-service/main.go
```diff
@@ -414,14 +412,17 @@ By default, the state will be printed to stdout.
...
 							Usage: "consensus component to export data from [raft, crdt]",
 						},
 					},
 					Action: func(c *cli.Context) error {
 						locker.lock()
 						defer locker.tryUnlock()
 
+						cfgMgr, ident, cfgs := makeAndLoadConfigs()
+						defer cfgMgr.Shutdown()
+						mgr := newStateManager(c.String("consensus"), ident, cfgs)
```
hsanjuan: Allow to fail before creating output files.

sharness/lib/test-lib.sh
```diff
@@ -121,7 +121,7 @@ cluster_kill(){
 }
 
 cluster_start(){
-    ipfs-cluster-service --config "test-config" daemon >"$IPFS_OUTPUT" 2>&1 &
+    ipfs-cluster-service --config "test-config" daemon --consensus crdt >"$IPFS_OUTPUT" 2>&1 &
```
lanzafame: Does this mean that we are no longer performing the sharness tests with `raft`? If we are maintaining that both `crdt` and `raft` are viable consensus components, then we should continue to test both as fully as we can?

docker-compose.yml
```diff
@@ -82,14 +82,14 @@ services:
     # to it.
     command: >-
       sh -c '
-        cmd="daemon --upgrade"
+        cmd="daemon --consensus raft"
```
lanzafame: Shouldn't we make this an environment variable?

lanzafame: The consensus component choice, that is.

sharness/lib/test-lib.sh
hsanjuan: sharness tests run on single-peer so it does not matter much. I agree the export/import tests should be done with those, as that's actually something sharning IS testing and has two distinct codepaths.

docker-compose.yml
hsanjuan: mm no, other component-selection flags are not configurable via env-var.

hsanjuan: technically, for docker containers, it should be mandatory to pass `--consensus` when doing docker run. That should be possible for every tool that wants to start a docker container though (even if they also configure other things with env-vars).

hsanjuan: I'll deal with docker in other PR, I just wanted to maintain behaviour "as is" on this one.

hsanjuan:
> LGTM, but comments
> 
>     * Dockerfile, docker-compose, docker/cluster-restart.sh uses raft, while sharness now uses crdt. Any reason or is it random?

Read convo on lanzafame's comment.

> 
>     * We can have consensus flag. If we can make this choice part of the configuration, I think it would be more user friendly. The scenario where, peers are initializing using some other cluster peer's configuration would also work well, as peers in the same cluster would need same consensus.

We could autoselect if only `raft` or only `crdt` is present in the config. But other than that, component selection is a daemon feature while configuration file configures components (not how to start the cluster). I'm a bit lost between "making it easy" and adding "yet another way of doing the same thing" (which needs documentation alignment).

kishansagathiya:
> We could autoselect if only raft or only crdt is present in the config.

Yes

> But other than that, component selection is a daemon feature 

Why? It is not like consensus can change every time daemon run. It can, but changed consensus would not be same as other peers and it would fail. So, it seems like something we can set during initialization.

hsanjuan:
> Why? It is not like consensus can change every time daemon run. It can, but changed consensus would not be same as other peers and it would fail. So, it seems like something we can set during initialization.

Because the configuration is configuring components. The daemon is putting a choice of components to launch the main component. There is no place in the configuration to say that "X" component should be used because it's configuring the components themselves, not the daemon. What we can do is to let an env variable set the daemon flag, or to make a polite guess based on whether the configuration contains configuration for a component or not and take that as the hint.

kishansagathiya:
> There is no place in the configuration to say that "X" component should be used because it's configuring the components themselves

We don't need a config element for that

> or to make a polite guess based on whether the configuration contains configuration for a component or not and take that as the hint.

^ We can use this. We can ask during init and fill up configuration accordingly. If `service init --consensus raft`, fill up raft config and keep crdt config empty. If user fills both for some reason, we use have a default consensus. But I see filling both configs as a human error, since peers/cluster are not meant to change consensus.

hsanjuan:
> ^ We can use this. We can ask during init and fill up configuration accordingly. If `service init --consensus raft`, fill up raft config and keep crdt config empty. If user fills both for some reason, we use have a default consensus. But I see filling both configs as a human error, since peers/cluster are not meant to change consensus.

Yes, this is a good point/idea. The flag would be mandatory for `init`. But not for `daemon`. And that's actually backwards compatible, because old configs only have `raft` section. Can you open an issue and work on this?

PR #857, opened 2019-07-27
Auto-trust bootstrap peers

### LINKED ISSUE #834: [Daemon] Auto-trust bootstrap peers

Peers what we bootstrap with (`--bootstrap`) should be 1) trusted in consensus 2) Added to crdt trusted_peers when using crdts and saved in the config, unless a `--no-trust` flag is set.

**Additional context**

It seems the expected flow is that when you bootstrap to a peer, you implicitally trust it.

kishansagathiya:
Add bootstrap peers to list of trusted peers for crdt.
Just adding them to crdt trusted_peers in config is enough.
 
Fixes #834

### DIFF
--- a/cmd/ipfs-cluster-service/daemon.go
+++ b/cmd/ipfs-cluster-service/daemon.go
@@ -62,6 +62,10 @@ func daemon(c *cli.Context) error {
 
 	defer cfgMgr.Shutdown()
 
+	if len(bootstraps) > 0 && !c.Bool("no-trust") {
+		cfgs.crdtCfg.TrustedPeers = append(cfgs.crdtCfg.TrustedPeers, ipfscluster.PeersFromMultiaddrs(bootstraps)...)
+	}
+
 	if c.Bool("stats") {
 		cfgs.metricsCfg.EnableStats = true
 	}
--- a/cmd/ipfs-cluster-service/main.go
+++ b/cmd/ipfs-cluster-service/main.go
@@ -393,7 +393,7 @@ multiaddresses.
 				},
 				cli.BoolFlag{
 					Name:  "no-trust",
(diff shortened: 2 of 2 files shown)


kishansagathiya:
Worth adding a sharness test by running a two peer cluster?

cmd/ipfs-cluster-service/main.go
```diff
@@ -391,6 +391,10 @@ multiaddresses.
 					Name:  "tracing",
 					Usage: "enable tracing collection",
 				},
+				cli.BoolFlag{
+					Name:  "no-trust",
+					Usage: "do not trust bootstrap peers",
```
hsanjuan:
```suggestion
					Usage: "do not trust bootstrap peers (only for \"crdt\" consensus",
```

cmd/ipfs-cluster-service/daemon.go
```diff
@@ -62,6 +62,10 @@ func daemon(c *cli.Context) error {
 
 	defer cfgMgr.Shutdown()
 
+	if len(bootstraps) > 0 && !c.Bool("no-trust") {
```
hsanjuan: The first condition is not necessary but it's ok like this

cmd/ipfs-cluster-service/main.go
```diff
@@ -391,6 +391,10 @@ multiaddresses.
 					Name:  "tracing",
 					Usage: "enable tracing collection",
 				},
+				cli.BoolFlag{
+					Name:  "no-trust",
+					Usage: "do not trust bootstrap peers (only for \"crdt\" consensus",
```
hsanjuan:
```suggestion
					Usage: "do not trust bootstrap peers (only for \"crdt\" consensus)",
```

Issue #211, opened 2017-10-27
Update ipfs-cluster to make use of hRaft LeaderObservation type 


ZenGround0:
Track the reversion of commit b7e8a11 in hashicorp raft referenced here https://github.com/hashicorp/raft/issues/254 and update consensus/raft/raft.go to make use of observers in the function WaitForLeader.

lanzafame:
If I understand correctly, this is no longer blocked by hashicorp/raft#254 as of hashicorp/raft#259.

@hsanjuan is this as simple as uncommenting the code in `raftWrapper.WaitForLeader`?

hsanjuan:
It's blocked because it requires a prior update of the gxed raft, which involves updating some deps below (since we're at it). And `raft` hasn't had an official `tag` since 1.0.0, that includes this change. Since generally things work on this front, I'm not in a hurry to do this.

hsanjuan:
wop, a new Raft release is out and this is not blocked anymore after #769

kishansagathiya:
Using LeaderObservation does not achieve any extra good results here, just creates problems.

Currently we waitForLeader when a new peer gets added. If you use LeaderObservation type, you might get stuck. Leader could get elected before even you register it in WaitForLeader, in which case you will never observe it. So, if using `LeaderObservation` type, it has to go in a goroutine like `observePeers`

But even then, it is likely that leader will remain same after a new peer is added, in which case you will never observe a new leader and are stuck(actually context will timeout) at `WaitForLeader`.

So, it appears to me that if we were to use it, we require some design changes that code.

hsanjuan:
This is just removing the 32-bit safeguard and uncommenting a few lines

hsanjuan:
Keep the `ticker`

PR #858, opened 2019-07-29
Make use of hRaft LeaderObservation type

kishansagathiya:
This is mainly to illustrate https://github.com/ipfs/ipfs-cluster/issues/211#issuecomment-515831913

### DIFF
--- a/consensus/raft/raft.go
+++ b/consensus/raft/raft.go
@@ -38,11 +38,6 @@ var RaftMaxSnapshots = 5
 // This is used to reduce disk I/O for the recently committed entries.
 var RaftLogCacheSize = 512
 
-// Are we compiled on a 64-bit architecture?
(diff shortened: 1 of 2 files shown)


consensus/raft/raft.go
```diff
@@ -260,25 +255,9 @@ func (rw *raftWrapper) WaitForLeader(ctx context.Context) (string, error) {
...
 	for {
 		select {
-		case obs := <-obsCh:
-			_ = obs
-			// See https://github.com/hashicorp/raft/issues/254
-			// switch obs.Data.(type) {
-			// case hraft.LeaderObservation:
-			// 	lObs := obs.Data.(hraft.LeaderObservation)
-			// 	logger.Infof("Raft Leader elected: %s",
-			// 		lObs.Leader)
```
kishansagathiya: `Leader` attribute is private now.

PR #861, opened 2019-07-29
Avoid recontacting peers that never authorized us

### LINKED ISSUE #849: Avoid recontacting peers that never authorized us

If making an RPC request to a peer results in Authorization denied, we should avoid recontacting that peer on subsequent requests. whyrusleeing/timecache may serve as example: if we get an authorization error, we put the peer in it for a while. Before making a request we check if a peer is in the timecache and, if so, skip it.

Possibly worth to allow a hidden configuration here, with 0 to disable. But the default time would be quite high.

**Additional context**

Right now authorization errors are ignored when doing things like Status and other broadcast calls and we simply ignore those answers in the final result. Repeteadly contacting those peers on subsequent requests only makes things slower (peers are very unlikely to have "allowed" this peer).

Along with this we might choose to show authorization errors when they first happen, explaining that we will not try to recontact that peer for X hours.

Initially, this would apply only for broadcast calls in `cluster.go`.

kishansagathiya:
This commit introduces an lru cache that keeps entries valid till
specified duration.

When an rpc request fails due to authorization error, it(peer, service
name, service method) will be saved in the cache. We would avoid making
that request for some specified near future, since peers are unlikely to
change authorization policies frequently.
This duration can be set in cluster config as `authorization_cache_span`

So far, this limits only to rpc broadcast calls (MultiCall)

Fixes #849


timecache.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"time"
+
+	peer "github.com/libp2p/go-libp2p-core/peer"
+)
+
+// TimeCache is an lru cache where entries are valid upto a specified
+// duration.
+type TimeCache struct {
```
kishansagathiya: I have not bothered to add a mechanism to clean this cache at some interval. Worth doing so, may be for large clusters?

cluster.go
```diff
@@ -43,6 +43,74 @@ const (
...
+
+	err := rpcW.rpcClient.CallContext(ctx, dest, svcName, svcMethod, args, reply)
+	if err != nil && err.Error() == authErr {
+		rpcW.tc.Add(dest, svcName+svcMethod)
+	}
+
+	return err
+}
+
+func (rpcW *rpcClientWrapper) MultiCall(
```
kishansagathiya: Same as in rpc library

timecache.go
hsanjuan: Use https://github.com/whyrusleeping/timecache/

cluster.go
```diff
@@ -43,6 +43,74 @@ const (
 	reBootstrapInterval = 30 * time.Second
 )
 
+type rpcClientWrapper struct {
+	rpcClient *rpc.Client
+	tc        *TimeCache
+}
+
+func (rpcW *rpcClientWrapper) CallContext(
```
hsanjuan: I asked for A wrapper for the Multicall function, not for the rpc client.

cluster.go
hsanjuan: Do not re-implement multicall. This is not a wrapper. This is a re-implementation.

timecache.go
kishansagathiya: This is taken from `whyrusleeping/timecache` and modified for our requirements.

cluster.go
kishansagathiya: `MultiCall` uses wrapped `CallContext`

timecache.go
hsanjuan: I don't see any similarities between this and whyrusleeping/timecache. It's worse by forgetting to delete things when they expire. Plus there is no need to modify anything for our requirements. Note the fact that one method was not authorized implies any other method will also not be authorized for the kind of rpc requests we're dealing with.

cluster.go
hsanjuan: Exactly, that's why it is re-implementing rpc.Multicall rather than wrapping it.

timecache.go
kishansagathiya: By ` kind of rpc requests` you mean service name?

hsanjuan: I mean you are using service name for the timecache and that's not necessary.

kishansagathiya:
I have messed up commits, will have to force push

timecache.go
kishansagathiya: Now it uses whyrusleeping/timecache directly. Uses only peer id for timecache( if one request is not allowed, all wil be considered not allowed.)

cluster.go
kishansagathiya: I have dropped the struct. It's a cluster method now.

cluster.go
```diff
@@ -59,6 +60,8 @@ type Cluster struct {
 	rpcClient   *rpc.Client
 	peerManager *pstoremgr.Manager
 
+	tc *timecache.TimeCache
```
hsanjuan: meaningful name: `unauthPeerTimeCache`

cluster.go
```diff
@@ -1738,3 +1743,50 @@ func diffPeers(peers1, peers2 []peer.ID) (added, removed []peer.ID) {
...
+	if !checkMatchingLengths(len(ctxs), len(dests), len(replies)) {
+		panic("ctxs, dests and replies must match in length")
+	}
+
+	errs := make([]error, len(dests))
+	var actualCtxs []context.Context
+	var actualDests []peer.ID
+	var actualReplies []interface{}
+
+	authErr := fmt.Sprintf("client does not have permissions to this method, service name: %s, method name: %s", svcName, svcMethod)
```
hsanjuan: make this a global variable at the top. Let it say that it is skipping because it was not authorized before.

hsanjuan: Like the message below. Avoid returning one error to the user and logging a different one.

cluster.go
```diff
@@ -1738,3 +1743,50 @@ func diffPeers(peers1, peers2 []peer.ID) (added, removed []peer.ID) {
...
+
+	actualErrors := c.rpcClient.MultiCall(actualCtxs, actualDests, svcName, svcMethod, args, actualReplies)
+
+	k := 0
+	for i, e := range errs {
+		if e != nil {
+			continue
+		}
+
+		if actualErrors[k] != nil && actualErrors[k].Error() == authErr {
```
hsanjuan: `rpc.IsAuthorizationError(...)`

cluster.go
```diff
@@ -1738,3 +1743,50 @@ func diffPeers(peers1, peers2 []peer.ID) (added, removed []peer.ID) {
...
+			c.tc.Add(peer.IDB58Encode(dests[i]))
```
hsanjuan: You can add `string(dests[i])` directly, probably.

cluster_config.go
```diff
@@ -23,19 +23,20 @@ const configKey = "cluster"
...
+	DefaultMonitorPingInterval    = 15 * time.Second
+	DefaultPeerWatchInterval      = 5 * time.Second
+	DefaultReplicationFactor      = -1
+	DefaultLeaveOnShutdown        = false
+	DefaultDisableRepinning       = false
+	DefaultPeerstoreFile          = "peerstore"
+	DefaultConnMgrHighWater       = 400
+	DefaultConnMgrLowWater        = 100
+	DefaultConnMgrGracePeriod     = 2 * time.Minute
+	DefaultAuthorizationCacheSpan = 48 * time.Hour
```
hsanjuan: More like 30 seconds

cluster_config.go
```diff
@@ -134,6 +135,11 @@ type Config struct {
 	// libp2p host peerstore addresses. This file is regularly saved.
 	PeerstoreFile string
 
+	// AuthorizationCacheSpan is the duration upto which a peer will be
+	// considered to be not trusting current peer, if an rpc request failed
+	// because of authorization error.
+	AuthorizationCacheSpan time.Duration
```
hsanjuan: This variable has a bad name. Can you understand what it actually does from the current name and explanation?

`UnauthorizedRequestCacheSpan` would be slightly better. With a description saying that certain rpc requests are not performed to things in this cache.

cluster.go
kishansagathiya: The message is same as it would have been be returned by server if request were sent. I think this similarity will benefit us while checking authorization errors (will save us from checking an extra separate error message)

> Let it say that it is skipping because it was not authorized before.
We are skipping, is more of a warning than error (since we ourselves are doing it)

Should I still change?

hsanjuan: Never check if an error is certain type of error by checking the string (if you can). You are making assumptions that break sooner or later.  The library gives you a way to perform the check. `rpc.IsAuthorizationError(...)`. And you are lying to the user, the request did not fail because it was not authorized now, but because you did not perform it based on a previous failure.

cluster.go
```diff
@@ -44,6 +44,21 @@ const (
 	reBootstrapInterval = 30 * time.Second
 )
 
+var authMsg = "avoiding request to this peer since it unauthorized recently"
+
+type skipErr struct {
```
hsanjuan: I don't know what this is. Remove?

cluster.go
```diff
@@ -44,6 +44,21 @@ const (
 	reBootstrapInterval = 30 * time.Second
 )
 
+var authMsg = "avoiding request to this peer since it unauthorized recently"
```
hsanjuan:
```go
var (
    errUnauthPeerCache = errors.New(...)
)
```

cluster.go
```diff
@@ -1479,7 +1494,7 @@ func (c *Cluster) Peers(ctx context.Context) []*api.ID {
 			continue
 		}
 
-		if rpc.IsAuthorizationError(err) {
+		if rpc.IsAuthorizationError(err) || isSkipErr(err) {
```
hsanjuan: not here. we're not touching Peers

cluster.go
```diff
@@ -1760,12 +1775,9 @@ func (c *Cluster) multiCallWrapper(
...
 	var actualReplies []interface{}
 
-	authErr := fmt.Sprintf("client does not have permissions to this method, service name: %s, method name: %s", svcName, svcMethod)
-
 	for i, p := range dests {
-		if c.tc.Has(peer.IDB58Encode(p)) {
-			logger.Warningf("avoiding the request since it was unauthorized recently: peer %s, service name: %s, method name: %s", dests[i], svcName, svcMethod)
-			errs[i] = errors.New(authErr)
+		if c.unauthPeerTimeCache.Has(string(p)) {
+			errs[i] = fmt.Errorf("%s, peer: %s", authMsg, dests[i])
```
hsanjuan: the `isSkipError` will likely not work because you are not creating a skip error here.

`errs[i] = errUnauthPeerCache`

cluster_config.go
```diff
@@ -135,10 +135,11 @@ type Config struct {
 	// libp2p host peerstore addresses. This file is regularly saved.
 	PeerstoreFile string
 
-	// AuthorizationCacheSpan is the duration upto which a peer will be
+	// UnauthorizedRequestCacheSpan is the duration upto which a peer will be
 	// considered to be not trusting current peer, if an rpc request failed
-	// because of authorization error.
-	AuthorizationCacheSpan time.Duration
+	// because of authorization error. If a peer has unauthorized us, we won't
+	// make an rpc request to it until this duration.
```
hsanjuan: Check the grammar in the whole sentence.

cluster.go
```diff
@@ -44,6 +44,21 @@ const (
...
+
+type skipErr struct {
+	peer peer.ID
+}
+
+func (s *skipErr) Error() string {
+	return fmt.Sprintf("%s, peer: %s", authMsg, s.peer)
+}
+
+func isSkipErr(e error) bool {
```
hsanjuan: `return err == errUnauthPeerCache`

cluster_config.go
```diff
@@ -135,10 +135,8 @@ type Config struct {
 	// libp2p host peerstore addresses. This file is regularly saved.
 	PeerstoreFile string
 
-	// UnauthorizedRequestCacheSpan is the duration upto which a peer will be
-	// considered to be not trusting current peer, if an rpc request failed
-	// because of authorization error. If a peer has unauthorized us, we won't
-	// make an rpc request to it until this duration.
+	// UnauthorizedRequestCacheSpan is the duration upto which keys would be
+	// kept in unauthPeerTimeCache.
```
hsanjuan: instead of fixing grammar you have removed the explanation...

cluster.go
```diff
@@ -175,6 +180,8 @@ func NewCluster(
 	}
 	c.setupRPCClients()
 
+	c.unauthPeerTimeCache = timecache.NewTimeCache(c.config.UnauthorizedRequestCacheSpan)
```
hsanjuan: move this up. Can initialize directly in the cluster object. By doing it here you open a race condition where the API can send requests but the cache is still nil.

cluster_config.go
hsanjuan: sorry. I was looking at the older commit

hsanjuan:
Don't work on tests yet, I will leave this unmerged until we fully approach the collaborative clusters usecase. What it gets overlaps with follower_mode.

PR #863, opened 2019-07-30
Tests: AddLocal, increase delay

hsanjuan:
The only way I can reproduce this failure is to reduce the delay.


hsanjuan:
Tests are failing for other reasons now at least

PR #864, opened 2019-07-30
Fix #803: Add "follower_mode" to the config

### LINKED ISSUE #803: [CRDT] Untrusted Peers can remove pins and this causes weird situations
Untrusted peers can gain control of pins if they are the only ones allocated for that pin.
Forced example
```
$ ipfs-cluster-ctl -l <trusted_peer> pin add -r 1 --allocs <untrusted_peer> <hash>
$ ipfs-cluster-ctl -l <untrusted_peer> pin rm <hash>
# works!!!
$ ipfs-cluster-ctl -l <trusted_peer> pin ls
<hash> |  | PIN | Repl. Factor: 1--1 | Allocations: [<untrusted_peer>] | Recursive
# trusted peers still think it is around
```

This is an issue in the case of a cluster that has more untrusted peers than trusted peers, which is the model we are going for.
In a cluster that has a default replication of 3, uses the freespace allocator, has 10 peers, 4 trusted, 6 untrusted, once 3 of the untrusted peers have more freespace than the trusted peers they will get allocated all the pins and will have full control over those pins, with the trusted peers having no idea whether the pin is actually on the untrusted peer. 

//cc @hsanjuan we need to figure out how we are going to approach this scenario

hsanjuan:
Peers configured with follower_mode = true fail to add/pin/unpin.

Additionally they do not contact other peers when doing Status, Sync or
Recover and report on themselves.

They still contact other peers when doing "peers ls", as this is an OpenRPC
endpoint.

This is merely improving user interaction with a cluster peer and avoids
getting into confusing places:

* pin/unpin seems to work even no one trusts them
* status will query all peers in the peerset only to get auth errors and
ignore them, becoming way slower than it could be

This is not a security feature.

#803

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -43,6 +43,10 @@ const (
 	reBootstrapInterval = 30 * time.Second
 )
 
+var (
+	errFollowerMode = errors.New("this peer is configured as in follower mode. Write operations are disabled")
+)
+
 // Cluster is the main IPFS cluster component. It provides
 // the go-API for it and orchestrates the components that make up the system.
 type Cluster struct {
@@ -1183,6 +1187,7 @@ func (c *Cluster) PinGet(ctx context.Context, h cid.Cid) (*api.Pin, error) {
 func (c *Cluster) Pin(ctx context.Context, h cid.Cid, opts api.PinOptions) (*api.Pin, error) {
 	_, span := trace.StartSpan(ctx, "cluster/Pin")
 	defer span.End()
+
 	ctx = trace.NewContext(c.ctx, span)
 	pin := api.PinWithOpts(h, opts)
 	result, _, err := c.pin(ctx, pin, []peer.ID{})
@@ -1283,6 +1288,10 @@ func (c *Cluster) pin(
 	ctx, span := trace.StartSpan(ctx, "cluster/pin")
 	defer span.End()
 
+	if c.config.FollowerMode {
+		return nil, false, errFollowerMode
+	}
+
 	if pin.Cid == cid.Undef {
 		return pin, false, errors.New("bad pin object")
 	}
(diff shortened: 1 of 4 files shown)


hsanjuan:
I should have left this as Draft PR. I put `FollowerMode` in the `cluster` config to use it avoid doing broadcasts and prevent peers from trying to add things, but this increases complexity vs putting it as part of the `crdt` component.

We could have an additional `BroadcastPeers` method in the crdt to use for broadcasts except `peers()` but that seems a bit weird too.

kishansagathiya:
Is this ready for review?

hsanjuan:
now yes

cluster.go
```diff
@@ -43,6 +43,10 @@ const (
 	reBootstrapInterval = 30 * time.Second
 )
 
+var (
+	errFollowerMode = errors.New("this peer is configured as in follower mode. Write operations are disabled")
```
kishansagathiya:
```suggestion
	errFollowerMode = errors.New("this peer is configured to be in follower mode. Write operations are disabled")
```

ipfscluster_test.go
```diff
@@ -1897,3 +1898,84 @@ func TestClustersDisabledRepinning(t *testing.T) {
...
+		}
+	})
+
+	t.Run("follower status itself only", func(t *testing.T) {
+		gpi, err := clusters[1].Status(ctx, test.Cid1)
+		if err != nil {
+			t.Error("status should work")
+		}
+		if len(gpi.PeerMap) != 1 {
+			t.Fatal("globalPinInfo[0] should only have one peer")
```
kishansagathiya:
```suggestion
			t.Fatal("globalPinInfo should only have one peer")
```

ipfscluster_test.go
```diff
@@ -1897,3 +1898,84 @@ func TestClustersDisabledRepinning(t *testing.T) {
...
+	defer shutdownClusters(t, clusters, mock)
+
+	_, err := clusters[0].Pin(ctx, test.Cid1, api.PinOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = clusters[0].Pin(ctx, test.ErrorCid, api.PinOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
```
kishansagathiya: don't really need above two tests.

hsanjuan: yes, the above is needed so the `sync` test below works.

PR #867, opened 2019-08-05
CRDT: TrustAll by default.

hsanjuan:


kishansagathiya:
Why do we need this?

hsanjuan:
Sorry, it was not good to post without explanation. The rationale is:

* The secret keeps being recommended and it isolates the cluster from anyone not knowing
* We are not going to document follower-peers for the moment, probably providing a ipfs-cluster-follower binary in the future.
* In the general use-case, it seems that trusting all peers is a way more straightforward to get the cluster started than having the user figure out the peer IDs in advance and then putting that in the configuration.
* Ideally a default configuration can be used to form a cluster (by just setting the right secret).

@lanzafame also your opinion would be valuable here..

lanzafame:
@kishansagathiya The idea being that we want users to use CRDTs instead of Raft as a consensus component but we are not at the point where we can provide the appropriate level of polish for the follower feature set that CRDTs enable.

PR #869, opened 2019-08-06
Fix #732: Introduce native pin/update

### LINKED ISSUE #732: [Proxy] Intercept pin/update

hsanjuan:
This introduces a pin/update operation which allows to Pin a new item to
cluster indicating that said pin is an update to an already-existing pin.

When this is the case, all the configuration for the existing pin is copied to
the new one (including allocations). The IPFS connector will then trigger
pin/update directly in IPFS, allowing an efficient pinning based on
DAG-differences. Since the allocations where the same for both pins,
the pin/update can proceed.

PinUpdate does not unpin the previous pin (it is not possible to do this
atomically in cluster like it happens in IPFS). The user can manually do it
after the pin/update is done.

Internally, after a lot of deliberations on what the optimal way for this is,
I opted for adding a `PinUpdate` option to the `PinOptions` type (carries the
CID to update from). In order to carry this option from the REST API to the
IPFS Connector, it is serialized in the Protobuf (and stored in the
datastore). There is no other way to do this in a simple fashion since the Pin
object is piece of information that is sent around.

Additionally, making it a PinOption plays well with the Pin/PinPath APIs which
need little changes. Effectively, you are pinning a new thing. You are just
indicating that it should be configured from an existing one.

Fixes #732


hsanjuan:
TODO: ipfshttp test.

hsanjuan:
@Stebalien you said you wanted to review sometimes... this is an interesting one by the number of places it touches...

ipfsconn/ipfshttp/ipfshttp_test.go
```diff
@@ -99,24 +100,57 @@ func testPin(t *testing.T, method string) {
...
+func testPinUpdate(t *testing.T) {
+	ctx := context.Background()
+	ipfs, mock := testIPFSConnector(t)
+	defer mock.Close()
+	defer ipfs.Shutdown(ctx)
+
+	pin := api.PinCid(test.Cid1)
+	pin.PinUpdate = test.Cid1
+	// enforce pin/update even though it would be skipped
+	ipfs.config.PinMethod = "update"
```
kishansagathiya: This should work without setting PinMethod

`ipfs.config.PinMethod` should remain the same for one daemon run. Plus, `refs` and `pin` are something that you can use for all pin operation, but `update` is useful only for a kind of pin operation. Meaning `update` is the value you need only for one pin operation.

ipfsconn/ipfshttp/ipfshttp_test.go
```diff
@@ -99,24 +100,57 @@ func testPin(t *testing.T, method string) {
...
+	err = ipfs.Pin(ctx, pin)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
 func TestIPFSPin(t *testing.T) {
 	t.Run("method=pin", func(t *testing.T) { testPin(t, "pin") })
 	t.Run("method=refs", func(t *testing.T) { testPin(t, "refs") })
+	t.Run("method=update", func(t *testing.T) { testPinUpdate(t) })
```
kishansagathiya: here as well

sharness/t0030-ctl-pin.sh
```diff
@@ -88,6 +88,14 @@ test_expect_success IPFS,CLUSTER "pin data to cluster with user allocations" '
     echo $allocations | grep -q ${pid}
 '
 
+test_expect_success IPFS,CLUSTER "pin update a pin" '
+   cid1=`docker exec ipfs sh -c "echo test | ipfs add -q"`
+   ipfs-cluster-ctl pin add "$cid1"
```
kishansagathiya: We can't replace these two lines with `echo test | ipfs-cluster-ctl add -q`.  Would it be helpful to have that someday?

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -610,6 +610,63 @@ although unpinning operations in the cluster may take longer or fail.
...
+existing one, including name. This means that the new pin will bypass the
+allocation process and will be allocated to the same peers as the existing
+one.
+
+The cluster peers will try to Pin the new item on IPFS using the "pin update"
+command. This is especially efficient when the content of two pins (their DAGs)
+are similar.
+
+Unlike the "pin update" command in the ipfs daemon, this will not unpin the
+existing item from the cluster. Please run "pin rm" for that.
```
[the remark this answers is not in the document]
hsanjuan: No. I have considered all this.

1) First, we cannot do atomic Pin/Rm after PinAdd (update) operation. Doing a non-atomic pin rm opens the door to leaving the system in inconsistent state. If the unpin operation fails, you have an added pin in there. This could be done as a CLI feature (where --unpin is a shortcut for running pin update + unpin), BUT
2) In Cluster, pinning is an async operation. Pinning un cluster does not mean that the content is pinned in IPFS. Unpinning after pinning will likely unpin content from IPFS while the pinning process assocaited to pin/update has not completed. This can get to ugly places it it happens during GC.

> I would be cleaner to have a separate api endpoint for pin/update

A pin/update is just a pin/add: it creates a pin resource (or an allocation). The extra endpoint means adding extra code for an non-restful endpoint. the `/pins/<cid>/update/<newcid>` is also weird because, for the reasons above, the original pin is not modified.

>  Plus it carries api.Pin unnecessarily. We only need to and from

We cannot call an RPC endpoint with two IN arguments. We need to wrap them in a single object. We use the Pin for that because that's the right object for creating a new pin and for general consistency, given the fact that it calls the Pin rpc endpoint. Again, you're asking for more boilerplate code for 0 gain. The approach here is the iteration upon all those ideas into the most consistent/simple approach.

>  since we are getting options from PinGet

For now, we don't know if in the future we may extend options. Useful feedback here would have been to realize that we should probably respect the `Name` option for the update-pin and not use the one from the old pin. I will change that.

ipfsconn/ipfshttp/ipfshttp_test.go
hsanjuan: Please read the test. The connector will only attempt to use `pin/update` when the from-argument is pinned. This prevents situations where a pin/update failed (pin error) and it is recovered at a later moment, after the from-pin has been unpinned. That would result in an unrecoverable pin.

But the simplest way to check that the connector hits the "pin/update" endpoint (the purpose here), was to hit it in a way that fails in a expected manner. Therefore I need to generate an error by using pin/update when the `from` Cid is not pinned yet and the way to do it is this. After that the test runs the normal workflow by resetting.

Yes, it's a hack. The alternative is to launch a custom ipfs mock that I can inspect (making sure the calls hit the right endpoint). I can do that too if you think it's better... (it's more code and an extra mock but makes the test less hacky).

ipfsconn/ipfshttp/ipfshttp_test.go
hsanjuan: no, we are testing the case when we trigger a pin/update and not a pin.

sharness/t0030-ctl-pin.sh
hsanjuan: Yes, it would be useful. New feature request?

cmd/ipfs-cluster-ctl/main.go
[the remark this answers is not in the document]
hsanjuan: Also, note that the requesters of this feature use it with --unpin=false, meaning unpin is not always the natural step even though i thought the same.

kishansagathiya:
This is good to go after `make check` is fixed

hsanjuan:
ouch there's a bug, not all tests are being run

hsanjuan:
I will rewrite the pin/update test in upcoming PR

PR #871, opened 2019-08-06
Improve error handling on `add`

### LINKED ISSUE #852: Improve error handling on "add"

The add process is aborted whenever one of the peers receiving blocks errors. This makes it hard to add things to some clusters if a peer goes away for example. We should:

* Keep adding to peers that never errored
* Remember the peers that errored so that we don't try to put more blocks on them
* as long as the content is added to one place, the peers that errored will be able to fetch it later when pinning.

**Additional context**

This is part of the `adder` submodule and how we handle RPC requests to `IPFSConnector.PutBlock` from there (both local and sharding).

Tests adding to peers which are down and peers which are not should be included.

(Of course all peers fail, adding should stop).

kishansagathiya:
With this commit
- We don't stop adding if one peer failed. We keep adding to peers that
never errored
- We remember peers(only till current `/add` request) that errored and
don't try to put more block them

TODO: Tests

Fixes #852

### DIFF
--- a/add_test.go
+++ b/add_test.go
@@ -5,11 +5,16 @@ package ipfscluster
 import (
 	"context"
 	"mime/multipart"
+	"sync"
 	"testing"
 	"time"
 
+	cid "github.com/ipfs/go-cid"
+	files "github.com/ipfs/go-ipfs-files"
 	"github.com/ipfs/ipfs-cluster/api"
 	"github.com/ipfs/ipfs-cluster/test"
+	peer "github.com/libp2p/go-libp2p-core/peer"
+	rpc "github.com/libp2p/go-libp2p-gorpc"
 )
 
 func TestAdd(t *testing.T) {
@@ -101,4 +106,124 @@ func TestAddPeerDown(t *testing.T) {
 
 		runF(t, clusters, f)
 	})
+}
+
+func TestAddOnePeerFails(t *testing.T) {
+	ctx := context.Background()
+	clusters, mock := createClusters(t)
+	defer shutdownClusters(t, clusters, mock)
+	sth := test.NewShardingTestHelper()
+	defer sth.Clean(t)
+
+	waitForLeaderAndMetrics(t, clusters)
+
+	t.Run("local", func(t *testing.T) {
+		params := api.DefaultAddParams()
+		params.Shard = false
+		params.Name = "testlocal"
+		lg, closer := sth.GetRandFileReader(t, 50000) // 50 MB
+		defer closer.Close()
+
+		mr := files.NewMultiFileReader(lg, true)
+
+		r := multipart.NewReader(mr, mr.Boundary())
+
+		var err error
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			_, err = clusters[0].AddFile(r, params)
+			if err != nil {
+				t.Fatal(err)
+			}
+		}()
+
+		err = clusters[1].Shutdown(ctx)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		wg.Wait()
+	})
+}
+
+// Kishan: Uncomment after https://github.com/ipfs/ipfs-cluster/issues/761
+// is resolved. This test would pass, but not for the reason we want it to.
+// Add fails waiting for the leader.
+func TestAddAllPeersFail(t *testing.T) {
+	ctx := context.Background()
+	clusters, mock := createClusters(t)
+	defer shutdownClusters(t, clusters, mock)
+	sth := test.NewShardingTestHelper()
+	defer sth.Clean(t)
+
+	waitForLeaderAndMetrics(t, clusters)
+
+	t.Run("local", func(t *testing.T) {
+		params := api.DefaultAddParams()
+		params.Shard = false
+		params.Name = "testlocal"
+		lg, closer := sth.GetRandFileReader(t, 50000) // 50 MB
+		defer closer.Close()
+
+		mr := files.NewMultiFileReader(lg, true)
+
+		r := multipart.NewReader(mr, mr.Boundary())
+		params.PinOptions.ReplicationFactorMax = 2
+		params.PinOptions.ReplicationFactorMin = 2
+		clusters[0].allocator = &mockPinAllocator{
+			peers: []peer.ID{clusters[1].id, clusters[2].id},
+		}
+
+		var err error
+		// var cid cid.Cid
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			_, err = clusters[0].AddFile(r, params)
+			if err == nil {
+				t.Fatalf("expected error")
+			}
+		}()
+
+		for i := 1; i < 3; i++ {
+			err = clusters[i].Shutdown(ctx)
+			if err != nil {
+				t.Fatal(err)
+			}
+		}
+
+		wg.Wait()
+
+		// pinDelay()
+
+		// pini, err := clusters[0].Status(ctx, cid)
+		// if err != nil {
+		// 	t.Error(err)
+		// }
+		// fmt.Println(pini.String())
+
+		// pin, err := clusters[0].PinGet(ctx, cid)
+		// if err != nil {
+		// 	t.Fatal(err)
+		// }
+
+		// fmt.Println(pin.Allocations)
+	})
+}
+
+type mockPinAllocator struct {
+	peers []peer.ID
+}
+
+// SetClient does nothing in this allocator
+func (alloc mockPinAllocator) SetClient(c *rpc.Client) {}
+
+// Shutdown does nothing in this allocator
+func (alloc mockPinAllocator) Shutdown(_ context.Context) error { return nil }
+
+func (alloc mockPinAllocator) Allocate(ctx context.Context, c cid.Cid, current, candidates, priority map[peer.ID]*api.Metric) ([]peer.ID, error) {
+	return alloc.peers, nil
 }
--- a/add_test.go
+++ b/add_test.go
@@ -9,12 +9,11 @@ import (
 	"testing"
 	"time"
 
-	cid "github.com/ipfs/go-cid"
 	files "github.com/ipfs/go-ipfs-files"
+	"github.com/ipfs/ipfs-cluster/adder"
 	"github.com/ipfs/ipfs-cluster/api"
 	"github.com/ipfs/ipfs-cluster/test"
 	peer "github.com/libp2p/go-libp2p-core/peer"
-	rpc "github.com/libp2p/go-libp2p-gorpc"
 )
 
 func TestAdd(t *testing.T) {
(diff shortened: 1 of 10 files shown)


adder/util.go
```diff
@@ -14,8 +14,20 @@ import (
 	rpc "github.com/libp2p/go-libp2p-gorpc"
 )
 
+type BlockPutHelper struct {
+	failedPeers map[peer.ID]struct{}
+	rpcClient   *rpc.Client
+}
+
+func NewBlockPutHelper(rpcClient *rpc.Client) *BlockPutHelper {
```
hsanjuan: is there anything preventing you from initializing with the destinations as well, and removing items from that slice as they fail (which saves you an amount of looping in every block)?

obo20:
@kishansagathiya could you explain this more?
> * We remember peers(only till current `/add` request) that errored and
>   don't try to put more block them

Are you saying that cluster only remembers that a particular peer errored out for the duration of that particular add process? Or are you saying that all future add attempts to that peer will be blocked?

Also, how does cluster handle parallelization in this scenario? Let's say that Node A is adding two different pieces of content to cluster in parallel to both itself and Node B. If Node B errors for one **but not both** of the add operations, how does cluster handle that? 

Does A stop the add operation that didn't error out and proceed to reroute both add operations to Node C? Or does Node A just reroute the operation that fails?

kishansagathiya:
> Are you saying that cluster only remembers that a particular peer errored out for the duration of that particular add process? Or are you saying that all future add attempts to that peer will be blocked?

Not for future attempts, only for that `add` process(`/add` rest api)

> Does A stop the add operation that didn't error out and proceed to reroute both add operations to Node C? 

Not changing allocations here. There are two operations here, `BlockPut` and `Pin`. Say block in question is `k` and allocation are node A and B. If we couldn't BlockPut `k` on node B, but we could put that block on at least one machine( in this case A), no worries we move on to next blocks. If we couldn't put a block on any node `/add` would error. With remaining block we know that something is wrong with B(since our previous request failed), so we don't try to put any blocks there. 
In the end we pin all blocks on all nodes. So for node B, it would attempt to put those blocks again and pin them.

obo20:
@kishansagathiya I think I worded that last question poorly. I was asking about two separate add operations. For example user A is adding some content to the cluster and then midway through that user B also starts an add operation to the cluster. 

Let's say for whatever reason that user A's add operation errors, but user B's is still going fine. (I'm not sure if this is even possible. I'm not sure if there are any edge cases where an individual block could become corrupt or something like that). If user A's operation errors out, does that also cause user B's operation to error out as well?

Or are you saying that all parallel content add operations get lumped together in the same `BlockPut` operation?

kishansagathiya:
>  If user A's operation errors out, does that also cause user B's operation to error out as well?

No, these two parallel `/add` requests are independent. Information about peers that failed during an `/add` request are only limited to this request. The other parallel request is unaffected.

obo20:
@kishansagathiya Thanks for the clarification and thanks for putting this PR together. It should be super helpful

adder/util.go
```diff
@@ -14,8 +14,23 @@ import (
...
+func NewBlockPutHelper(rpcClient *rpc.Client, dests []peer.ID) *BlockPutHelper {
+	return &BlockPutHelper{
+		dests:     dests,
+		rpcClient: rpcClient,
+	}
+}
+
 // PutBlock sends a NodeWithMeta to the given destinations.
-func PutBlock(ctx context.Context, rpc *rpc.Client, n *api.NodeWithMeta, dests []peer.ID) error {
+func (bph *BlockPutHelper) PutBlock(ctx context.Context, n *api.NodeWithMeta) error {
```
hsanjuan: Now, since the `Helper` is actually looking a lot like a https://godoc.org/github.com/ipfs/go-ipld-format#NodeAdder, let's move into that direction and rename  `BlockPutHelper` to `BlockAdder`. 

`PutBlock` -> `Add`.

`PutDAG` -> `AddMany`

adder/util.go
```diff
@@ -26,19 +41,59 @@ func PutBlock(ctx context.Context, rpc *rpc.Client, n *api.NodeWithMeta, dests [
...
+		// request to all peers failed
+		return rpcutil.CheckErrs(errs)
+	}
+
+	bph.dests = actDests
+	return nil
+}
+
+// PutDAG puts multiple nodes to given destinations.
+func (bph *BlockPutHelper) PutDAG(ctx context.Context, nodes []ipld.Node) error {
```
hsanjuan: Becomes `AddMany`

adder/local/dag_service.go
```diff
@@ -50,6 +53,7 @@ func (dgs *DAGService) Add(ctx context.Context, node ipld.Node) error {
 			return err
 		}
 		dgs.dests = dests
+		dgs.bph.SetDests(dests)
```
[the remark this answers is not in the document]
kishansagathiya: dests are used as allocations while pinning.

Also, we can't use BlockPutHelper.dests as allocations for pinnning, since it would not have failed peers.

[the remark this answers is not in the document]
kishansagathiya: don't want to set a one time thing (BlockPutHelper) in a method that would be called multiple times. 
I actually wanted to move setting `dests` to the constructor.

> Also, chances are the main object does not need to remember dests anymore, since the BlockPutHelper does.

dests are used as allocations while pinning.

Also, we can't use BlockPutHelper.dests as allocations for pinnning, since it would not have failed peers.

adder/local/dag_service.go
```diff
@@ -62,7 +66,7 @@ func (dgs *DAGService) Add(ctx context.Context, node ipld.Node) error {
 		CumSize: size,
 	}
 
-	return adder.PutBlock(ctx, dgs.rpcClient, nodeSerial, dgs.dests)
+	return dgs.bph.PutBlock(ctx, nodeSerial)
```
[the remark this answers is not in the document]
kishansagathiya: NodeWithMeta let's pass `format`

BlockPut uses `v0`, but PutDAG uses `cbor` (v1) (why?)

 I will check if passing format makes things cleaners, but changing NodeWithMeta to IPLD Node will lead to quite a few changes including breaking rpc api for BlockPut

adder/local/dag_service.go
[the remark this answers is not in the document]
hsanjuan: > dests are used as allocations while pinning.

ok

> don't want to set a one time thing (BlockPutHelper) in a method that would be called multiple times

but you are doing exactly that wth blockputhelper.dests + having an extra method.

adder/local/dag_service.go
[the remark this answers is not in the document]
hsanjuan: Format tells ipfs what codec to give to the CID of the block. Depends on whether we are sending a block from the dag builder (v0 protobufs) or a clusterDAGnode which we use cbor for. I believe, format can also be a field in the helper, since it's fixed for all the things that helper needs to put?

adder/local/dag_service.go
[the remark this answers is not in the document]
kishansagathiya: > I actually wanted to move setting dests to the constructor.

This should make things better

adder/local/dag_service.go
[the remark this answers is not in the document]
kishansagathiya: > I believe, format can also be a field in the helper, since it's fixed for all the things that helper needs to put?

No. 

bph is part of shard object which uses protobuf on PutBlock and cbor on PutDAG. 

We can make format part of bph and use a new bph everytime PutDAG is called.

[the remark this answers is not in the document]
hsanjuan: Or check whether the node passed is cbor `_, ok := n.(*cbor.Node)`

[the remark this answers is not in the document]
hsanjuan: whatever you prefer

add_test.go
```diff
@@ -97,3 +101,88 @@ func TestAddPeerDown(t *testing.T) {
...
+				}
+
+				if p.Status != api.TrackerStatusPinned {
+					t.Error("item should not be pinned")
+				}
+			}
+		})
+
+		t.Run("all_peers_fail", func(t *testing.T) {
+			params.PinOptions.UserAllocations = []peer.ID{clusters[1].id, clusters[2].id, clusters[3].id, clusters[4].id}
```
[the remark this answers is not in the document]
kishansagathiya: I have commented this test with a note of uncommenting it when allocations are supported

add_test.go
```diff
@@ -121,68 +119,69 @@ func TestAddWithFailingPeers(t *testing.T) {
...
+			t.Fatal(err)
+		}
 
-			wg.Wait()
-		})
+		wg.Wait()
 	})
 }
+
+// Kishan: Uncomment after https://github.com/ipfs/ipfs-cluster/issues/761
```
hsanjuan: No, find a way to test this. You should be able to shutdown allocated cluster peers while adding. You don't need UserAllocations.

adder/util.go
```diff
@@ -14,88 +14,86 @@ import (
...
+	if err != nil {
+		return err
+	}
+	nodeSerial := &api.NodeWithMeta{
+		Cid:     node.Cid(),
+		Data:    node.RawData(),
+		CumSize: size,
+	}
+
+	format, ok := cid.CodecToStr[nodeSerial.Cid.Type()]
```
hsanjuan: good way!

adder/util.go
```diff
@@ -14,88 +14,86 @@ import (
...
 }
 
-// PutBlock sends a NodeWithMeta to the given destinations.
-func (bph *BlockPutHelper) PutBlock(ctx context.Context, n *api.NodeWithMeta) error {
-	format, ok := cid.CodecToStr[n.Cid.Type()]
+// Add puts an ipld node to allocated destinations.
+func (ba *BlockAdder) Add(ctx context.Context, node ipld.Node) error {
+	size, err := node.Size()
+	if err != nil {
+		return err
```
hsanjuan: just warn, we don't even need the size for anything..

adder/util.go
```diff
@@ -14,88 +14,86 @@ import (
...
-			Format: "cbor",
-		}
-		//logger.Debugf("Here is the serialized ipld: %x", b.Data)
-
-		logger.Debugf("DAG block put %s", n.Cid())
-		err := bph.PutBlock(ctx, b)
+// AddMany puts multiple ipld nodes to allocated destinations.
+func (ba *BlockAdder) AddMany(ctx context.Context, nodes []ipld.Node) error {
+	for _, node := range nodes {
+		logger.Debugf("DAG block put %s", node.Cid())
```
hsanjuan: this will be logged in `Add` as well. no need to log here.

add_test.go
hsanjuan: Also, if you ever want to skip a test, there is `t.Skip()`

kishansagathiya: Haven't figured this out yet. I was able to set allocations with a mockAllocator, but somehow rpc calls are not erroring on peers which are shutdown(true for all tests in `add_test.go`)

hsanjuan: You have resolved the other conversations but I don't see the commits. I will jump in and fix this test.

kishansagathiya: It's in the same commit https://github.com/ipfs/ipfs-cluster/pull/871/commits/38d1e2f62085179828f67985ed07a71d1cc4fe53#diff-79784c46c699792f79bdfe22c9dbf73e

hsanjuan: sorry, didn't see

hsanjuan:
approved in side-band chat

PR #874, opened 2019-08-09
Service: Select consensus on "init" (not on "daemon")

### LINKED ISSUE #865: Mandatory `--consensus` for `service init`

`ipfs-cluster-service init --consensus raft|crdt`

Add a mandatory `consensus` option to `init` command

Fill up config (service.json) based on provided consensus, i.e. if raft is provided only raft config should be filled and crdt config should be empty.

When the daemon is run, it should infer the consensus from config, i.e. use one for which config is present. If config for both (raft and crdt) are present, pick default consensus(what should it be for now?)
 
**Additional context**
From discussion from https://github.com/ipfs/ipfs-cluster/pull/856#issuecomment-516344599

hsanjuan:
Fixes #865.

This makes the necessary changes so that consensu is selected on "init" with a
flag set, by default, to "crdt". This generates only a "crdt" or a "raft"
section, not both.

If the configuration file has a "raft" section, "raft" will be used to start
the daemon. If it has a "crdt" section, "crdt" will be used. If it has none or
both sections, an error will happen.

This also affects "state *" commands, which will now autoselect how to work
from the existing configuration.

### DIFF
--- a/cmd/ipfs-cluster-service/daemon.go
+++ b/cmd/ipfs-cluster-service/daemon.go
@@ -62,21 +62,25 @@ func daemon(c *cli.Context) error {
 
 	cfgs := cfgHelper.Configs()
 
-	if !c.Bool("no-trust") {
-		crdtCfg := cfgs.Crdt
-		crdtCfg.TrustedPeers = append(crdtCfg.TrustedPeers, ipfscluster.PeersFromMultiaddrs(bootstraps)...)
-	}
-
 	if c.Bool("stats") {
 		cfgs.Metrics.EnableStats = true
 	}
 	cfgHelper.SetupTracing(c.Bool("tracing"))
 
-	// Cleanup state if bootstrapping
+	// Setup bootstrapping
 	raftStaging := false
-	if len(bootstraps) > 0 && cfgHelper.GetConsensus() == cfgs.Raft.ConfigKey() {
-		raft.CleanupRaft(cfgs.Raft)
-		raftStaging = true
+	switch cfgHelper.GetConsensus() {
+	case cfgs.Raft.ConfigKey():
+		if len(bootstraps) > 0 {
+			// Cleanup state if bootstrapping
+			raft.CleanupRaft(cfgs.Raft)
+			raftStaging = true
+		}
+	case cfgs.Crdt.ConfigKey():
+		if !c.Bool("no-trust") {
+			crdtCfg := cfgs.Crdt
+			crdtCfg.TrustedPeers = append(crdtCfg.TrustedPeers, ipfscluster.PeersFromMultiaddrs(bootstraps)...)
+		}
 	}
 
 	if c.Bool("leave") {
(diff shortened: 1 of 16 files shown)


hsanjuan:
Note docker-compose is broken but will get fixed once we merge to master (it uses `ipfs-cluster:latest `which does not work with this compose).

sharness/t0021-service-init.sh
```diff
@@ -18,6 +18,12 @@ test_expect_success "cluster-service init without --peers succeeds and creates e
     [ ! -s "test-config/peerstore" ]
 '
 
+test_expect_success "cluster-service init with raft generates only raft config" '
+    ipfs-cluster-service --config "test-config" init -f --consensus raft &&
+    [ "$(jq -M -r .consensus.raft test-config/service.json)" != "null" ] && 
+    [ "$(jq -M -r .consensus.crdt test-config/service.json)" == "null" ]
+'
+
```
kishansagathiya: same test for crdt ?

cmd/ipfs-cluster-service/daemon.go
```diff
@@ -78,7 +74,7 @@ func daemon(c *cli.Context) error {
```
kishansagathiya: line 65, check if consensus is crdt?

hsanjuan:
I'll deal with docker in a different PR.

cmd/ipfs-cluster-service/main.go
```diff
@@ -249,6 +255,11 @@ multiaddresses.
 			),
 			ArgsUsage: "[http-source-url]",
 			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:  "consensus",
+					Usage: "select consensus component: 'crdt' or 'raft'",
+					Value: defaultConsensus,
```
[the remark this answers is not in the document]
hsanjuan: well, it's not mandatory.. I think this way of running cluster is the simplest and the one that most matches the user's expectations therefore we autoselect it. I'm happy to listen to strong reasons against this though. 

(@lanzafame chime in too?)

sharness/t0021-service-init.sh
hsanjuan: added

cmd/ipfs-cluster-service/daemon.go
hsanjuan: done

cmd/ipfs-cluster-service/main.go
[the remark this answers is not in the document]
lanzafame: Yeah, I think default to crdt as that has the simplest bootstrapping workflow. And a user needs to know why they would want to use raft over crdts, before they use it so I am fine with this. 

Caveat that it should be made clear in the startup logs which consensus component is being used. Actual, the startup logs need to be revamped, I will create an issue for that.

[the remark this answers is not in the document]
hsanjuan: Thanks

PR #875, opened 2019-08-09
ipfshttp: Remove "refs" method

hsanjuan:
go-ipfs no longer performs a global lock when two simultaneous requests to pin happen. This allows us to remove a bunch of code.

Made on top of #869 , will re-push once merged.


kishansagathiya:
Will review this once #869 is merged

hsanjuan:
> need to remove `pin_method` from test data config_test.go and sharness (searching `pin_method` should show all instances)

done

PR #878, opened 2019-08-11
Sort addresses in `/id`

### LINKED ISSUE #327: Arrays in HTTP API responses are not sorted
Somewhat annoying but minor issue.

When making API requests to ipfs-cluster, it seems that the response sometimes changes the order of the arrays. So the first request returns `{"peers": ["A", "B"]}` and the second one could return `{"peers": ["B", "A"]}` randomly. This is a bit annoying since responses are harder to cache, since the ordering in arrays are important.

Sorting the arrays before letting the HTTP endpoint respond, would solve this issue.

The endpoints I've found so far not properly sorting before responding:

Endpoint | Attributes(s) not sorted
------------ | -------------
`/id` | `addresses`, `cluster_peers_addresses` and `ipfs.addresses`
`/allocations` | Response is non-sorted array
`/peers` | Each peer has the same response from `/id` so needs sorting on the same attributes 

I'm sure I missed others, but I've not used all API endpoints.

kishansagathiya:
and thus in GET `/peers` as well

Fixes #327


cluster.go
```diff
@@ -727,7 +727,14 @@ func (c *Cluster) ID(ctx context.Context) *api.ID {
 	for _, addr := range c.host.Addrs() {
 		addrsSet[addr.String()] = struct{}{}
 	}
+
+	var addrSorted []string
```
hsanjuan: I am not sure why we are "filtering dups" above. This is very old code. Was `host.Addrs()` producing duplicates?

If we removed that, would we need sorting? Probably the sorting would not flap in that case.

And can we use the canonical https://godoc.org/github.com/libp2p/go-libp2p-core/peer#AddrInfoToP2pAddrs and deprecate `MustLibp2pMultiaddrJoin` thing?

hsanjuan: There is code to dedup addresses in libp2p: https://github.com/libp2p/go-libp2p/blob/master/p2p/host/basic/basic_host.go#L592

hsanjuan: (and it does not list from the map so the order is certainly kept)

pstoremgr/pstoremgr.go
```diff
@@ -139,9 +140,11 @@ func (pm *Manager) filteredPeerAddrs(p peer.ID) []ma.Multiaddr {
 	}
 
 	if len(peerDNSAddrs) > 0 {
+		sort.Sort(utils.ByString(peerDNSAddrs))
```
hsanjuan: not sure this is necessary as these are not flapping

kishansagathiya:
This however still doesn't guarantee that results will be same everytime you run `/id` 

It is likely that second time you get more addresses than the first time.

cluster.go
```diff
@@ -721,22 +721,11 @@ func (c *Cluster) ID(ctx context.Context) *api.ID {
...
-	for k := range addrsSet {
-		addrSorted = append(addrSorted, k)
-	}
-	sort.Strings(addrSorted)
-
-	for _, k := range addrSorted {
-		addr, _ := api.NewMultiaddr(k)
-		addrs = append(addrs, api.MustLibp2pMultiaddrJoin(addr, c.id))
+	var addrs []api.Multiaddr
+	mAddrs, _ := peer.AddrInfoToP2pAddrs(&peer.AddrInfo{ID: c.id, Addrs: c.host.Addrs()})
```
hsanjuan: error checking

kishansagathiya:
> And then, because go map iteration is random, places where we have a peermap and we print it on screen (like ctl status) might also show different orders between the peers. And this should also be ordered

status can be sorted or order can be kept by modifying the cluster.StatusAll() logic, but considering large number of pins  keeping the order might affect performance

sharness/t0033-ctl-peers.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
hsanjuan: sharness is not the right place to test this as this is not ctl functionality. I am very much against adding sharness tests for things that can be tested without sharness (in this case, in a function that calls `Peers()` in `ipfscluster_test.go`).

Sharness is difficult to maintain, the least we use of it the better.

Also, you are making 2 requests to a single-peer cluster. Obviously the results in the array cannot have different orders because there is only one element in that array. So this is not testing much.

hsanjuan:
> status can be sorted or order can be kept by modifying the cluster.StatusAll() logic, but considering large number of pins keeping the order might affect performance

Let's leave this (status) until pintrackers are re-worked.

ipfscluster_test.go
```diff
@@ -565,6 +567,29 @@ func TestClustersPeers(t *testing.T) {
...
 
+func TestClustersPeersRetainOrder(t *testing.T) {
+	ctx := context.Background()
+	clusters, mock := createClusters(t)
+	defer shutdownClusters(t, clusters, mock)
+
+	delay()
+
+	j := rand.Intn(nClusters) // choose a random cluster peer
+	peers1, err := json.Marshal(clusters[j].Peers(ctx))
```
hsanjuan: If you had not made the changes and had written this test before, chances are it would have passed just fine.

The order depended on metrics. You should probably do WaitForMetrics() in between. And use different peers. And repeat the comparison for like 5 times.

PR #879, opened 2019-08-12
Make sure all API errors are in JSON

### LINKED ISSUE #657: REST API improvements
Inspired from discussions that happened in https://github.com/ipfs/ipfs-cluster/pull/634

- [ ] Support trailing slashes in URLs. That is, it should not redirect, but behave just the same way as URLs without a trailing slash(same response, same http code, no multiple requests or redirect)
- [ ] Errors returned by our APIs are in JSON format. Requests that doesn't reach our handlers and are responded by Gorilla are not in JSON format. Make sure that those responses are also in JSON format. So, when we fail to match a route instead of text message `error not found 404`, it should return a JSON response.
- [ ] Effects on tests. We have `makePostWithContentType` and other similar test helper functions, which should work for given content type judging them by their names, but there logic assumes that we are always using Content-Type as `application/json`. `processResp` and `checkHeaders` assumes that response should have. I don't see a point in having `WithContentType` functions when we know all responses should return json responses. 
- [ ] Add tests for various APIs, make sure that trailing slash support is tested and returns JSON response. Test that APIs are not returning text `error not found 404`, but a JSON response.
 
Should be done after https://github.com/ipfs/ipfs-cluster/pull/634 is merged. https://github.com/ipfs/ipfs-cluster/pull/634 won't be perfect, but I think changes in this issue should make it so.

Thoughts?

kishansagathiya:
This commit introduces a catchAllHandler(which returns error not found),
which will be used if path doesn't match with any other pattern.

Fixes #657


api/rest/restapi.go
```diff
@@ -263,6 +263,10 @@ func (api *API) addRoutes(router *mux.Router) {
 				),
 			)
 	}
+	router.NotFoundHandler = ochttp.WithRouteTag(
+		http.HandlerFunc(api.notFoundHandler),
+		"/"+"notfound",
```
hsanjuan: you must mean `"/notfound"` or `"/404"` (better), but not `"/"+"notfound"`

kishansagathiya:
```suggestion
		"/notfound",
```

PR #882, opened 2019-08-13
Improve Docker support for CRDTs

hsanjuan:
cluster: add mDNS service discovery

I always thought the libp2p node would do this, but it is not the case.
With this, CRDT peers are able to autodiscover on local networks.

More commits coming.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,18 +1,28 @@
 version: '3.4'
 
-# This is an example docker-compose file for IPFS Cluster
-# It runs two Cluster peers (cluster0, cluster1) attached to two
-# IPFS daemons (ipfs0, ipfs1).
+# This is an example docker-compose file to quickly test an IPFS Cluster
+# with multiple peers on a contained environment.
+
+# It runs 3 cluster peers (cluster0, cluster1...) attached to go-ipfs daemons
+# (ipfs0, ipfs1...) using the CRDT consensus component. Cluster peers
+# autodiscover themselves using mDNS on the docker internal network.
 #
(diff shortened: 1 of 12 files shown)


hsanjuan:
Ok, this has what it should have. Fully ready now.

docker-compose.yml
```diff
@@ -25,10 +35,10 @@ services:
   ipfs0:
     container_name: ipfs0
     image: ipfs/go-ipfs:release
-    ports:
-          - "4001:4001" # ipfs swarm
-#         - "5001:5001" # expose if needed/wanted
-#         - "8080:8080" # exposes if needed/wanted
+#    ports:
+#         - "4001:4001" # ipfs swarm - expose if needed/wanted
```
kishansagathiya: what changed?

docker-compose.yml
```diff
@@ -38,61 +48,69 @@ services:
...
+##################################################################################
+## Cluster PEER 2 ################################################################
+##################################################################################
+
+# See Cluster PEER 0 for comments (all removed here and below)
+  ipfs2:
+    container_name: ipfs2
+    image: ipfs/go-ipfs:release
+    ports:
+          - "4201:4001"
```
kishansagathiya: why keep swarm port open for just this?

docker-compose.yml
hsanjuan: I think it does not make a difference for the testing purposes, since it's not a node that will be dialed from the outside

docker-compose.yml
hsanjuan: mistake

hsanjuan: fixed

PR #890, opened 2019-08-16
Added UserAllocations support for `add`

### LINKED ISSUE #761: UserAllocations support for `add`
https://github.com/ipfs/ipfs-cluster/pull/647 introduced support for user allocations to `pin add`. Since `add` does pinning as well, it would make sense to add user allocations support for `add` as well

kishansagathiya:
Fixes #761

### DIFF
--- a/api/add.go
+++ b/api/add.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/url"
 	"strconv"
+	"strings"
 
 	cid "github.com/ipfs/go-cid"
 )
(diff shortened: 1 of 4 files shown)


api/add.go
```diff
@@ -101,9 +102,14 @@ func AddParamsFromQuery(query url.Values) (*AddParams, error) {
 	params.Layout = layout
 
 	chunker := query.Get("chunker")
-	params.Chunker = chunker
+	if chunker != "" {
```
hsanjuan: why?

kishansagathiya: default chunker is size-262144, which at this point is already set. This is avoiding setting it to empty.

kishansagathiya:
I have avoided `calling allocate() only if pin.Allocations is empty` , because pin.Allocations are supposed to be empty at that point
if we want to put a safety check we can throw an error or warning if pin.Allocations are filled at that point.

PR #891, opened 2019-08-19
Metadata can be set and shown from `ctl`

### LINKED ISSUE #670: Allow pins to have arbitrary metadata
#### Pre-check

#### Basic information

  * [ ] Bug
  * [X] Feature request
  * [ ] Enhancement

#### Description

Currently, ipfs-cluster allows users to attach a `title` metadata which is stored together with the pins but separate from the pin itself.

For Cube (and other use cases as well), it'd be very helpful if ipfs-cluster could store additional metadata as well. Things that come into my mind is things like tags, links to other resources and description. Hopefully it can be implemented in a way that arbitrary structures can be supported.

The way the metadata would be created could be either when creating the pin itself, after the creation of a pin in ipfs-cluster (attaching metadata). We would also need to have a way of updating/deleting metadata. With this, comes the need of having a log of the modifications to metadata. Not sure if this is needed in a first iteration on ipfs-clusters side.

On the other hand, this issue comes from https://github.com/ipfs-shipyard/cube/issues/12 which brings up the thought that maybe metadata is not something that we need directly in ipfs-cluster, maybe should stay a user-land thing instead. I'll leave that up to the cluster team to think about.

kishansagathiya:
This PR adds
- `metadata` option to `pin add` and `add` commands
- ability to set metadata from `/add` rest API

Fixes #670


api/add.go
```diff
@@ -146,6 +148,18 @@ func AddParamsFromQuery(query url.Values) (*AddParams, error) {
 		return nil, err
 	}
 
+	params.Metadata = make(map[string]string)
```
hsanjuan: No need to allocate the map since it's already allocated in DefaultParams?

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -396,6 +400,12 @@ cluster "pin add".
 				p := api.DefaultAddParams()
 				p.ReplicationFactorMin = c.Int("replication-min")
 				p.ReplicationFactorMax = c.Int("replication-max")
+				metadata := make(map[string]string)
+				for _, str := range c.StringSlice("metadata") {
+					parts := strings.Split(str, "=")
+					metadata[parts[0]] = parts[1]
```
hsanjuan: Can I make this panic by passing `--metadata hello`? Probably yes.

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -553,11 +567,18 @@ would stil be respected.
 							}
 						}
 
+						metadata := make(map[string]string)
+						for _, str := range c.StringSlice("metadata") {
+							parts := strings.Split(str, "=")
```
hsanjuan: worth having a function to parse metadata flags.

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -517,6 +527,10 @@ would stil be respected.
 							Value: "",
 							Usage: "Sets a name for this pin",
 						},
+						cli.StringSliceFlag{
+							Name:  "metadata",
+							Usage: "Arbitrary metadata to be stored with pins",
```
hsanjuan: `Pin metadata: key=value. Can be added multiple times`

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -1089,6 +1078,19 @@ func waitFor(
 	return client.WaitFor(ctx, globalClient, fp)
 }
 
+func parseMetadata(metadata []string) map[string]string {
+	metadataMap := make(map[string]string)
+	for _, str := range metadata {
+		parts := strings.Split(str, "=")
```
hsanjuan: Use `SplitN(2)`. Allow for additional `=` to be part of the value.

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -188,7 +188,18 @@ func textFormatPrintPin(obj *api.Pin) {
...
 
-	fmt.Printf(" | %s\n", recStr)
+	fmt.Printf(" | %s", recStr)
+
+	if len(obj.Metadata) == 0 {
+		fmt.Printf("\n")
+		return
+	}
+
+	fmt.Printf(" | Metadata:")
```
hsanjuan: Just print `Metadata: yes|no`. Otherwise it just likely that the output will get too polluted. Users can always inspect with --enc=json.

hsanjuan:
> And another for adding.

This one is still missing

PR #893, opened 2019-08-21
Pin recover on all peers

### LINKED ISSUE #763: `/pins/recover` should work without the local flag as well
Currently, for `/pins/recover` has local flag mandatory and it has to be `true` (i.e, throws bad request if not passed or passed but other than true)

This is equal to not having the flag. I think recover should work without the flag as well.

kishansagathiya:
- recover works without `--local` flag as well (recovers all pins on all
peers)
- remove extra space from rpc policy

Fixes #763


ipfscluster_test.go
```diff
@@ -1096,6 +1096,41 @@ func TestClustersRecover(t *testing.T) {
...
+	pinDelay()
+
+	_, err := clusters[rand.Intn(nClusters)].RecoverAll(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	delay()
+
+	info, err := clusters[rand.Intn(nClusters)].Status(ctx, h2)
```
hsanjuan: The other Recover tests are not a very good example on how to test this as they are old, but at least it trigger the recover pathway.

Here you are pinning a normal Cid and then calling recover which will not do much. But assuming that is more or less "covered" by the other test, you should check that you get the expected global pin info object when recovering an `ErrorCid`.

In short:

```
Pin(cid1)
Pin(errorCid)
RecoverAll(errorCid)
```

Should return a globalPinInfo on which you check:

* Has two items
* len(peermap) == nClusters for each item

The rest is not really necessary, as Status is tested in many other places.

rpc_policy.go
```diff
@@ -6,7 +6,7 @@ package ipfscluster
 // endpoint type. See rpcutil/policygen.go as a quick way to generate this
 // without missing any endpoint.
 var DefaultRPCPolicy = map[string]RPCEndpointType{
-	// Cluster methods
+	//  Cluster methods
```
hsanjuan: Can you fix the generator so it does not add this extra space?

hsanjuan: Seems you did, can you fix the space though?

test/rpc_api_mock.go
```diff
@@ -297,6 +297,11 @@ func (mock *mockCluster) SyncLocal(ctx context.Context, in cid.Cid, out *api.Pin
 	return mock.StatusLocal(ctx, in, out)
 }
 
+func (mock *mockCluster) RecoverAll(ctx context.Context, in struct{}, out *[]*api.GlobalPinInfo) error {
+	*out = make([]*api.GlobalPinInfo, 0, 0)
```
hsanjuan: rely on `mock.StatusAll(...)` (like Sync does).

rpc_policy.go
```diff
@@ -19,6 +19,7 @@ var DefaultRPCPolicy = map[string]RPCEndpointType{
 	"Cluster.PinPath":            RPCClosed,
 	"Cluster.Pins":               RPCClosed, // Used in stateless tracker, ipfsproxy, restapi
 	"Cluster.Recover":            RPCClosed,
+	"Cluster.RecoverAll":         RPCClosed,
```
hsanjuan: `RecoverAllLocal` and `RecoverLocal` should be RPCTrusted though (with comment `// Called in broadcast from RecoverAll() / Recover()`)

hsanjuan:
There is a fail:

```
--- FAIL: TestAPIRecoverAllEndpoint (1.53s)

    --- FAIL: TestAPIRecoverAllEndpoint/in-parallel (0.00s)

        --- FAIL: TestAPIRecoverAllEndpoint/in-parallel/libp2p (0.73s)

            restapi_test.go:1034: bad response length

        --- FAIL: TestAPIRecoverAllEndpoint/in-parallel/http (1.00s)

            restapi_test.go:1034: bad response length
```

PR #894, opened 2019-08-21
Logging for Cluster HTTP APIs

### LINKED ISSUE #574: HTTP API: Logging
#### Basic information

  * [x] Master
  * [ ] Latest stable version
  * [ ] Bug
  * [ ] Feature request
  * [x] Enhancement

#### Description

Our HTTP API currently does not do any logging that includes request, return codes, duration etc... (webserver-like). We should enable this just like any other webserver. We should also have a way to log, in the case of authorized requests (with basic auth), which requests were performed by which user.

We likely want these logs to be specific to the API component and have them written out separately from the usual output, and stored somewhere (`/var/log/ipfs-cluster`?).

To discuss:

* Where to put these logs (if anywhere)
* If this should be enabled by default
* If this compels us to start using https://echo.labstack.com/ (and rewriting the API with it?)

cc. @lanzafame

kishansagathiya:
This commit introduces logging for Cluster HTTP APIs.
It adds a config element `send_logs_to_file`, which tells whether logs
should be saved in a file or shown in standard output.

Requests are logged as per Apache Common Log Format (CLF)
http://httpd.apache.org/docs/2.2/logs.html#common

Fixes #574


kishansagathiya:
This logs only cluster APIs, worth adding proxy endpoints as well?

api/rest/config.go
```diff
@@ -185,7 +206,7 @@ func (cfg *Config) Default() error {
 	cfg.CORSExposedHeaders = DefaultCORSExposedHeaders
 	cfg.CORSAllowCredentials = DefaultCORSAllowCredentials
 	cfg.CORSMaxAge = DefaultCORSMaxAge
-
+	cfg.HTTPLogFile = ""
```
hsanjuan: Pay attention to context. You have attached the HTTPLogFile option to the bottom next to some CORS options. It would fit better after Auth. Also keep consistency with rest:

```
// Logs
cfg.HTTPLogFile = ""
```

api/rest/config.go
```diff
@@ -147,6 +153,21 @@ type jsonConfig struct {
 	CORSExposedHeaders   []string `json:"cors_exposed_headers"`
 	CORSAllowCredentials bool     `json:"cors_allow_credentials"`
 	CORSMaxAge           string   `json:"cors_max_age"`
+	HTTPLogFile          string   `json:"http_log_file"`
+}
+
+// GetHTTPLogPath gets full path of the file where http logs should be
+// saved.
+func (cfg *Config) GetHTTPLogPath() string {
```
hsanjuan: Add to the bottom of the file.

hsanjuan: Does it need to be exported? Doesn't seem so...

api/rest/config.go
```diff
@@ -147,6 +153,21 @@ type jsonConfig struct {
 	CORSExposedHeaders   []string `json:"cors_exposed_headers"`
 	CORSAllowCredentials bool     `json:"cors_allow_credentials"`
 	CORSMaxAge           string   `json:"cors_max_age"`
+	HTTPLogFile          string   `json:"http_log_file"`
```
hsanjuan: You have attached the HTTPLogFile option to the bottom next to some CORS options. It would fit better after or before Auth

api/rest/config.go
```diff
@@ -122,6 +122,12 @@ type Config struct {
 
 	// Tracing flag used to skip tracing specific paths when not enabled.
 	Tracing bool
+
+	// HTTPLogFile is path of the file that would save HTTP API logs. If this
+	// path is empty, HTTP logs would be sent to standard output. This path
+	// should either be absolute or relative to cluster base directory. Its
+	// default value is empty.
+	HTTPLogFile string
```
hsanjuan: You have attached the HTTPLogFile option to the bottom. It would be better after BasicAuthCredentials. Tracing is always last.

sharness/t0055-service-api-log.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
hsanjuan: This is using sharness to test something that is not CLI dependent. Use a simple test in `api/rest/` to check that the log file is created.

api/rest/restapi.go
```diff
@@ -1153,3 +1167,34 @@ func (api *API) setHeaders(w http.ResponseWriter) {
 
 	w.Header().Add("Content-Type", "application/json")
 }
+
+func customLog(writer io.Writer, params handlers.LogFormatterParams) {
```
hsanjuan: We cannot become maintainers of how to log in Apache format, least of all hardcode it in a way we might logging different things depending on whether we log to file or to stdout because one uses gorilla and other not.

The right approach is to make a custom `io.Writer` to pass to `LoggingHandler` which instead of writing to stdout, writes to the logger.

Having to copy-paste unexported code from a library should raise red flags as a generally bad approach to something.

kishansagathiya:
api logs are not coming on stdout for some reason. Will fix it and push

hsanjuan:
> api logs are not coming on stdout for some reason. Will fix it and push

you need to update `logging.go` with the new facility name.

api/rest/restapi_test.go
```diff
@@ -1039,6 +1040,38 @@ func TestAPIRecoverAllEndpoint(t *testing.T) {
 	testBothEndpoints(t, tf)
 }
 
+func TestAPILogging(t *testing.T) {
+	ctx := context.Background()
+	cfg := &Config{}
+	cfg.Default()
+
+	cfg.HTTPLogFile = "/tmp/http.log"
```
hsanjuan: set to a relative path. There's nothing wrong, but cluster tests do not write anything outside the local repository folder.

api/rest/restapi_test.go
```diff
@@ -1039,6 +1040,38 @@ func TestAPIRecoverAllEndpoint(t *testing.T) {
...
+
+	rest := testAPIwithConfig(t, cfg, "log_enabled")
+	defer rest.Shutdown(ctx)
+	defer os.Remove(cfg.HTTPLogFile)
+
+	data, err := ioutil.ReadFile(cfg.HTTPLogFile)
```
hsanjuan: check out https://golang.org/pkg/os/#Stat . But I'm not sure we have to test that an empty file exists right away. Better to test would be to make sure than an existing file is not replaced on a subsequent start.

api/rest/restapi_test.go
```diff
@@ -1039,6 +1040,38 @@ func TestAPIRecoverAllEndpoint(t *testing.T) {
...
+	}
+
+	if len(data) > 0 {
+		t.Errorf("expected empty log file, %+v", data)
+	}
+
+	id := api.ID{}
+	makeGet(t, rest, httpURL(rest)+"/id", &id)
+
+	data, err = ioutil.ReadFile(cfg.HTTPLogFile)
```
hsanjuan: https://golang.org/pkg/os/#Stat

sharness/config/test/identity.json
```diff
@@ -0,0 +1,4 @@
+{
```
hsanjuan: delete

sharness/config/test/service.json
```diff
@@ -0,0 +1,136 @@
+{
```
hsanjuan: delete

api/rest/restapi_test.go
```diff
@@ -1045,31 +1046,57 @@ func TestAPILogging(t *testing.T) {
 	cfg := &Config{}
 	cfg.Default()
 
-	cfg.HTTPLogFile = "/tmp/http.log"
+	baseDir, err := filepath.Abs("")
+	if err != nil {
+		t.Fatal(err)
+	}
+	cfg.BaseDir = baseDir
+	cfg.HTTPLogFile = "http.log"
```
hsanjuan: does a simpler `cfg.HTTPLogFile,_ = filepath.Abs("http.log")` not work?

api/rest/restapi_test.go
```diff
@@ -1045,31 +1046,57 @@ func TestAPILogging(t *testing.T) {
...
 	if err != nil {
 		t.Fatal(err)
 	}
-	if len(data) == 0 {
+	size2 := info.Size()
+	if size2 == 0 {
 		t.Error("did not expect an empty log file")
 	}
+
+	if !(size2 > size1) {
```
hsanjuan: you mean, `if size2 <= size1`

kishansagathiya: Yes, but I think this is more readable,
if size2 is not greater than size1 --> size did not grow

hsanjuan:
> worth adding proxy endpoints as well?

Please, on separate PR.

PR #899, opened 2019-08-24
crdt: warn about trust all mode

hsanjuan:


consensus/crdt/consensus.go
```diff
@@ -257,6 +257,11 @@ func (css *Consensus) setup() {
 		return
 	}
 	css.state = clusterState
+
+	if css.config.TrustAll {
+		logger.Info("'trust all' mode enabled. Any peer in the cluster can modify the pinset.")
```
kishansagathiya: Didn't you want to use `logger.Warn` ?

hsanjuan: No, because I don't want to warn for something that is a default setting. I want to INFOrm the user. Commit message not very accurate in this regard.

PR #901, opened 2019-08-26
Error queue is full

### LINKED ISSUE #377: Pin queue is full when starting a node
Initial state sync on large pinsets may likely call `Track` enough times to trigger a `pin queue is full`. (https://github.com/ipfs/ipfs-cluster/issues/160#issuecomment-380192112). It should be investigated when this can happen, I'm not 100% sure. Also, double check how critical is this (I think subsequent ipfs-syncs will pin any items that were left out).

kishansagathiya:
- abort if a Track() calls fails due to queue being full
- increase max pin queue size to 1 million
- hind max_pin_queue_size from configuration
- use an elaborated error message

Fixes #377

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -13,7 +13,6 @@ import (
 	"github.com/ipfs/ipfs-cluster/adder/local"
 	"github.com/ipfs/ipfs-cluster/adder/sharding"
 	"github.com/ipfs/ipfs-cluster/api"
-	"github.com/ipfs/ipfs-cluster/pintracker/maptracker"
 	"github.com/ipfs/ipfs-cluster/pstoremgr"
 	"github.com/ipfs/ipfs-cluster/rpcutil"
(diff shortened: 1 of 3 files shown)


cluster.go
```diff
@@ -942,7 +943,10 @@ func (c *Cluster) StateSync(ctx context.Context) error {
 		_, tracked := trackedPinsMap[pin.Cid.String()]
 		if !tracked {
 			logger.Debugf("StateSync: tracking %s, part of the shared state", pin.Cid)
-			c.tracker.Track(ctx, pin)
+			err = c.tracker.Track(ctx, pin)
+			if err == maptracker.ErrFullQueue {
```
hsanjuan: You should not import an specific pintracker here. The Cluster only knows about PinTracekrs, not their particular implementations. Moreover, since you don't know how a pintracker is implemented, you should abort on any error,

PR #905, opened 2019-08-27
Fix: trust-all remained enabled always

hsanjuan:
:/ good that I added a warning

### DIFF
--- a/consensus/crdt/config.go
+++ b/consensus/crdt/config.go
@@ -105,6 +105,10 @@ func (cfg *Config) LoadJSON(raw []byte) error {
 
 func (cfg *Config) applyJSONConfig(jcfg *jsonConfig) error {
 	config.SetIfNotDefault(jcfg.ClusterName, &cfg.ClusterName)
+
+	// Whenever we parse JSON, TrustAll is false unless an '*' peer exists
+	cfg.TrustAll = false
+
 	for _, p := range jcfg.TrustedPeers {
 		if p == "*" {
 			cfg.TrustAll = true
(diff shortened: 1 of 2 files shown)


consensus/crdt/config.go
```diff
@@ -117,6 +117,9 @@ func (cfg *Config) applyJSONConfig(jcfg *jsonConfig) error {
 		}
 		cfg.TrustedPeers = append(cfg.TrustedPeers, pid)
 	}
+	if len(cfg.TrustedPeers) > 0 {
```
kishansagathiya: `TrustAll` is a short hand for the condition `len(cfg.TrustedPeers) == 0`, then why have a default for it at all?

hsanjuan: It's not that. 

['*'] -> TrustAll. 

[] -> No trust all. 

['something'] -> No trust all.

kishansagathiya: It's value is still definitive from content of `jcfg.TrustedPeers`

But `cfg.TrustAll` helps in telling whether `jcfg.TrustedPeers` should be `[]string{"*"}` or empty. So, we can let it be.

> [] -> No trust all.

because of DefaultTrustAll being true, this won't happen. I tried and this comes out to be true (when `jcfg.TrustedPeers` is `[]`).

Also the description for `TrustAll` says it is `regardless of the TrustedPeers contents.`  (but it isn't)

You have thought of it as mode. But as per the code, the mode is enabled or disabled by `DefaultTrustAll` and not `cfg.TrustAll`, which is fine. We can just change the description for that element.

hsanjuan: > because of DefaultTrustAll being true, this won't happen. I tried and this comes out to be true (when jcfg.TrustedPeers is []).

Ok, I fixed that now I think

> Also the description for TrustAll says it is regardless of the TrustedPeers contents. (but it isn't)

That's correct because that applies to the `Config` object. It's the `jsonConfig` that is different by merging both TrustedPeers and TrustAll into a single []string.

PR #907, opened 2019-08-28
`--local` flag for `ctl add`

### LINKED ISSUE #848: "ipfs-cluster-ctl add --local" adds to the local peer

~~A --local flag to "ipfs-cluster-ctl add" should allow to add and pin only on the local peer (with replication-factor=1).~~

**Additional context**

~~Adding may be cumbersome if trying to reach a high replication factor with unavailable peers. This should allow users add to their local node only (they can then re-pin if they want to increase the replication factor).~~

~~This is almost the equivalent of adding with ReplicationFactor=1 + UserAllocations=LocalPeer. However, it might be probably best to carry it as a new, separate add option (to cover the case where there are no local metrics. In that case, the given UserAllocations would be ignored in favour of some other peer).~~

kishansagathiya:
Fixes #848


adder/single/dag_service.go
```diff
@@ -62,19 +68,14 @@ func (dgs *DAGService) Add(ctx context.Context, node ipld.Node) error {
 func (dgs *DAGService) Finalize(ctx context.Context, root cid.Cid) (cid.Cid, error) {
 	// Cluster pin the result
 	rootPin := api.PinWithOpts(root, dgs.pinOpts)
-	rootPin.Allocations = dgs.dests
+	if dgs.local {
+		return root, adder.Pin(ctx, dgs.rpcClient, rootPin)
+	}
```
hsanjuan: The allocations should be respected. We are block-putting locally, but the file should be pinned normally as it would have been pinned if we were block-putting to other places.

adder/single/dag_service.go
```diff
@@ -30,22 +30,28 @@ type DAGService struct {
...
+		local:     local,
 	}
 }
 
 // Add puts the given node in the destination peers.
 func (dgs *DAGService) Add(ctx context.Context, node ipld.Node) error {
+	if dgs.local {
+		return adder.AddNodeLocal(ctx, dgs.rpcClient, node)
+	}
+
```
hsanjuan: you need to do the blockallocate even if you don't use it to send the blocks to those destinations because you are running on local node.

hsanjuan: Below, if you are in local-mode, your BlockAdder destinations are []("") (or better, modify the block adder to have the local parameter). There is no reason to add new code to add blocks to a place when you already have code to do that.

adder/util.go
```diff
@@ -142,6 +142,21 @@ func Pin(ctx context.Context, rpc *rpc.Client, pin *api.Pin) error {
 	)
 }
 
+// AddNodeLocal puts the given node to the local peer.
+func AddNodeLocal(ctx context.Context, rpc *rpc.Client, node ipld.Node) error {
```
hsanjuan: as noted above, this is duplicate code not really needed when you have a working blockAdder that does this.

cluster.go
```diff
@@ -1484,7 +1484,7 @@ func (c *Cluster) AddFile(reader *multipart.Reader, params *api.AddParams) (cid.
 	if params.Shard {
 		dags = sharding.New(c.rpcClient, params.PinOptions, nil)
 	} else {
-		dags = local.New(c.rpcClient, params.PinOptions)
+		dags = single.New(c.rpcClient, params.PinOptions, false)
```
hsanjuan: You have the `params.Local` at hand to pass it in rather than hardcoding to false.

adder/single/dag_service.go
kishansagathiya: pin operation would take care of that since we care passing pin options with rootPin,
Checked it to be respecting options.

adder/single/dag_service.go
kishansagathiya: can do this by passing an empty dest

kishansagathiya:
Little trick I used to check if local add works
```
➜  ✗ ipfs-cluster-ctl add ~/Downloads/what.png 
added QmSq28uXPP3JMXTAZZSbRBaiUtqqJh8AYSg7d1NBgdPz4H what.png
➜  ✗ ipfs --api /ip4/127.0.0.1/tcp/5001 repo gc
removed QmPk85r5P6iaDwTC4VGwdvdNNuciQxBHA2tDM87JRwk3ie
➜  ✗ ipfs --api /ip4/127.0.0.1/tcp/5101 repo gc
removed QmPk85r5P6iaDwTC4VGwdvdNNuciQxBHA2tDM87JRwk3ie
➜  ✗ ipfs-cluster-ctl add --local ~/Downloads/what.png
added QmSq28uXPP3JMXTAZZSbRBaiUtqqJh8AYSg7d1NBgdPz4H what.png
➜  ✗ ipfs --api /ip4/127.0.0.1/tcp/5001 repo gc       
removed QmPk85r5P6iaDwTC4VGwdvdNNuciQxBHA2tDM87JRwk3ie
➜  ✗ ipfs --api /ip4/127.0.0.1/tcp/5101 repo gc       
➜  ✗ 
```

local ipfs would have extra blocks to gc

adder/single/dag_service.go
kishansagathiya: https://github.com/ipfs/ipfs-cluster/pull/907/commits/ca81048ffa0b3cd62dc5e44c158f640f26264eec

adder/single/dag_service.go
```diff
@@ -18,7 +18,7 @@ import (
 
 var errNotFound = errors.New("dagservice: block not found")
 
-var logger = logging.Logger("localdags")
+var logger = logging.Logger("singledags")
```
hsanjuan: Need to update `logging.go` too.

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -338,6 +338,10 @@ content.
 					Usage: "Hash function to use. Implies cid-version=1",
 					Value: defaultAddParams.HashFun,
 				},
+				cli.BoolFlag{
+					Name:  "local",
+					Usage: "Puts blocks only in local IPFS peer and then pin as per PinOptions",
```
hsanjuan: It's not clear what `PinOptions` is here. Better `Add to local peer but pin normally`.

And then update the related paragraph in the long description:

```
Once the adding process is finished, the content is fully added to all
allocations and pinned in them. This makes cluster add slower than a local
ipfs add, but the result is a fully replicated CID on completion.
If you prefer faster adding, use the --local flag to add directly to the local IPFS node
and pin in the destinations after that. Note that the local IPFS node may not be
among the destinations, which will leave the unpinned content in it.
```

adder/single/dag_service.go
kishansagathiya: Made this change in the latest commit, but interestingly this log is never used.

PR #909, opened 2019-09-02
Explanation of 0 replication factor

### LINKED ISSUE #755: ctl add --replication-value help output about default 0 value confusing
#### Pre-check

#### Basic information

  * [x] Master
  * [ ] Latest stable version
  * [ ] Bug
  * [ ] Feature request
  * [x] Enhancement

#### Description

```
$ ipfs-cluster-ctl pin add -h
...
OPTIONS:
   --replication value, -r value          Sets a custom replication factor (overrides -rmax and -rmin) **(default: 0)**
   --replication-min value, --rmin value  Sets the minimum replication factor for this pin **(default: 0)**
   --replication-max value, --rmax value  Sets the maximum replication factor for this pin **(default: 0)**
...
```
It is correct to say that the default value is 0 but what this means in practice is that the `cluster.replication-min` and `cluster.replication-max` values defined in `.ipfs-cluster/service.json` will be used not that the default replication factor is actually 0.

kishansagathiya:
Fixes #755

> Might get away with it by making it a string flag and parsing the values...

this would work, but still we would have to explain the value 0 replication factor for someone who passes `-r "0"`

So I have avoided changing the flag

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -69,8 +69,6 @@ require (
 	github.com/libp2p/go-ws-transport v0.1.0
 	github.com/mattn/go-isatty v0.0.9 // indirect
 	github.com/miekg/dns v1.1.16 // indirect
-	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
-	github.com/modern-go/reflect2 v1.0.1 // indirect
 	github.com/multiformats/go-multiaddr v0.0.4
 	github.com/multiformats/go-multiaddr-dns v0.0.3
 	github.com/multiformats/go-multiaddr-net v0.0.1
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -69,6 +69,8 @@ require (
 	github.com/libp2p/go-ws-transport v0.1.0
 	github.com/mattn/go-isatty v0.0.9 // indirect
 	github.com/miekg/dns v1.1.16 // indirect
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
+	github.com/modern-go/reflect2 v1.0.1 // indirect
```
hsanjuan: ?

PR #911, opened 2019-09-05
monitor/metrics endpoint and display improvements

### LINKED ISSUE #374: ipfs-cluster-ctl health: report whether cluster is stable
When setting up a cluster, it would useful to know when the cluster has become stable after adding a peer(s). A command with following example output would be useful:
```bash
$ # Example Stable State
$ ipfs-cluster-ctl health
Cluster: Stable
Leader: QmNpXvAXVbEmfXnBfBuexa9KufGKLj2Bsf187s4pBwjDTQ | <hostname> | <health>
Followers: 
  - QmdVqxKtQMJo78QFTTvC3wDmDn3hr16apuTxT7UdZ2VHoU | <hostname> | <health>
  - QmeqKDbL3vSNp1NCt6SmZm59BftWFEsYjpJekPqNeRTBVe | <hostname> | <health>
```

```bash
$ # Example Unstable State
$ ipfs-cluster-ctl health
Cluster: Unstable (No Leader)
Candidates: 
  - QmNpXvAXVbEmfXnBfBuexa9KufGKLj2Bsf187s4pBwjDTQ | <hostname> | <health>
  - QmdVqxKtQMJo78QFTTvC3wDmDn3hr16apuTxT7UdZ2VHoU | <hostname> | <health>
Peers: 
  - QmeqKDbL3vSNp1NCt6SmZm59BftWFEsYjpJekPqNeRTBVe | <hostname> | <health>
```

```bash
$ # Example Health of Single Node
$ ipfs-cluster-ctl health QmNpXvAXVbEmfXnBfBuexa9KufGKLj2Bsf187s4pBwjDTQ
Health: QmNpXvAXVbEmfXnBfBuexa9KufGKLj2Bsf187s4pBwjDTQ | <hostname> | <health> | <leader|candidate|follower>
```

@hsanjuan @ZenGround0 thoughts on whether you would find this useful?

kishansagathiya:
- Human-sizes for freespace metrics. Display whether if metric is
expires in something like "expires in 3m".
- When not passing metric name `ipfs-cluster-ctl health metrics` hits
the the metrics endpoint which returns a list of available metrics and
displays to user

Fixes #374


api/rest/client/methods.go
```diff
@@ -334,6 +334,16 @@ func (c *defaultClient) Metrics(ctx context.Context, name string) ([]*api.Metric
 	return metrics, err
 }
 
+// MetricNames lists names of all metrics.
+func (c *defaultClient) MetricNames(ctx context.Context) ([]string, error) {
+	ctx, span := trace.StartSpan(ctx, "client/Metrics")
```
lanzafame:
```suggestion
	ctx, span := trace.StartSpan(ctx, "client/MetricNames")
```

ipfscluster.go
```diff
@@ -176,6 +176,8 @@ type PeerMonitor interface {
 	// LatestMetrics returns a map with the latest metrics of matching name
 	// for the current cluster peers.
 	LatestMetrics(ctx context.Context, name string) []*api.Metric
+	// MetricNames returns a list of metric names.
+	MetricNames() []string
```
lanzafame: Always pass contexts to interface methods, especially, when there a corresponding API endpoint, so that request cancellation propogates the entire way through the stack and also so it doesn't need to be added later for metrics/tracing stuff.

kishansagathiya: thanks

api/rest/client/lbclient.go
```diff
@@ -369,6 +369,19 @@ func (lc *loadBalancingClient) Metrics(ctx context.Context, name string) ([]*api
 	return metrics, err
 }
 
+// MetricNames returns the list of metrics.
+func (lc *loadBalancingClient) MetricNames(ctx context.Context) ([]string, error) {
```
lanzafame: The loadbalancing client should have tracing as well. Though that may be a seperate PR since all the methods are missing it.

go.mod
```diff
@@ -69,6 +69,8 @@ require (
 	github.com/libp2p/go-ws-transport v0.1.0
 	github.com/mattn/go-isatty v0.0.9 // indirect
 	github.com/miekg/dns v1.1.16 // indirect
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
```
lanzafame: I didn't see any new imports, so I am guessing you may need to run `go mod tidy`

monitor/pubsubmon/pubsubmon.go
```diff
@@ -245,3 +245,11 @@ func (mon *Monitor) LatestMetrics(ctx context.Context, name string) []*api.Metri
 func (mon *Monitor) Alerts() <-chan *api.Alert {
 	return mon.checker.Alerts()
 }
+
+// MetricNames list all metric names.
+func (mon *Monitor) MetricNames(ctx context.Context) []string {
+	ctx, span := trace.StartSpan(ctx, "monitor/pubsub/MetricNames")
+	defer span.End()
+
+	return mon.metrics.MetricNames()
```
lanzafame:
```suggestion
	return mon.metrics.MetricNames(ctx)
```

rpc_api.go
```diff
@@ -616,3 +616,9 @@ func (rpcapi *PeerMonitorRPCAPI) LatestMetrics(ctx context.Context, in string, o
 	*out = rpcapi.mon.LatestMetrics(ctx, in)
 	return nil
 }
+
+// MetricNames runs PeerMonitor.MetricNames().
+func (rpcapi *PeerMonitorRPCAPI) MetricNames(ctx context.Context, in struct{}, out *[]string) error {
+	*out = rpcapi.mon.MetricNames()
```
lanzafame:
```suggestion
	*out = rpcapi.mon.MetricNames(ctx)
```

kishansagathiya: This is in `store`, we dont use context here

monitor/pubsubmon/pubsubmon.go
kishansagathiya: This is in `store`, we dont use context here

api/rest/client/lbclient.go
kishansagathiya: Yes

rpc_api.go
lanzafame: 👍

monitor/pubsubmon/pubsubmon.go
lanzafame: 👍

monitor/pubsubmon/pubsubmon.go
```diff
@@ -245,3 +245,11 @@ func (mon *Monitor) LatestMetrics(ctx context.Context, name string) []*api.Metri
 func (mon *Monitor) Alerts() <-chan *api.Alert {
 	return mon.checker.Alerts()
 }
+
+// MetricNames list all metric names.
```
hsanjuan:
```suggestion
// MetricNames lists all metric names.
```

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -223,3 +224,9 @@ func trackerStatusAllString() string {
 	sort.Strings(strs)
 	return strings.Join(strs, "\n")
 }
+
+func textFormatPrintList(obj []string) {
```
hsanjuan: add a 

```
case string:
   fmt.Println(resp)
```

towards the top of `textFormatObject`

the `[]string` case should do the loop of the items like the other slice types (for consistency)

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -223,3 +224,9 @@ func trackerStatusAllString() string {
...
+	for _, s := range obj {
+		fmt.Printf("- %s\n", s)
```
hsanjuan: avoid the `- `. Print directly

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -914,7 +914,10 @@ but usually are:
 					Action: func(c *cli.Context) error {
 						metric := c.Args().First()
 						if metric == "" {
-							checkErr("", errors.New("provide a metric name"))
+							fmt.Printf("Try with one of the metrics names listed below:\n\n")
```
hsanjuan: avoid random text getting into the output. This does not make a text interface easy to work with when sedding, grepping, etc.

Instead, update the documentation of the command which is outdated.

rpc_api.go
hsanjuan: No, you do need to pass the content here

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -204,8 +210,7 @@ func textFormatPrintAddedOutputQuiet(obj *addedOutputQuiet) {
 }
 
 func textFormatPrintMetric(obj *api.Metric) {
-	date := time.Unix(0, obj.Expire).UTC().Format(time.RFC3339)
-	fmt.Printf("%s: %s | Expire: %s\n", peer.IDB58Encode(obj.Peer), obj.Value, date)
+	fmt.Printf("%s: %s | Expires in: %s\n", peer.IDB58Encode(obj.Peer), obj.Value, time.Until(time.Unix(0, obj.Expire)).String())
```
hsanjuan: I was going to open an extra issue but let's fix it here:

`ipfs-cluster-ctl heath metrics freespace` should display the value in human form. We should also write the metric type (for all metrics): `<peer id> | freespace: 25GB | ...`. Also for ping we should write `<peer id> | ping | ...` (without an empty `: `).

There is a go-humanize library to do this (we already use it in sharding).

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -904,6 +904,8 @@ graph of the connections.  Output is a dot file encoding the cluster's connectio
 This commands displays the latest valid metrics of the given type logged
 by this peer for all current cluster peers.
 
+If the argument is not passed, it would list current metric types.
```
hsanjuan:
```suggestion
If no argument is provided, the command retrieves all currently existing metric types.
```

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -210,7 +213,13 @@ func textFormatPrintAddedOutputQuiet(obj *addedOutputQuiet) {
 }
 
 func textFormatPrintMetric(obj *api.Metric) {
-	fmt.Printf("%s: %s | Expires in: %s\n", peer.IDB58Encode(obj.Peer), obj.Value, time.Until(time.Unix(0, obj.Expire)).String())
+	if obj.Name == "freespace" {
+		u, err := strconv.ParseUint(obj.Value, 10, 64)
+		checkErr("parsing to uint64", err)
+		fmt.Printf("%s | freespace: %s | Expires in: %s\n", peer.IDB58Encode(obj.Peer), humanize.Bytes(u), humanize.Time(time.Unix(0, obj.Expire)))
+	} else if obj.Name == "ping" {
```
hsanjuan: we can leave this case as a general fallback, not only for `ping`

monitor/metrics/util.go
```diff
@@ -14,13 +14,22 @@ func PeersetFilter(metrics []*api.Metric, peerset []peer.ID) []*api.Metric {
...
 			continue
 		}
+		metricsByPeer[metric.Peer] = metric
+	}
+
+	for _, peer := range peerset {
+		metric, ok := metricsByPeer[peer]
+		if !ok {
+			continue
+		}
```
hsanjuan: This is changing behavior. The function now only allows 1 metric per peer. It is used only in that case, but it should be reflected in the godocs that this function is not doing that it says it's doing anymore.

Also, if doing it this way, then re-use peerMap to store the metrics rather than adding another map with peers as keys.

kishansagathiya:
What the reason allow nil `mon.peers` ?

ipfscluster_test.go
```diff
@@ -2038,3 +2038,21 @@ func TestClustersFollowerMode(t *testing.T) {
...
+	metrics1 := clusters[0].monitor.LatestMetrics(ctx, clusters[0].informer.Name())
+	metrics2 := clusters[0].monitor.LatestMetrics(ctx, clusters[0].informer.Name())
+
+	for i, m := range metrics1 {
+		if *m != *metrics2[i] {
+			t.Error("expected both metrics to be same")
+			break
+		}
+	}
+}
```
hsanjuan: This test is not really testing. It can happily pass if metrics happen to be in the same order by chance, or it may fail if new metrics come in between the calls to LatestMetrics. I prefer to not have the extra code and remove this.

monitor/metrics/store.go
```diff
@@ -75,9 +76,16 @@ func (mtrs *Store) LatestValid(name string) []*api.Metric {
...
+	peers := []peer.ID{}
+	for p := range byPeer {
+		peers = append(peers, p)
+	}
+	sortedPeers := peer.IDSlice(peers)
+	sort.Stable(sortedPeers)
+
+	metrics := make([]*api.Metric, 0, len(peers))
+	for _, p := range peers {
+		m, err := byPeer[p].Latest()
```
hsanjuan: let's do this cleanly:

* create MetricSlice type
* Implement the sort.Interface
* Use that type here and call sort.Stable(metrics)
* I'm not sure if better to sort by peer or by expiry time (perhaps first by peer, second by expire time).

This will save a bunch of allocations, make everything cleaner and more readable.

api/types.go
```diff
@@ -875,6 +875,18 @@ func (m *Metric) Discard() bool {
 	return !m.Valid || m.Expired()
 }
 
+// MetricSlice is a sortable Metric array.
+type MetricSlice []*Metric
+
+func (es MetricSlice) Len() int      { return len(es) }
+func (es MetricSlice) Swap(i, j int) { es[i], es[j] = es[j], es[i] }
+func (es MetricSlice) Less(i, j int) bool {
+	if string((*es[i]).Peer) == string((*es[j]).Peer) {
```
hsanjuan: this looks wrong. 

```
	if es[i].Peer == es[j].Peer {
		return es[i].Expire < es[j].Expire
	}
	return es[i].Peer < es[j].Peer
```

no need to overcomplicate it. Go dereferences pointers automatically and a peer.ID is already a string.

kishansagathiya:
https://github.com/ipfs/ipfs-cluster/issues/374#issuecomment-545427757

PR #913, opened 2019-09-06
Fix: consolidate parsing of PinOptions

hsanjuan:
This removes duplicated code to parse PinOptions.

### DIFF
--- a/api/add.go
+++ b/api/add.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/url"
 	"strconv"
-	"strings"
 
 	cid "github.com/ipfs/go-cid"
 )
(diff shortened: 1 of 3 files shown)


api/add.go
```diff
@@ -144,44 +146,11 @@ func AddParamsFromQuery(query url.Values) (*AddParams, error) {
...
-		params.UserAllocations = StringsToPeers(strings.Split(allocs, ","))
-	}
-
 	err = parseIntParam(query, "cid-version", &params.CidVersion)
 	if err != nil {
 		return nil, err
 	}
 
-	if v := query.Get("shard-size"); v != "" {
-		shardSize, err := strconv.ParseUint(v, 10, 64)
```
kishansagathiya: ShardSize is part of PinOptions. Shouldn't this be part of AddParams instead?

hsanjuan: I found it weird too, but I did not check if there's a reason. I'll check and come back.

hsanjuan: The reason is that AddParams contain solely parameters for the Adder (non-cluster things, more like, everything you can customize when doing an `ipfs add`). `ShardSize` is a PinOption however, because it affects how big cluster makes the shards on one side, and in the other it is used to store the ShardSize (the real one) when a shard is pinned (so that it is possible to check how big a shard is).

I don't know if this is all made in the best way but changing this here is out of scope of this as ShardSize being in PinOptions is not casual and it would involve a deep dive into sharding.

PR #923, opened 2019-09-18
Pin Expiration

### LINKED ISSUE #481: Pin expiration
#### Basic information

  * [x] Master
  * [ ] Latest stable version
  * [ ] Bug
  * [x] Feature request
  * [ ] Enhancement

#### Description

IPFS Cluster could automatically expire Pins given an expiring date, associated to the Pin. The pinset could be watched regularly and unpins triggered for expired items.

This would probably be a feature of the main component, as pin parameters are handled from there. It would imply storing expiry dates associated to Pins in the state.

Note, unpinning from cluster would just mean that content in the IPFS nodes can be garbage collected, but not that the content will stop being available from the network (if it has been fetched from somewhere else).

kishansagathiya:
This commit introduces a pin option called `expire`(time in
nanoseconds), after which the given item will be unpinned.

Fixes #481


kishansagathiya:
I think it would be more user friendly to accept duration in command and api query. But in pin options we can store expire time.

api/types.go
```diff
@@ -464,6 +464,7 @@ type PinOptions struct {
 	Name                 string            `json:"name" codec:"n,omitempty"`
 	ShardSize            uint64            `json:"shard_size" codec:"s,omitempty"`
 	UserAllocations      []peer.ID         `json:"user_allocations" codec:"ua,omitempty"`
+	Expire               int64             `json:"expire" codec:"e,omitempty"`
```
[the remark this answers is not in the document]
kishansagathiya: Too many problem in using time.Time

changed expire to int64 from time.Time
`time.Now()` has a monotonic clock reading, but `time.Unix(0, nano)` doesn't
have it. This creates problems while equating two times. Have faced some
more problems. It is safer to just use int64 instead.

If I keep time.Time, there will be back and forth conversion from time.Time to int64(string) which converting from pin options to query and back. The time that you get back with `time.Unix(0, nano)` is not the same. 
Plus with default value is simpler with int64, for time.Time  `time.Unix(0, 0)` is undefined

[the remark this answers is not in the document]
hsanjuan: Please use `time.Time`. We need to represent a date. It's the same on why we don't carry a []byte slice instead of peer.ID etc.

We don't have to use monotonic clocks. I don't even understand why you would need nanosecond accuracy for this.

hsanjuan:
Actually there is a big issue here: several peers may be running StateSync and run unpin on expired items at the same time which is not very optimal.

kishansagathiya:
> Actually there is a big issue here: several peers may be running StateSync and run unpin on expired items at the same time which is not very optimal.

Fixing this requires a way to select the same peer definitely by any node that is running StateSync.
  
If I just pick the first peer, then certain peers will have more workload then other (those which comes earlier in the sorting order), I think a better way would be to pick the closest peer to the cid. 

Plus, 
If we want only one peer to call Unpin, we have to make sure that peer is trusted by others. That means we need to know which peers are trusted by others and whether you (a peer) are trusted by others or not. For that we need notion of trust in terms of cluster, .i.e, trusted by cluster ( right now what we have let's call it `trusted by peers`)
As a work around for this, we can pick a peer from our list of trusted peers (closest peer to cid).

If a follower node `a` is running StateSync and it happens to select another follower node `b` to unpin it, `b`'s Unpin would fail. But that's okay. Somebody else will run StateSync would not trust `b` ( since it's a follower node), and pick a more trusted `c` to unpin things.

cluster.go
```diff
@@ -988,8 +988,9 @@ func (c *Cluster) StateSync(ctx context.Context) error {
 	}
 
 	// a. Untrack items which should not be tracked
-	// b. Track items which should not be remote as local
-	// c. Track items which should not be local as remote
+	// b. Untrack items which have expired
```
kishansagathiya:
```suggestion
	// b. Unpin items which have expired
```

cluster.go
```diff
@@ -1021,6 +1040,29 @@ func (c *Cluster) StateSync(ctx context.Context) error {
 	return nil
 }
 
+// getClosestTrustedPeer gets the highest peer that is smaller than cid.
```
kishansagathiya:
```suggestion
// getClosestTrustedPeer gets the highest trusted peer that is smaller than cid.
```

cluster_test.go
```diff
@@ -773,6 +773,43 @@ func TestClusterPins(t *testing.T) {
...
+	}
+
+	pinDelay()
+
+	pins, err := cl.Pins(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(pins) != 1 {
+		t.Fatal("pin should be part of the state")
```
kishansagathiya:
```suggestion
		t.Error("pin should be part of the state")
```

cluster_test.go
```diff
@@ -773,6 +773,43 @@ func TestClusterPins(t *testing.T) {
...
+
+	// try after expiry time and state sync interval
+	time.Sleep(cl.config.StateSyncInterval + 1*time.Second)
+
+	pins, err = cl.Pins(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(pins) != 0 {
+		t.Fatal("pin should not be part of the state")
```
kishansagathiya:
```suggestion
		t.Error("pin should not be part of the state")
```

api/types.go
```diff
@@ -510,6 +512,10 @@ func (po *PinOptions) Equals(po2 *PinOptions) bool {
 		return false
 	}
 
+	if po.ExpireAt.Format(time.RFC3339) != po2.ExpireAt.Format(time.RFC3339) {
```
hsanjuan: https://golang.org/pkg/time/#Time.Equal

api/types.go
```diff
@@ -530,6 +536,10 @@ func (po *PinOptions) ToQuery() string {
 	q.Set("name", po.Name)
 	q.Set("shard-size", fmt.Sprintf("%d", po.ShardSize))
 	q.Set("user-allocations", strings.Join(PeersToStrings(po.UserAllocations), ","))
+	var t time.Time
+	if po.ExpireAt != t {
+		q.Set("expire-at", fmt.Sprintf("%s", po.ExpireAt.Format(time.RFC3339)))
```
hsanjuan: https://golang.org/pkg/time/#Time.MarshalText

api/types.go
```diff
@@ -573,6 +583,20 @@ func (po *PinOptions) FromQuery(q url.Values) error {
 		po.UserAllocations = StringsToPeers(strings.Split(allocs, ","))
 	}
 
+	if v := q.Get("expire-at"); v != "" {
+		tm, err := time.Parse(time.RFC3339, v)
```
hsanjuan: https://golang.org/pkg/time/#Time.UnmarshalText

cluster.go
```diff
@@ -1003,6 +1004,24 @@ func (c *Cluster) StateSync(ctx context.Context) error {
 			continue
 		}
 
+		var t time.Time
+		if !currentPin.PinOptions.ExpireAt.Equal(t) && currentPin.PinOptions.ExpireAt.Before(time.Now()) {
```
hsanjuan: don't call time.Now() for every pin.

api/types.go
```diff
@@ -691,6 +718,7 @@ func (pin *Pin) ProtoMarshal() ([]byte, error) {
 		// UserAllocations:      pin.UserAllocations,
 		Metadata:  pin.Metadata,
 		PinUpdate: pin.PinUpdate.Bytes(),
+		ExpireAt:  expireAtProto,
```
hsanjuan: Given that the protobuf timestamp type is a pointer, leave it as null when the expireAt is less than now. This will very probably save lots of space by not having to store an actual time.

cluster.go
hsanjuan: This is shaky. Better to define a ZeroTime constant in types.h (per Protobufs definition it is the result of : `time.Unix(0, 0).UTC()` (https://github.com/golang/protobuf/blob/master/ptypes/timestamp.go#L95) (or better, the result of doing pbuf.Timestamp(nil)) and compare on that.

cluster.go
```diff
@@ -1021,6 +1040,29 @@ func (c *Cluster) StateSync(ctx context.Context) error {
...
+		if cidStr > peer.IDB58Encode(p) || closest > p {
+			continue
+		}
+
+		closest = p
+	}
+
+	return closest
+}
+
```
hsanjuan: Rework all this into something that makes sense. You are comparing something like a peer ID B58 encoding (which maybe looks like `Qm...` with the `String()` of a cid which may be looks like `ba...` or whatever. This comparison does not make any sense and you are likely to always result in the same peer ID. I have no clue what you are doing because there is an additional peer.ID comparison involved and that's happening on a stringified byte-slice.

The right way to do it is exemplified by the dht code: you have really useful functions here: https://github.com/libp2p/go-libp2p-kbucket/blob/master/util.go#L55 (usef them)

Remember to ConvertKey() before-hand on the CID so the actual two things that you are calculating the distance on are barely comparable.

cluster.go
```diff
@@ -1003,6 +1004,24 @@ func (c *Cluster) StateSync(ctx context.Context) error {
...
+				ctx,
+				getClosestTrustedPeer(ctx, pCid.String(), currentPin.Allocations, c.consensus),
+				"Cluster",
+				"Unpin",
+				currentPin,
+				&pinObj,
+			)
+			if err != nil {
+				logger.Warning(err)
+			}
```
hsanjuan: Don't randomly hand off to other peers. If you have 1000 peers, one of them will receive 1000 requests to unpin a pin that has expired. What you want to figure out is "am I the peer most likely to be in charge of unpinning this item?" and then proceed if that is so.

cluster.go
```diff
@@ -1003,6 +1004,24 @@ func (c *Cluster) StateSync(ctx context.Context) error {
 			continue
 		}
 
+		var t time.Time
+		if !currentPin.PinOptions.ExpireAt.Equal(t) && currentPin.PinOptions.ExpireAt.Before(time.Now()) {
+			logger.Infof("StateSync: unpinning %s: pin expired", pCid)
```
hsanjuan: `Pin expired at...` and log the expiry time.

api/types.go
kishansagathiya: This is intentional, because those two times are actually not equal.
1) different monotonic clocks
2) number of nanosecond (I am using a format that doesn't have nano second precision)

I can use a format with nsec precision and fix this.

hsanjuan: ` those two times are actually not equal`

Then why do you  consider them equal?

Monotonic clocks are only involved when doing `Time.Now`. Other than expire-in (which we do Time.Now() for the user), the time set by the user should not mutate into something different should be directly comparable to the same time set by the user on a different instance.

cluster.go
hsanjuan: There's also an IsZero().

kishansagathiya: Yes, but `time.Unix(0,0)` is not zero
An uninitialized time variable is zero.

Even though, I didn't anything like you can't manually get zero time, I have inferred it so.
The function `t.UnixNano()`  (say we have `var t time.Time`) results in undefined for before year 1678 or after year 2262 (limitations of int64). So naturally you can't get back zero time using `time.Unix(sec int64, nsec int64)` 

I think it can be achieved by using `Sub(d duration)`, but all that pain doesn't seem worth taking.

hsanjuan: Concentrate on the point of checking that the "unset" time you get out of the protobuf can be identified as an unset time. If IsZero()  != time.Unix(0,0), then the line above is buggy which is exactly why I asked to define ZeroTime exactly as what protobuf considers as a zero/unset time.

cluster.go
```diff
@@ -1046,13 +1046,13 @@ func (c *Cluster) isClosest(ctx context.Context, cid cid.Cid) (bool, error) {
...
 
-	cidKey := cid.KeyString()
+	cidKey := string(kbucket.ConvertKey(cid.Hash().B58String()))
 	for _, p := range peers {
 		if p == c.id || !c.consensus.IsTrustedPeer(ctx, p) {
 			continue
 		}
 
-		if !kbucket.Closer(p, c.id, cidKey) {
+		if !kbucket.Closer(c.id, p, cidKey) {
```
hsanjuan: Have a look to `Closer` implementation. It already converts keys. Have you seen `SortClosestPeers()`? We just need to closest peer so maybe you can just use other available functions to sort depending on what the library exposes.

Please study what a `Cid` is, the cost of Hash().B58Stringing(). Why would you need a B58 encoded string at all? If anything you need Bytes() (and then cast to string as necessary).

cluster.go
```diff
@@ -1004,21 +1006,18 @@ func (c *Cluster) StateSync(ctx context.Context) error {
...
-			var pinObj api.Pin
-			err := c.rpcClient.CallContext(
-				ctx,
-				getClosestTrustedPeer(ctx, pCid.String(), currentPin.Allocations, c.consensus),
-				"Cluster",
-				"Unpin",
-				currentPin,
-				&pinObj,
-			)
+		if !currentPin.PinOptions.ExpireAt.Round(0).Equal(api.ZeroTime) && currentPin.PinOptions.ExpireAt.Before(timeNow) {
```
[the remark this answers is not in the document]
hsanjuan: Actually they are different so you need to verify/adjust things before converting to protobuf which probably amount to `if t.IsZero() { proto.ExpireAt = ProtobufZeroTime }`. And you can do the reverse when parsing protobuf and simplify things.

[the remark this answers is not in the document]
hsanjuan: First of all, do study if it makes a different to pass an unitialized time.Time  to protobuf, or a zero time as protobuf understands  it (unix(0,0)) or simply leaving the protobuf timestamp as nil. when encoding/decoding. I don't have time to look into all this. Can you answer that question?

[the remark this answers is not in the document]
kishansagathiya: Protobuf understands uninitialized time as zero time, won't leave it nil.

cluster.go
kishansagathiya: Compared to `SortClosestPeers`, one we have is more efficient. We return when we find at least one peer greater than us.

api/types.go
```diff
@@ -795,12 +790,10 @@ func (pin *Pin) ProtoUnmarshal(data []byte) error {
 	// pin.UserAllocations = opts.GetUserAllocations()
 	t := opts.GetExpireAt()
```
hsanjuan: Does this ever return nil?

cluster.go
```diff
@@ -1003,6 +1006,21 @@ func (c *Cluster) StateSync(ctx context.Context) error {
...
+			if isClosest {
+				logger.Infof("StateSync: unpinning %s, pin expired at %s", pCid, currentPin.PinOptions.ExpireAt.Format(time.RFC3339))
+				_, err := c.Unpin(ctx, pCid)
+				if err != nil {
+					logger.Warning(err)
+				}
+			}
+			continue
+		}
+
```
hsanjuan: This needs some sort of refactoring, it's too complex to read, and I think buggy (I think  a `continue` statement is missing after the warning?).

api/types.go
kishansagathiya: if time is zero(uninitialized), we are keeping protobuf timestamp as nil, so yes.

I can pass uninitialized time, but this saves some memory.

cluster.go
kishansagathiya: To find whether we are closest or not, in worst case we take O(n)
Go's sort function has complexity of O(nlogn)

hsanjuan: Thanks for bringing that up. Now please, benchmark it and you will see that in the worst case (where you do N comparisons), your approach is slower for at least up to 1000 peers. 

On the average case (when you do O(n/2)), your function is only slightly faster) with less than 20 peers. And only twice as fast with 1000 peers.

For any reasonable sized cluster you managed to lose all your O(n) advantage in the worst case and reduce it by 80% on the average case. How much time do I have to spend telling you the way you did it is very wrong apart from more complex? I'm awfully tired of this.

And the worst is that coming up with an O(n) solution which is 10x more performant is a matter of seconds.

hsanjuan: A few more thoughts:

* We can use a faster function than sha256 (blake2b) is twice as fast
* The average case is actually not n/2 calls to Closer(), it's less (more the probability of finding a peer closer than youself), although always one peer will hit the worse case for each cid. This makes the current solution better in comparison to SortCloserPeers, than i assumed, but still bad because of the repeated hashing.
* IsTrustedPeer checks a syncmap in crdts. We need to benchmark what is faster, to do this 1000 times or to produce a list of trusted peers up front and loop it (with the necessary interface changes to consensus).
* The hashing operation should happen at most once for each thing (need a cache for peer id hashes).
* The whole function needs to be split and refactored

This path is perf critical because this runs regularly on potentially very large pinsets.

kishansagathiya: > to produce a list of trusted peers up front and loop it 

I have used a map. It is faster, often twice as fast.

cluster.go
```diff
@@ -987,9 +988,30 @@ func (c *Cluster) StateSync(ctx context.Context) error {
...
+	peers, err := c.consensus.Peers(ctx)
+	if err != nil {
+		return err
+	}
+
+	hashCache := make(utils.HashCache)
+
+	for _, p := range peers {
+		// We are also part of trusted peers and we must avoid a comparision
+		// with otherselves. Since a peer is not closer than itself, that
```
hsanjuan: `otherselves` is not a word.
`comparision` -> `comparison`.

cluster.go
```diff
@@ -987,9 +988,30 @@ func (c *Cluster) StateSync(ctx context.Context) error {
 		}
 	}
 
+	var trustedPeers []peer.ID
+	peers, err := c.consensus.Peers(ctx)
```
hsanjuan: The logic here assumes the local peer is a trusted peer itself in the cluster. There is no way around this in the general case as we don't know who trusts us,  but `c.config.FollowerMode` indicates that this peer declares itself as not trusted by anyone else and therefore can skip the whole checking-for-expiration thing.

cluster.go
```diff
@@ -987,9 +988,30 @@ func (c *Cluster) StateSync(ctx context.Context) error {
...
+	for _, p := range peers {
+		// We are also part of trusted peers and we must avoid a comparision
+		// with otherselves. Since a peer is not closer than itself, that
+		// comparision would result in false, making us leave unpinning for
+		// some other peer.
+		if p == c.id || !c.consensus.IsTrustedPeer(ctx, p) {
+			continue
+		}
+		trustedPeers = append(trustedPeers, p)
+	}
```
hsanjuan: A helper function to get a slice of trusted peers that does not include ourselves seems like a useful thing to have.

cluster.go
```diff
@@ -1021,6 +1038,52 @@ func (c *Cluster) StateSync(ctx context.Context) error {
...
+func (c *Cluster) isClosest(ctx context.Context, ci cid.Cid, trustedPeers []peer.ID, hashCache utils.HashCache) bool {
+	ciHash := utils.ConvertKey(ci.KeyString())
+	myDistance := utils.XOR(ciHash, hashCache.ConvertPeerID(c.id))
+	for _, p := range trustedPeers {
+		if !myDistance.Less(utils.XOR(ciHash, hashCache.ConvertPeerID(p))) {
+			return false
+		}
+	}
+	return true
+}
```
hsanjuan: Would it make sense to have a 

```go
type distanceChecker struct {
}

func buildDistanceChecker(local peer.ID, otherPeers []peer.ID) distanceChecker {
}

func (dc distanceChecker) closest(cid.Cid) bool {

}
```

that wraps the whole hashcache/xoring/hashing functionality inside? I find this code still quite spaguetti (it was spaguetti before too).

hsanjuan: Before this PR I mean

utils/id_hash.go
```diff
@@ -1,4 +1,4 @@
-package keys
+package utils
```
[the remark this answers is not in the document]
kishansagathiya: `utils` module was already there, I just added a file to it. I could have put it in `./util.go`, but functions in there cannot be reusable by all module (cyclic-dependency). These functions are not used by anything else at the moment, but I thought it would be nice to have that ability.

utils/id_hash.go
```diff
@@ -0,0 +1,45 @@
...
+func XOR(a, b ID) ID {
+	return ID(u.XOR(a, b))
+}
+
+// ConvertPeerID creates a DHT ID by hashing a Peer ID (Multihash).
+func (h HashCache) ConvertPeerID(id peer.ID) ID {
+	hash, ok := h[id]
+	if !ok {
+		hashBytes := blake2b.Sum256([]byte(id))
+		hash = hashBytes[:]
```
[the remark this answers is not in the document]
kishansagathiya: it (`hash`) is being being returned at the end.

cluster.go
kishansagathiya: This reduces arguments passed to isClosest function.
We can simply call one isClosest function. Creating a struct and then calling two different functions is unnecessary extra complexity,

utils/id_hash.go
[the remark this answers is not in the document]
hsanjuan: That's because I did not realize you had created an additional module on a previous PR. Note that catch-all util modules are even officially discouraged: https://blog.golang.org/package-names

utils/id_hash.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/libp2p/go-libp2p-core/peer"
+
+	u "github.com/ipfs/go-ipfs-util"
+	blake2b "golang.org/x/crypto/blake2b"
+)
+
+// ID signifies that its contents have been hashed from either a
+// peer.ID or a util.Key. This unifies the keyspace.
+type ID []byte
```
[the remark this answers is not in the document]
kishansagathiya: The difference is insignificant

byte.Compare function takes a slice. To convert that to array, I have used copy(not sure if this could be done any other way), which could make it slower. 

benchmarks here
https://gist.github.com/kishansagathiya/d47935941e649a563e8f18a54ba6ec9b

utils/id_hash.go
```diff
@@ -0,0 +1,45 @@
...
+// XOR return xor of two byte arrays.
+func XOR(a, b ID) ID {
+	return ID(u.XOR(a, b))
+}
+
+// ConvertPeerID creates a DHT ID by hashing a Peer ID (Multihash).
+func (h HashCache) ConvertPeerID(id peer.ID) ID {
+	hash, ok := h[id]
+	if !ok {
+		hashBytes := blake2b.Sum256([]byte(id))
```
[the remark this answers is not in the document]
kishansagathiya: ConvertKey takes a string, id is not.

I could use ConvertKey(string(id)), but that would give a different hash as []byte(id) and []byte(string(id)) are not same.
I could have the same function to hash both of these things. ConvertKey(interface{}), but since we don't want to cache cid hashes (they appear only once), having ConvertKey helps.

utils/id_hash.go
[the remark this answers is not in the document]
hsanjuan: I don't see where you need to convert slices to arrays. You need to convert arrays to slices which does not require copy.

kishansagathiya:
I had added some extra commits during rebasing, had to forcepush

cluster_test.go
```diff
@@ -773,6 +773,43 @@ func TestClusterPins(t *testing.T) {
 	}
 }
 
+func TestClusterPinsWithExpiration(t *testing.T) {
```
hsanjuan: This test is more valuable if run with several clusters (as part of ipfscluster_test.go). It will cover more code.

cluster_test.go
```diff
@@ -773,6 +773,43 @@ func TestClusterPins(t *testing.T) {
...
+	pins, err := cl.Pins(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(pins) != 1 {
+		t.Error("pin should be part of the state")
+	}
+
+	// try after expiry time and state sync interval
+	time.Sleep(cl.config.StateSyncInterval + 1*time.Second)
```
hsanjuan: The StateSyncInterval is set to 1 minute, so all the test suite is going to spend 59 seconds sleeping and last 10% more thanks to this. Reduce to 1 or 2 seconds (config_test.go, hopefully doesn't break something else).

kishansagathiya: Instead of changing state sync interval for all tests, I have changed only for pin expiration test.

This would avoid statesync running too often in tests. 

Just having a global variable to do this is much less code change then being able to pass configs while creating cluster peers.

ipfscluster_test.go
```diff
@@ -2108,3 +2111,47 @@ func TestClustersFollowerMode(t *testing.T) {
 		}
 	})
 }
+
+func TestClusterPinsWithExpiration(t *testing.T) {
+	ctx := context.Background()
+
+	stateSyncInterval = 2 * time.Second
```
hsanjuan: Please don't use a global variable. You can actually trigger StateSync manually in this test. That is the simple option.

api/pb/types.proto
```diff
@@ -25,4 +27,5 @@ message PinOptions {
   reserved 5; // reserved for UserAllocations
   map<string, string> Metadata = 6;
   bytes PinUpdate = 7;
+  google.protobuf.Timestamp ExpireAt = 8;
```
hsanjuan: lets do uint64

api/types.go
```diff
@@ -573,6 +587,21 @@ func (po *PinOptions) FromQuery(q url.Values) error {
 		po.UserAllocations = StringsToPeers(strings.Split(allocs, ","))
 	}
 
+	if v := q.Get("expire-at"); v != "" {
+		var tm time.Time
+		err := tm.UnmarshalText([]byte(v))
+		if err != nil {
+			return fmt.Errorf("parameter expire-at invalid")
+		}
+		po.ExpireAt = tm
```
hsanjuan: if `tm` is after Time.Now() print a warning.

api/types.go
```diff
@@ -573,6 +587,21 @@ func (po *PinOptions) FromQuery(q url.Values) error {
...
+	} else if v = q.Get("expire-in"); v != "" {
+		d, err := time.ParseDuration(v)
```
hsanjuan: if duration is less than 1 second, then ignore and print a warning.

api/types.go
```diff
@@ -683,6 +712,16 @@ func (pin *Pin) ProtoMarshal() ([]byte, error) {
 		allocs[i] = bs
 	}
 
+	var expireAtProto *tspb.Timestamp
+	var err error
+	// Only set the protobuf field with non-zero times.
+	if !pin.ExpireAt.IsZero() {
```
hsanjuan: set time to 0 if IsZero otherwise to Unix seconds.

api/types.go
```diff
@@ -749,8 +789,17 @@ func (pin *Pin) ProtoUnmarshal(data []byte) error {
 	pin.Name = opts.GetName()
 	pin.ShardSize = opts.GetShardSize()
 	// pin.UserAllocations = opts.GetUserAllocations()
+	t := opts.GetExpireAt()
+	// Only set a non-zero time if the protobuf field was used. Note that
+	// Protobuf interprets nil as 0s0ms unix time, which is different from
+	// an IsZero() time in Go.
+	if t != nil {
```
hsanjuan: only parse unix seconds if time > 0.

api/types.go
```diff
@@ -823,6 +872,11 @@ func (pin *Pin) IsRemotePin(pid peer.ID) bool {
 	return true
 }
 
+// ExpiredAt returns whether the pin has expired at the given time.
+func (pin *Pin) ExpiredAt(t time.Time) bool {
+	return !(pin.ExpireAt.IsZero() || pin.ExpireAt.After(t))
```
hsanjuan: add an unexported constant `unixZero` for quicker comparison and check that t is not `unixZero` in which case it will not be expired. This is an extra safeguard in case someday someone ends up setting ExpireAt to Unix(0) when parsing a protobuf.

kishansagathiya:
@hsanjuan 
I think I should resume this after `removing map pintracker` (https://github.com/ipfs/ipfs-cluster/pull/944). This two PRs affect each other quite a bit. After understanding pin trackers better, 
- StateSync function depends a lot on how pintracker works. It should rather be part of the pintracker interface, if it stays.
- pin expiration should also be handled by pin tracker
- With stateless pin tracker, StateSync becomes redundant. Instead of keeping it around just for pin expiration, what we could do is check expiration when `Status` is called. If we find that a pin is expired, we call status `TrackerStatusExpired`. If we are responsible to unpin it (closest peer), call Cluster.Unpin. Otherwise, call IPFS Unpin. This assumes that StatusAll will be called regularly on peers by users. I am however not sure if that is a safe assumption to make.
wdyt?

hsanjuan:
> @hsanjuan
> I think I should resume this after `removing map pintracker` (#944). This two PRs affect each other quite a bit. After understanding pin trackers better,

For the sake of reducing the chances of breaking things and the amount of things I need to review, please finish this here.

> 
>     * StateSync function depends a lot on how pintracker works. It should rather be part of the pintracker interface, if it stays.

> 
>     * pin expiration should also be handled by pin tracker

It is part of the cluster main component because it works with any pintracker implementation. The same thing should not be implemented all over again if there are several pin tracker implementations. The StateSync function may dissapear as such though, but given that the loop does pinset maintenance I don't think it hurts to keep it where it is for the moment.

> 
>     * With stateless pin tracker, StateSync becomes redundant. Instead of keeping it around just for pin expiration, what we could do is check expiration when `Status` is called. If we find that a pin is expired, we call status `TrackerStatusExpired`. If we are responsible to unpin it (closest peer), call Cluster.Unpin. Otherwise, call IPFS Unpin. 

Status is a read operation that should not have side-effects. I see some potential improvements letting the PinTracker handle the unpinning of its own expired pins, but this can be addressed later (you are welcome to take ownership of such future improvement and open an issue, describe it and propose an approach).

> This assumes that StatusAll will be called regularly on peers by users. I am however not sure if that is a safe assumption to make.

That's a bad assumption.

>       wdyt?

I think the changes requested need about 15 minutes, so please finish this here and don' t leave it hanging waiting on a different PR that I haven't started to look into. Every PR should be as small as possible and not mix new features with refactors.

api/types.go
kishansagathiya: tm (expiry time) is supposed to be after time.Now(). Did you mean before time.Now?

hsanjuan: yes, sorry, before

kishansagathiya:
because of time to seconds conversion, deserialize(serialize(pins)) != pins. Most like this is going to fail one of the tests.

hsanjuan:
> because of time to seconds conversion, deserialize(serialize(pins)) != pins. Most like this is going to fail one of the tests.

I don't see why

RubenKelevra:
Thanks for this handy new feature, I have some questions:

How can I 'bump' the time of expire forwards on an existing pin in the cluster?

How can I add a expire time on an already existing pin?

When I update a pin with a new object, will the new object get the same timeframe to expire or the same time at which it will expire or won't it get the expire tag at all?

hsanjuan:
@RubenKelevra 

> How can I 'bump' the time of expire forwards on an existing pin in the cluster?
> How can I add a expire time on an already existing pin?

Just `pin add` again that same pin.

> When I update a pin with a new object, will the new object get the same timeframe to expire or the same time at which it will expire or won't it get the expire tag at all?

If by updating a pin you mean the `pin update` command, then everything should be copied (including expiration). The only thing that can be overriden right now on `pin update` is the pin name, when specified. Do open an issue if you think expiration should also be override-able, as this should be easy to add.

RubenKelevra:
> @RubenKelevra
> 
> > How can I 'bump' the time of expire forwards on an existing pin in the cluster?
> > How can I add a expire time on an already existing pin?
> 
> Just `pin add` again that same pin.

This won't move any data inside the cluster when I just `pin add` with just the `expire-in`?

Will it change any other metadata like the name or replication to a default value if I don't respecify it?

Do I need to unpin the pin before I run `pin add`?

Sorry for the questions, there's just no documentation atm and I want to make sure I run the commandos correctly without loosing data :)

hsanjuan:
> This won't move any data inside the cluster when I just pin add with just the expire-in?

It won't move any data as since the pin exists it should get the same allocations, but you need to provide the same options you provided with the original pin (name, custom replication factors).

> Do I need to unpin the pin before I run pin add?

No

By the way, just to make sure, you are aware that PinUpdate is to add a new pin based on a previous one, not to update an existing one.

RubenKelevra:
> > This won't move any data inside the cluster when I just pin add with just the expire-in?
> 
> It won't move any data as since the pin exists it should get the same allocations, but you need to provide the same options you provided with the original pin (name, custom replication factors).

Thanks for the clarification, I hoped for exactly that!

> > Do I need to unpin the pin before I run pin add?
> 
> No

Alright!

> By the way, just to make sure, you are aware that PinUpdate is to add a new pin based on a previous one, not to update an existing one.

Thanks for the heads-up, I was aware of this. Thought about using this to replace large files which for added with the rolling hashsum option, to reduce the update size, if they need to be updated in place.

Since I want to keep the new version around as long as the old version (when it was added), I need currently to run an pin-add afterwards.

hsanjuan:
> Since I want to keep the new version around as long as the old version (when it was added), I need currently to run an pin-add afterwards.

Note that PinUpdate does not unpin the original pin, it creates a new one only:

```
ipfs-cluster-ctl pin update --help
NAME:
   ipfs-cluster-ctl pin update - Pin a new item based on an existing one

USAGE:
   ipfs-cluster-ctl pin update [command options] <existing-CID> <new-CID|Path>

DESCRIPTION:
   
This command will add a new pin to the cluster taking all the options from an
existing one, including name. This means that the new pin will bypass the
allocation process and will be allocated to the same peers as the existing
one.

The cluster peers will try to Pin the new item on IPFS using the "pin update"
command. This is especially efficient when the content of two pins (their DAGs)
are similar.

Unlike the "pin update" command in the ipfs daemon, this will not unpin the
existing item from the cluster. Please run "pin rm" for that.


OPTIONS:
   --no-status, --ns                 Prevents fetching pin status after unpinning (faster, quieter)
   --wait, -w                        Wait for all nodes to report a status of pinned before returning
   --wait-timeout value, --wt value  How long to --wait (in seconds), default is indefinitely (default: 0s)
```

RubenKelevra:
> > Since I want to keep the new version around as long as the old version (when it was added), I need currently to run an pin-add afterwards.
> 
> Note that PinUpdate does not unpin the original pin, it creates a new one only:
> 
> ```
> ipfs-cluster-ctl pin update --help
> NAME:
>    ipfs-cluster-ctl pin update - Pin a new item based on an existing one
> 
> USAGE:
>    ipfs-cluster-ctl pin update [command options] <existing-CID> <new-CID|Path>
> 
> DESCRIPTION:
>    
> This command will add a new pin to the cluster taking all the options from an
> existing one, including name. This means that the new pin will bypass the
> allocation process and will be allocated to the same peers as the existing
> one.
> 
> The cluster peers will try to Pin the new item on IPFS using the "pin update"
> command. This is especially efficient when the content of two pins (their DAGs)
> are similar.
> 
> Unlike the "pin update" command in the ipfs daemon, this will not unpin the
> existing item from the cluster. Please run "pin rm" for that.
> 
> 
> OPTIONS:
>    --no-status, --ns                 Prevents fetching pin status after unpinning (faster, quieter)
>    --wait, -w                        Wait for all nodes to report a status of pinned before returning
>    --wait-timeout value, --wt value  How long to --wait (in seconds), default is indefinitely (default: 0s)
> ```

I've decided to switch from a default timeout which I move forward in time when the files are still needed, to an "I will add a pin-expire date when I don't need the files anymore".

This way the cluster will wait a while until the pins are removed, to make sure any operation currently running (pinning by a client in the network) can continue and I also can keep a short history.

I maintain the file structure for the pins with `ipfs files cp` and `ipfs files rm` directly, while adding the files itself with `ipfs-cluster-ctl`.

When a file I know can be deduped will be replaced by a new version, I can simply update it, and add the pin-expire date to the old CID and remove the old CID from the filesystem and replace it with the new one.


Hope this all makes sense, I'm still just a week or two into reading the docs at this time. :)

PR #925, opened 2019-09-21
Improve connection graph output

### LINKED ISSUE #800: Cluster graph command should distinguish between trusted and untrusted peers

The output of `ipfs-cluster-ctl health graph` should distinguish between which peers are in the trusted peerset and which are not.

kishansagathiya:
- Distinguish between trusted nodes (dark green) and untrusted nodes
(light green)
- Distinguish between Cluster nodes (filled) and IPFS nodes (hollow)
- Straight edge between Cluster node and it's corresponding IPFS node
- Cluster nodes always on top row, IPFS nodes always on bottom row
- Shorter labels for nodes

Fixes #800

### DIFF
--- a/cmd/ipfs-cluster-ctl/graph.go
+++ b/cmd/ipfs-cluster-ctl/graph.go
@@ -6,8 +6,8 @@ import (
 	"io"
 	"sort"
 
+	dot "github.com/kishansagathiya/go-dot"
 	peer "github.com/libp2p/go-libp2p-peer"
-	dot "github.com/zenground0/go-dot"
 
 	"github.com/ipfs/ipfs-cluster/api"
 )
@@ -31,8 +31,10 @@ import (
 type nodeType int
 
 const (
-	tCluster nodeType = iota // The cluster node type
-	tIpfs                    // The IPFS node type
+	tSelfCluster    nodeType = iota //  cluster self node
+	tCluster                        //  cluster node
+	tTrustedCluster                 //  trusted cluster node
+	tIPFS                           //  IPFS node linked to a Cluster node
 )
 
 var errUnfinishedWrite = errors.New("could not complete write of line to output")
@@ -62,6 +64,9 @@ func makeDot(cg *api.ConnectGraph, w io.Writer, allIpfs bool) error {
 	dW := dotWriter{
 		w:                w,
 		dotGraph:         dot.NewGraph("cluster"),
+		self:             peer.IDB58Encode(cg.ClusterID),
+		trustMap:         cg.ClusterTrustLinks,
+		idToPeername:     cg.IDtoPeername,
 		ipfsEdges:        ipfsEdges,
 		clusterEdges:     cg.ClusterLinks,
 		clusterIpfsEdges: cg.ClustertoIPFS,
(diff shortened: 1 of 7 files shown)


cmd/ipfs-cluster-ctl/graph.go
```diff
@@ -6,8 +6,8 @@ import (
 	"io"
 	"sort"
 
+	dot "github.com/kishansagathiya/go-dot"
```
kishansagathiya:
```suggestion
	dot "github.com/zenground0/go-dot"
```

after https://github.com/ZenGround0/go-dot/pull/4 gets in

hsanjuan:
Per comment at https://github.com/ipfs/ipfs-cluster/issues/800#issuecomment-532171201, the main thing I don't like are the peer names. I think the labels should say: `IPFS: Qm*abc` (or similar). In the case of cluster, it would be interesting to use the Cluster `peername` when available. Other than that, it's looking really nice!

hsanjuan:
@kishansagathiya can I get a new screenshot?

kishansagathiya:
@hsanjuan 
Did you mean a sample graph generated from this change?

g

kishansagathiya:
> Have we tried how it looks with square boxes rather than circles? Might look better.

I had tried it. It did not look good.

hsanjuan:
> > Have we tried how it looks with square boxes rather than circles? Might look better.
> 
> I had tried it. It did not look good.

ok, no problem

kishansagathiya:
> If the ipfs daemon is not running it should display something like a orange blob with an error label (right now it simplu does not show it)

I tried this. Orange does match with the color scheme that I picked. I have gone with a empty ellipse for disconnected IPFS peer (linked to its cluster peer with dotted line), and filled for connected IPFS daemon. I can fill disconnected IPFS peers with orange/red color, but I think keeping it empty send a more clear message that it's not there for there for now.

> The black font on the dark green cluster peer might look better as white (and bold?). Might we worth using a color scheme where all colors fit as well (lots of color schemes generators online)

I have picked one of the schemes provided by dot lang, brbg11. It is simple and serves our purpose. Just two colours with various intensity. Times Roman isn't very nice in bold, so changed to Ariel.

> The self node should have the same type of label as the rest (name+iD). Maybe the border of the box should be in a different color (if possible, otherwise the box itself).

There isn't any attribute for that. I have kept double borders for self peer.

g

kishansagathiya:
One more example.
Light green is for untrusted
light brown is for swarm ipfs peer that are not linked to any cluster peer.

g

hsanjuan:
IPFS peers which are not there need to appear as error boxes because they should be there..

kishansagathiya:
Error IPFS nodes in box shape
g

hsanjuan:
Then make all ipfs daemons boxes?

hsanjuan:
I've mentioned this to @ericronne, see if someone from design has some ideas on how things look best.

cmd/ipfs-cluster-ctl/graph.go
```diff
@@ -78,53 +84,138 @@ type dotWriter struct {
...
+		node.FontColor = "6"
 		dW.clusterNodes[id] = &node
-	case tIpfs:
+	case tIPFS:
+		node.ID = fmt.Sprintf("I%d", len(dW.ipfsNodes))
+		node.Shape = "box"
+		ipfsID, ok := dW.clusterIpfsEdges[id]
+		if !ok {
+			node.Label = label("IPFS", "Errored")
+			node.ColorScheme = "X11"
```
hsanjuan: How about limiting to colors in the `rdylbu9` color scheme

cmd/ipfs-cluster-ctl/graph.go
```diff
@@ -31,8 +31,11 @@ import (
 type nodeType int
 
 const (
-	tCluster nodeType = iota // The cluster node type
-	tIpfs                    // The IPFS node type
+	tSelfCluster    nodeType = iota //  cluster self node
+	tCluster                        //  cluster node
+	tTrustedCluster                 //  trusted cluster node
+	tIPFS                           //  IPFS node linked to a Cluster node
+	tSwarmIPFS                      //  IPFS node linked to a tIPFS node
```
hsanjuan: What are tSwarmPeers vs tIPFS ? Am I understanding it right that you are printing nodes in the IPFS swarm that are not connected to any cluster peer? Does this mean that if I run connect graph on an IPFS peer connected to 400 other ipfs peers in the network I am going to get 400 boxes? If so, please remove. We're only interested in ipfs peers attached to cluster peers.

cmd/ipfs-cluster-ctl/graph.go
hsanjuan: (and only using colors from that scheme, for all things)

cmd/ipfs-cluster-ctl/graph.go
```diff
@@ -78,53 +84,138 @@ type dotWriter struct {
...
 	default:
 		return errUnknownNodeType
 	}
-	dW.dotGraph.AddVertex(&node)
 
+	graph.AddVertex(&node)
 	return nil
 }
 
+func shorten(id string) string {
```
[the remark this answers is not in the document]
kishansagathiya: that would return something like `<peer.ID p>`
It makes the string unnecessarily long, making the graph bigger and uglier.
I can how make the actual shorten peer id consistent by using `pid[:2] + "*" +pid[len(pid)-6:]`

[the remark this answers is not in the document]
hsanjuan: yes

cmd/ipfs-cluster-ctl/graph.go
kishansagathiya: It doesn't have white color. We would have to rely on light yellow or light blue for for fonts in darker ellipse/boxes, which I think would not be as good as white.

kishansagathiya:
I am leaving test modification for later.

hsanjuan:
When I test this I don't see the connections between the ipfs daemons in the cluster. I think you removed too much in your last commit. Also, let's make boxes for the cluster peers. Boxes looks better

kishansagathiya:
> When I test this I don't see the connections between the ipfs daemons in the cluster. I think you removed too much in your last commit.

I tried twice just now, and I see them connected.
g

Just a guess, did you modify your godot for testing and left it changed?

hsanjuan:
I think this is nice enough:


image

Let's move on on this

hsanjuan:
@kishansagathiya it seems tests failed randomly depending on the order of the map. I fixed that on other branch.

PR #930, opened 2019-09-27
Update/deps4

hsanjuan:
Update deps. Brings the adder functionality in line with the newer versions of go-ipfs-files and aligns the `ipfsadd` part to how go-ipfs does it.


adder/ipfsadd/add.go
```diff
@@ -283,13 +244,80 @@ func (adder *Adder) addNode(node ipld.Node, path string) error {
 	return nil
 }
 
-// AddFile adds the given file while respecting the adder.
-func (adder *Adder) AddFile(path string, file files.Node) error {
-	return adder.addFile(path, file)
+// AddAllAndPin adds the given request's files and pin them.
+// Cluster: we don'pin. Former AddFiles.
+func (adder *Adder) AddAllAndPin(file files.Node) (ipld.Node, error) {
```
kishansagathiya: Why are we naming functions with `pin` if the don't pin?

hsanjuan: This is a copy of a file from IPFS. The less differences between the two, the easier it is to maintain it. Ideally the diff between to two should be as small as possible. If we change function names, and then IPFS changes function names it becomes harder to figure out what was where.

adder/ipfsadd/add.go
```diff
@@ -102,25 +97,26 @@ func (adder *Adder) add(reader io.Reader) (ipld.Node, error) {
...
 }
 
-// RootNode returns the root node of the Added.
-func (adder *Adder) RootNode() (ipld.Node, error) {
-	// for memoizing
-	if adder.root != nil {
-		return adder.root, nil
-	}
-
+// RootNode returns the mfs root node
```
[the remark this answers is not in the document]
hsanjuan: needs fixing on ipfs first.

PR #932, opened 2019-09-28
Enable P2P Circuit relay and QUIC

### LINKED ISSUE #614: Improve experience with NAT'ed/Firewalled peers
#### Basic information

  * [x] Master
  * [x] Latest stable version
  * [ ] Bug
  * [ ] Feature request
  * [x] Enhancement

#### Description

Since we enabled DHT discovery we stopped manually opening connections to every peer when joining (which was a mess). A downside is that a firewalled peer can only join by talking directly to the leader and in some cases, by manually opening connections to the rest of peers in the cluster.

We should:

* Start testing QUIC and see if it improves NAT hole-punching
* Study libp2p p2p-circuit (relay) which is now enabled by default but I don't know if something else
is needed. In this case, it should enable any peer to reach firewalled peers through the peer it first contacted.

The scenario that should work is 2 peers online, reachable on the internet, and peers behind NAT or with incoming ports filtered should be able to join the cluster, and be reached by any other peer in it (even other firewalled peers on different networks).

kishansagathiya:
This PR
- enable p2p circuit relay with hop and discovery (become a relay and also keep finding other relays) and autorelay
- updates raft to 1.1.1

Fixes #614


kishansagathiya:
This was tested with 1 peer not behind NAT and 2 behind NAT

I was able to see logs mention opening of relay stream between two peers behind NAT
```
DEBUG [0;33mnet/identi: [0m/ipfs/id/1.0.0 sent message to 12D3KooWAMNwdgsGv8WR1UajmnwbxSt1XMx1A7jpdUHZ26uLgV1Z /ip4/192.168.225.211/tcp/9096 [0;37mid.go:181[0m
INFO [0;33m     relay: [0mnew relay stream from: 12D3KooWAMNwdgsGv8WR1UajmnwbxSt1XMx1A7jpdUHZ26uLgV1Z [0;37mrelay.go:242[0m
DEBUG [0;33mnet/identi: [0m/ipfs/id/1.0.0 received message from 12D3KooWAMNwdgsGv8WR1UajmnwbxSt1XMx1A7jpdUHZ26uLgV1Z /ip4/192.168.225.211/tcp/9096 [0;37mid.go:197[0m
DEBUG [0;33mnet/identi: [0midentify identifying observed multiaddr: /ip4/0.0.0.0/tcp/9096 [/p2p-circuit /ip4/127.0.0.1/tcp/9096 /ip4/192.168.225.26/tcp/9096 /ip4/172.17.0.1/tcp/9096 /ip4/172.18.0.1/tcp/9096 /ip4/172.19.0.1/tcp/9096] [0;37mid.go:516[0m
DEBUG [0;33mnet/identi: [0madded own observed listen addr: /ip4/0.0.0.0/tcp/9096 --> /ip4/192.168.225.26/tcp/9096 [0;37mid.go:528[0m
DEBUG [0;33mnet/identi: [0m12D3KooWMs8CeKWWQhYVkLNaHPqunaJMv37YmxQLcTbH3seHkCQ8 received listen addrs for 12D3KooWAMNwdgsGv8WR1UajmnwbxSt1XMx1A7jpdUHZ26uLgV1Z: [/ip4/127.0.0.1/tcp/9096 /ip4/192.168.225.211/tcp/9096] [0;37mid.go:359[0m
DEBUG [0;33m basichost: [0mhost 12D3KooWMs8CeKWWQhYVkLNaHPqunaJMv37YmxQLcTbH3seHkCQ8 finished dialing 12D3KooWAMNwdgsGv8WR1UajmnwbxSt1XMx1A7jpdUHZ26uLgV1Z [0;37mbasic_host.go:577[0m
DEBUG [0;33m    swarm2: [0m[12D3KooWMs8CeKWWQhYVkLNaHPqunaJMv37YmxQLcTbH3seHkCQ8] opening stream to peer [12D3KooWAMNwdgsGv8WR1UajmnwbxSt1XMx1A7jpdUHZ26uLgV1Z] [0;37mswarm.go:292[0m
DEBUG [0;33m     relay: [0mDiscovered hop relay 12D3KooWAMNwdgsGv8WR1UajmnwbxSt1XMx1A7jpdUHZ26uLgV1Z [0;37mnotify.go:45[0m
```

I was able to run status from a peer behind NAT and get all responses
I was able to add content. I was able to BlockPut content from one peer behind NAT to another one behind NAT.
Even after closing the peer not behind NAT(relay peer), one peer behind NAT was able to get Status from the other one behind NAT.

hsanjuan:
avoid the dependency update in this PR (I'm working on a major dep update)

kishansagathiya:
TODO
- [ ] Add config options
- [ ] Check if we need Auto NAT service in cluster

clusterhost.go
```diff
@@ -28,13 +32,41 @@ func NewClusterHost(
...
+		}),
+		libp2p.EnableRelay(relayOpts...),
+		libp2p.EnableAutoRelay(),
+	}
+
+	for _, addr := range cfg.ListenAddr {
+		out := false
+		for _, protocol := range addr.Protocols() {
+			if protocol.Code == ma.P_QUIC {
+				opts = append(opts, libp2p.Transport(libp2pquic.NewTransport))
```
hsanjuan: I think you are replacing all transports with QUIC. Use `ChainOptions` to prepend the quick transport to the default transports.

clusterhost.go
```diff
@@ -28,13 +32,41 @@ func NewClusterHost(
...
+		libp2p.Routing(func(h host.Host) (routing.PeerRouting, error) {
+			return dht.New(ctx, h)
+		}),
+		libp2p.EnableRelay(relayOpts...),
+		libp2p.EnableAutoRelay(),
+	}
+
+	for _, addr := range cfg.ListenAddr {
+		out := false
+		for _, protocol := range addr.Protocols() {
```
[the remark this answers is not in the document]
kishansagathiya: ipfs provides a way to disable QUIC. But I thought instead of instead of a config element we can provide an implicit way like this.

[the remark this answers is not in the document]
hsanjuan: I think we can leave it enabled in any case at this point.

clusterhost.go
```diff
@@ -46,20 +45,7 @@ func NewClusterHost(
...
-			if protocol.Code == ma.P_QUIC {
-				opts = append(opts, libp2p.Transport(libp2pquic.NewTransport))
-				out = true
-				break
-			}
-		}
-		if out {
-			break
-		}
+		libp2p.ChainOptions(libp2p.DefaultTransports, libp2p.Transport(libp2pquic.NewTransport)),
```
hsanjuan: PREPEND quic so it is used preferentially

hsanjuan: Do this is `newHost` so we run the tests with it.

cluster_config.go
```diff
@@ -24,6 +25,8 @@ const configKey = "cluster"
 // Configuration defaults
 const (
 	DefaultListenAddr          = "/ip4/0.0.0.0/tcp/9096"
+	DefaultQUICListenAddr      = "/ip4/0.0.0.0/udp/9096/quic"
```
hsanjuan: Make `DefaultListenAddrs`

ipfscluster_test.go
```diff
@@ -282,12 +282,13 @@ func createHosts(t *testing.T, clusterSecret []byte, nClusters int) ([]host.Host
 	pubsubs := make([]*pubsub.PubSub, nClusters, nClusters)
 	dhts := make([]*dht.IpfsDHT, nClusters, nClusters)
 
-	listen, _ := ma.NewMultiaddr("/ip4/127.0.0.1/tcp/0")
+	addr, _ := ma.NewMultiaddr("/ip4/127.0.0.1/tcp/0")
+	quicAddr, _ := ma.NewMultiaddr("/ip4/127.0.0.1/tcp/0/quic")
```
hsanjuan: `udp`

cluster_config.go
```diff
@@ -384,13 +402,18 @@ func (cfg *Config) applyConfigJSON(jcfg *configJSON) error {
...
 
-	clusterAddr, err := ma.NewMultiaddr(jcfg.ListenMultiaddress)
-	if err != nil {
-		err = fmt.Errorf("error parsing cluster_listen_multiaddress: %s", err)
-		return err
+	var listenAddrs []ma.Multiaddr
+	for _, addr := range jcfg.ListenMultiaddress {
+		listenAddr, err := ma.NewMultiaddr(addr)
+		if err != nil {
+			err = fmt.Errorf("error parsing a listen_multiaddress: %s", err)
```
hsanjuan: include the addr in the error so the user knows which one

cluster_config_test.go
```diff
@@ -17,7 +19,10 @@ var ccfgTestJSON = []byte(`
              "low_water": 500,
              "grace_period": "100m0s"
         },
-        "listen_multiaddress": "/ip4/127.0.0.1/tcp/10000",
+        "listen_multiaddress": [
+			"/ip4/127.0.0.1/tcp/10000",
+			"/ip4/127.0.0.1/udp/10000/quic"
+	],
```
hsanjuan: bad indent

clusterhost.go
```diff
@@ -28,13 +33,27 @@ func NewClusterHost(
...
+	relayOpts := []relay.RelayOpt{relay.OptDiscovery}
+	if cfg.EnableRelayHop {
+		relayOpts = append(relayOpts, relay.OptHop)
+	}
+
+	opts := []libp2p.Option{
+		libp2p.ListenAddrs(cfg.ListenAddr...),
+		libp2p.NATPortMap(),
+		libp2p.ConnectionManager(connman),
+		libp2p.Routing(func(h host.Host) (routing.PeerRouting, error) {
```
hsanjuan: With this the host is double-wrapped (no need to call routedHost at the end)

kishansagathiya:
Test Failure
```
➜  ipfs-cluster git:(feat/nat) ✗ go test -v -failfast -consensus crdt . --run TestAdd --loglevel DEBUG 
=== RUN   TestAdd
=== RUN   TestAdd/local
--- PASS: TestAdd (15.37s)
    --- PASS: TestAdd/local (10.39s)
=== RUN   TestAddWithUserAllocations
=== RUN   TestAddWithUserAllocations/local
--- PASS: TestAddWithUserAllocations (6.72s)
    --- PASS: TestAddWithUserAllocations/local (1.01s)
=== RUN   TestAddPeerDown
--- FAIL: TestAddPeerDown (3.36s)
    ipfscluster_test.go:356: max dial attempts exceeded
FAIL
exit status 1
FAIL	github.com/ipfs/ipfs-cluster	25.490s
➜
```
PeerAdd rpc call is failing
[debug.log](https://github.com/ipfs/ipfs-cluster/files/3744470/debug.log)

hsanjuan:
- Seems I can reproduce the error sometimes
- For simplicity I test with `-run TestClustersVersion`
- Also with `--loglevel debug --logfacs swarm2`
- From logs, the error comes when a testing peer is trying to join the first peer to bootstrap the cluster. This needs to call PeerAdd on the first peer and does it via an RPC call that needs to open a connection.
- No clear error on swarm2 level. Add a print statement for `bootstrapAddr` and `clusters[0].host.Addrs()` to see where the new peers are trying to bootstrap to.
- Peer addresses: [/ip4/127.0.0.1/tcp/43251 /ip4/127.0.0.1/udp/49230/quic]
- Bootstrap address: /ip4/127.0.0.1/tcp/43251/p2p/QmWZs2P3ULV2uToPsfQP22nsrCu1r6ShGpskNe75zVZm2w
- That looks OK-ish. What's happening when dialing? Go to https://github.com/libp2p/go-libp2p-swarm/blob/master/swarm_dial.go, check diagram.
- Notice DialAttempts set to 1. The logs show ` network for QmcjcfAe4FXSSN4BWjT779876mtQn7cACud8GnfoND9UpT finished dialing QmWZs2P3ULV2uToPsfQP22nsrCu1r6ShGpskNe75zVZm2w swarm_dial.go:228` which means an actual connection was returned.
- However https://github.com/libp2p/go-libp2p-swarm/blob/master/swarm.go#L315 means `bestConnToPeer` did not return one.
- Check bestConnToPeer() -> TODO says it won't choose transport by priority but simply choose among existing connections. Since our dialing seems to work on the first round, it would seem that `NewStream` is returning an error? Unfortunately swarm doesn't log it.
- Seems that calls OpenStream and that's transport land. go-tcp-transport uses `tcp-tpt`, quic does not seem to log at all. Re-run with `--logfacs tcp-tpt` just in case.
- Nothing new, fine. Add replace directive to go.mod `replace github.com/libp2p/go-libp2p-swarm => /home/hector/go/src/github.com/libp2p/go-libp2p-swarm`.  and log that error
- Log shows a tcp connection was stablished but NewStream returns an error with text `EOF`. Interesting... why would a stream throw EOF if the connection seems stablished, what is the only thing we have changed in this PR wrt. TCP transport?
- Comment the TLS security setting when we create the hosts. Run tests a few times and it doesn't fail. 
- Uncomment TLS again: notice also that sometimes things work but it happens to be using QUIC. Comment QUIC transport to force TCP -> sometimes things work too even though there are stream errors printed. If would seem that it retries sometimes and that can only happen because `c.conn.IsClosed()`  returns true, but for that to happen there should have been a connection in the first place.
- At this point, I am not sure of TLS only works in conjunction with quic (I think quic just bypasses it), or is broken, or there is some race. Ask on IRC #libp2p channel to get more information, wait for a bit and if no answer, then open an issue later.
- While writing the libp2p report I am force to describe how things work. Therefore I look again on how the connection is opened when we want to Join() a certain peer. I realize that we call `ImportPeer` with the peer address and set the `connect` parameter to `true`. This triggers an asnync connection attempt in a separate goroutine. This means that we are opening a stream (which will itself try to open a connection if none exists) at the same time that we are connecting. That is bad.
- I thought I had found the cause, but when I try with TLS and TCP only after setting connect to `false` in `ImportPeer` things seem to work more times than before but still get occasional errors. It seems this is a libp2p issue, therefore I write a bug report with as much useful info as I could gather: https://github.com/libp2p/go-libp2p/issues/732

hsanjuan:
Main suspect causing the errors  we see is simultaneous connection opening. The debugging here should look for how this happens. By the time PeerB opens a connection to PeerA to bootstrap it seems that PeerA has found about PeerB and is opening one back. @kishansagathiya you can have a look, I think it may be the DHT, which is bootstrapped on peer creation but even in that case I don't see how it finds the other peer.

kishansagathiya:
I ran the tests locally and they are all passing. But builds on travis are failing and will keep failing most of the times, because how logging is done in testing package. testing package doesn't log things immediately. 

In our tests we are not seeing any output for at least 10m and travis gives up

relevant issues
https://github.com/golang/go/issues/32757
https://github.com/golang/go/issues/24929

This was fixed 10 days ago, so hopefully updating go should fix it.

hsanjuan:
That is because some test is stuck and it happens only when running multi-module tests with code-coverage.

hsanjuan:
After more hardcore debugging I did find at least one problem: https://github.com/libp2p/go-libp2p-tls/pull/37 (I think there are more on the libp2p side)

clusterhost.go
```diff
@@ -72,13 +89,32 @@ func newHost(ctx context.Context, secret []byte, priv crypto.PrivKey, opts ...li
...
+	h, err := libp2p.New(
 		ctx,
 		finalOpts...,
 	)
+	if err != nil {
+		return nil, err
+	}
+
+	// need this for auto relay
+	_, err = autonat.NewAutoNATService(ctx, h, libp2p.PrivateNetwork(prot))
```
hsanjuan: This should also have tls and quic

hsanjuan:
I noticed that Autonat was created when autonat was not even used on `newHost`. As you have seen, autonat leaks a go-routine and a libp2p host so we should definitely not be leaking those on every single test * 6 peers.

hsanjuan:
This is ready but tests go way more unreliable (the new transports are probably more demanding than before). Let's merge as look for workarounds later.

kishansagathiya:
Sharness test for `ctl add` still fails, I am not yet sure why, `ctl add` works though. I am investigating why.

> As you have seen, autonat leaks a go-routine and a libp2p host so we should definitely not be leaking those on every single test * 6 peers.

You mean failures on TestClustersPeerRemoveSelf and TestClustersPeerRemoveLeader? Strange thing is that they always fail on Travis and never on local.

hsanjuan:
Sharness passes?

PR #943, opened 2019-10-21
Attempt to improve test runs on Travis

kishansagathiya:


ipfscluster_test.go
```diff
@@ -1530,90 +1530,91 @@ func TestClustersReplicationMinMaxNoRealloc(t *testing.T) {
...
-	expected := minInt(nClusters-2, 4)
-	if lenSA != expected {
-		t.Errorf("Insufficient reallocation, could have allocated to %d peers but instead only allocated to %d peers", expected, lenSA)
-	}
-
-	if lenSA < 3 {
-		t.Error("allocations should be more than rplMin")
-	}
-}
+// Repinning does not re-allocate anymore
```
kishansagathiya: https://github.com/ipfs/ipfs-cluster/blob/04a5881b073dff2fcaf6ae40a63a7ff582fda2c7/cluster.go#L1385-L1389

Shouldn't it reallocate on repin?

hsanjuan: The idea is that if you send the same pin with the same options, and the pin is already there, the operation should just try to re-pin whenever the pin was originally (similar to recover), even if the pin is in error somewhere. This is more natural than changing the allocations (Matt from Pinana convinced me). To change the allocations the replication factor can be changed, or the item unpinned and re-added.

Logic is a little brittle though and allocate() tries to respect existing allocations, so maybe it should indeed re-allocate (and should get the same allocataions whenever the allocated peers are still there).

ipfscluster_test.go
```diff
@@ -1530,90 +1530,91 @@ func TestClustersReplicationMinMaxNoRealloc(t *testing.T) {
...
+// 	alloc1 := peerIDMap[firstAllocations[0]]
+// 	alloc2 := peerIDMap[firstAllocations[1]]
+// 	safePeer := peerIDMap[firstAllocations[2]]
+
+// 	alloc1.Shutdown(ctx)
+// 	alloc2.Shutdown(ctx)
+
+// 	waitForLeaderAndMetrics(t, clusters)
+
+// 	// Repin - (although this might have been taken of if there was an alert
```
kishansagathiya: @hsanjuan It was right in front of me! alertsHandler would force reallocations.

kishansagathiya:
Let's not have `checkErr` (https://github.com/ipfs/ipfs-cluster/blob/7b499263af164883f3eeeb95964bbd36a08bfb56/ipfscluster_test.go#L130)

When an error occurs, it would point to checkErr instead of from where the error is actually coming from. like this https://travis-ci.com/ipfs/ipfs-cluster/jobs/247936708#L236

kishansagathiya:
https://github.com/ipfs/ipfs-cluster/blob/7b499263af164883f3eeeb95964bbd36a08bfb56/cluster.go#L1027-L1030

This is happening in a lot of methods. @lanzafame do you remember why this is avoid the passed `ctx` and using cluster's `c.ctx`?

lanzafame:
Because we use the context in the struct as a lifecycle management pattern, i.e. on shutdown we cancel the embedded context, we have to pass that context down, but we don't avoid the request `ctx`, we put it in the `span` and it gets propagated along for the tracing.

kishansagathiya:
increasing ready timeout would avoid a lot of errors that we see during first test (TestAdd)
30 sec -> 1 Min, It has never been changed

hsanjuan:
> increasing ready timeout would avoid a lot of errors that we see during first test (TestAdd)
> 30 sec -> 1 Min, It has never been changed

If your peers are not ready after 30 seconds then there is something going on. I'm not sure the increase fixes this. It may be by chance that you are seeing that.

hsanjuan:
I got a green light, but the Raft tests tend to fail very soon with Waiting to become a Voter or Waiting for Leader :-/ this was not like this...

hsanjuan:
Boom

hsanjuan:
Still getting some failures on the first test with Raft. It's really weird that it only fails on the tests with coverage :S

hsanjuan:
@kishansagathiya better to consolidate here and incorporate this to the other PRs. If you  find more things -> new branch/PR.

Also check if badger is still taking so much memory. I tested and it allocates way less memory after setting max_table_size without the need of having to persist things to disk. Thanks!

PR #944, opened 2019-10-23
Remove map pintracker

### LINKED ISSUE #929: Deprecate mappintracker and adopt stateless tracker
The `mappintracker` component has been used since the very early days.

It keeps an in-memory representation of the pinset along with the current PinStatus for each CID. This means that the actual status as seen by Cluster for a CID (pinned) may differ of that actual status in IPFS (unpinned). This also means that the map needs to be filled-in during cluster boot. 

To avoid these problems, the main component regularly runs a `StatusSync`. The first time it runs it will trigger a Track operation for each CID which results in a pin/ls operation on IPFS for each CID which is additionally heavy.

The `stateless` pin tracker was developed to avoid all this. Rather than keeping all the information in memory, it would be fetched from IPFS when needed. If we need to check if a pin ls pinned, we do `pin/ls` then.

Additionally, most getting-out-of-sync issues were solved so that essentially does not happen and the `sync` operation is very redundant at this point (for both trackers).

On the downside, the `stateless` pin tracker uses RPC calls to access the state, which is super innefficient and effectively loads the pinset onto memory as well by serializing and deserializing it.

Let's fix all this.

## Actionables

* [ ] Remove the mappintracker (module, tests, `service` flags)
* [ ] Remove the Sync operation from everywhere (ctl, api, pintrackers)
* [ ] Modify the stateless pintracker to use a `state.ReadOnly` so operations like `localStatus` can use that rather than calling RPC methods. The main problem here is that this is provided by the `State` methods in consensus, and this method can only be called when consensus is ready. Thus it will need to be provided in a delayed fashion (`SetState` method called appropiately from the Cluster main component (?))
* [ ] Ensure things are working like before. Overall this should simplify and remove more code than add. However sometimes some tests in `stateless` tracker fail. It is very important to understand how the `optracker` module works and how it is used to handle the pinning operation lifetime in scenarios with concurrency (very tricky). The `StateSync` operation in `cluster.go` will become a private method and should lose most of its logic (it needs to keep things like checking for pin expiration), probably it should have a different name.

kishansagathiya:
Use stateless pintracker everywhere

Fix #929


kishansagathiya:
In case of raft, `ipfs-cluster-ctl status` doesn't work well without restarting daemon once. Investigating why this happens.

kishansagathiya:
> In case of raft, `ipfs-cluster-ctl status` doesn't work well without restarting daemon once. Investigating why this happens.

This would fix it

pintracker/stateless/stateless.go
```diff
@@ -216,6 +229,13 @@ func (spt *Tracker) Shutdown(ctx context.Context) error {
 	return nil
 }
 
+// SetState sets the pin state and returns true if state was not empty or nil.
+func (spt *Tracker) SetState(st state.ReadOnly) bool {
```
hsanjuan: Why don' t you initialize the pintracker with a new function parameter:

`New(..., func() (state.ReadOnly, error)...)`

This function can block until consensus is ready, it can either be used to request GetState() from the consensus layer all the time or just once and keep it. i.e.:

```
getState := func() (state.ReadOnly, error) {
   <-cons.Ready()
   return cons.GetState()
}
```

pintracker/stateless/stateless.go
```diff
@@ -224,11 +244,8 @@ func (spt *Tracker) Track(ctx context.Context, c *api.Pin) error {
 
 	logger.Debugf("tracking %s", c.Cid)
 
-	// Sharded pins are never pinned. A sharded pin cannot turn into
-	// something else or viceversa like it happens with Remote pins so
-	// we just track them.
+	// Sharded pins are never pinned, so do nothing.
```
hsanjuan: don't rewrite comments into comments containing less information. You removed essential information about what sharded pins do: it's not they are never pinned, it's also that they never need a different operation.

pintracker/stateless/stateless.go
```diff
@@ -224,11 +244,8 @@ func (spt *Tracker) Track(ctx context.Context, c *api.Pin) error {
...
 	if c.Type == api.MetaType {
-		spt.optracker.TrackNewOperation(ctx, c, optracker.OperationShard, optracker.PhaseDone)
```
hsanjuan: ok, be very careful. You are changing the functionality of the pin tracker. Do not do that and do not break it. This code is not here for fun.

Sharded pins need to be potentially reported in the Status calls and need to be tracked. And even if this was not the case, you are not supposed to change functionality on this PR.

pintracker/stateless/stateless.go
```diff
@@ -241,12 +258,16 @@ func (spt *Tracker) Track(ctx context.Context, c *api.Pin) error {
...
 			op.SetError(err)
-		} else {
-			op.SetPhase(optracker.PhaseDone)
+			op.Cancel()
+			return nil
 		}
+
+		op.SetPhase(optracker.PhaseDone)
+		op.Cancel()
+		spt.optracker.Clean(ctx, op)
```
hsanjuan: Avoid refactorings, this is adding more lines.

Why is the operation being cleaned now?

pintracker/stateless/stateless.go
```diff
@@ -299,59 +320,65 @@ func (spt *Tracker) Status(ctx context.Context, c cid.Cid) *api.PinInfo {
...
-			logger.Error(err)
-			return &api.PinInfo{
-				Cid:    c,
-				Peer:   spt.peerID,
-				Status: api.TrackerStatusClusterError,
-				Error:  err.Error(),
-				TS:     time.Now(),
+	gpin := &api.Pin{}
+	var err error
+	if noState(spt.state) {
```
hsanjuan: avoid this. either you have a state or you have an error.

pintracker/stateless/stateless.go
```diff
@@ -299,59 +320,65 @@ func (spt *Tracker) Status(ctx context.Context, c cid.Cid) *api.PinInfo {
...
 		}
-		// not part of global state. we should not care about
-		return &api.PinInfo{
-			Cid:    c,
-			Peer:   spt.peerID,
-			Status: api.TrackerStatusUnpinned,
-			TS:     time.Now(),
+		if err != nil {
+			logger.Error(err)
+			return nil
```
hsanjuan: NEVER return nil from these methods. We will switch back to not using pointers here. Return a valid PinInfo with an error set.

pintracker/stateless/stateless.go
```diff
@@ -542,7 +444,7 @@ func (spt *Tracker) ipfsStatusAll(ctx context.Context) (map[string]*api.PinInfo,
 	ctx, span := trace.StartSpan(ctx, "tracker/stateless/ipfsStatusAll")
 	defer span.End()
 
-	var ipsMap map[string]api.IPFSPinStatus
+	ipsMap := make(map[string]api.IPFSPinStatus)
```
hsanjuan: why are you allocating a map which is immediately going to be discarded by setting the pointer to the map provided by the IPFSConnector

pintracker/stateless/stateless.go
```diff
@@ -575,23 +478,27 @@ func (spt *Tracker) ipfsStatusAll(ctx context.Context) (map[string]*api.PinInfo,
...
-		ctx,
-		"",
-		"Cluster",
-		"Pins",
-		struct{}{},
-		&statePins,
-	)
+	statePins := []*api.Pin{}
+	var err error
+	if noState(spt.state) {
```
hsanjuan: avoid this

pintracker/stateless/stateless.go
```diff
@@ -604,27 +511,27 @@ func (spt *Tracker) localStatus(ctx context.Context, incExtra bool) (map[string]
 		return nil, err
 	}
 
+	pininfos := make(map[string]*api.PinInfo)
```
hsanjuan: one thing I learnt is that make(map) takes an optional number which allows Go to be more intelligent when pre-allocating the needed memory. If this map will grow to the size of the statePins, then you can use that with make.

pintracker/stateless/stateless.go
kishansagathiya: This operation is already in PhaseDone. What TrackNewOperation does in this case
- Create the operation with phase done and at it to operations map
- it is not added to any queue (pinCh or unpinCh), so there is no task to be performed.
-  few seconds later (ticker on every 10 sec), it gets cleaned because it is already in phase done.
Why add it, if it is going to get removed without doing anything?

Metatype pins are only there for the reference to cluster dag (right?), for which it doesn't need to get pinned. It only has to be present in the shared state.

Status calls only checks the state and returns status sharded.

pintracker/stateless/stateless.go
kishansagathiya: It is already in phase done. There is nothing left to do.
If all operations are removed right after they are done, the ticker (that cleans operations every 10 seconds) can be removed.

kishansagathiya:
Remove timestamp from pininfos

kishansagathiya:
https://github.com/ipfs/ipfs-cluster/pull/951#discussion_r344444702

pintracker/stateless/mock.go
```diff
@@ -0,0 +1,74 @@
...
+type mockState struct {
+	slow bool
+}
+
+// NewMockState return a mock implementation of State.
+func NewMockState(slow bool) state.ReadOnly {
+	return mockState{slow: slow}
+}
+
+func (st mockState) List(ctx context.Context) ([]*api.Pin, error) {
```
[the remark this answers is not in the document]
kishansagathiya: > slow does not make things slow.

`slow` only make pin and unpin slow, which is done by rpc client. But normal rpc client and slow rpc client returns different results on `Pins` and `PinGet`. `slow` var here is to tell which result should be sent.

pintracker/stateless/stateless.go
kishansagathiya: I have reverted this.

pintracker/stateless/stateless.go
kishansagathiya: I have used getState to request GetState() from the consensus layer  all the time

pintracker/stateless/stateless.go
```diff
@@ -72,6 +84,7 @@ func New(cfg *Config, pid peer.ID, peerName string) *Tracker {
 // Used for both pinning and unpinning
 func (spt *Tracker) opWorker(pinF func(*optracker.Operation) error, opChan chan *optracker.Operation) {
 	logger.Debug("entering opworker")
+	// stop tracking operation shard and remove this ticker
```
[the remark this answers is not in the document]
kishansagathiya: removed this comment

pintracker/stateless/stateless.go
```diff
@@ -23,18 +23,28 @@ import (
 
 var logger = logging.Logger("pintracker")
 
+var (
+	// ErrNoChannel indicates that there doesn't exist a channel for given operation.
+	ErrNoChannel = errors.New("operation doesn't have an associated channel")
```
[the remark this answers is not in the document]
kishansagathiya: removed this error

pintracker/stateless/stateless.go
kishansagathiya: changed this. Any already existing `return nil` is kept as it is.

pintracker/stateless/mock.go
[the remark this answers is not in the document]
kishansagathiya: I guessed that this would be comparatively big. I am doing this separately in https://github.com/ipfs/ipfs-cluster/pull/958, to keep this PR small and keep it just about removing mappintracker.

pintracker/stateless/stateless.go
```diff
@@ -270,7 +283,6 @@ func (spt *Tracker) StatusAll(ctx context.Context) []*api.PinInfo {
 
 	pininfos, err := spt.localStatus(ctx, true)
 	if err != nil {
-		logger.Error(err)
```
kishansagathiya: localStatus logs all errors.

hsanjuan:
I made a few changes and I think it's good to go.

PR #946, opened 2019-10-28
Dockerfiles: download modules in advance

hsanjuan:
This should speed up Dockerfile builds, particularly when building after
updating the code since the modules will not be to be re-downloaded.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,11 +6,8 @@ MAINTAINER Hector Sanjuan <hector@protocol.ai>
 ENV GOPATH     /go
 ENV SRC_PATH   $GOPATH/src/github.com/ipfs/ipfs-cluster
 ENV GO111MODULE on
+# ENV CGO_ENABLED 0
 ENV GOPROXY=https://proxy.golang.org
-
-COPY . $SRC_PATH
-WORKDIR $SRC_PATH
-RUN make install
 
 ENV SUEXEC_VERSION v0.2
 ENV TINI_VERSION v0.16.1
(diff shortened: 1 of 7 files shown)


cmd/ipfs-cluster-service/Makefile
```diff
@@ -4,7 +4,7 @@ SRC := $(shell find .. -type f -name '*.go')
 all: ipfs-cluster-service
 
 ipfs-cluster-service: $(SRC)
-	go build -ldflags "-X main.commit=$(shell git rev-parse HEAD)"
+	go build -mod=readonly -ldflags "-X main.commit=$(shell git rev-parse HEAD)"
```
kishansagathiya: :+1:

Dockerfile
```diff
@@ -6,12 +6,9 @@ MAINTAINER Hector Sanjuan <hector@protocol.ai>
 ENV GOPATH     /go
 ENV SRC_PATH   $GOPATH/src/github.com/ipfs/ipfs-cluster
 ENV GO111MODULE on
+# ENV CGO_ENABLED 0
```
kishansagathiya: remove this? multiple occurrences.

Dockerfile
```diff
@@ -27,6 +24,14 @@ RUN set -x \
 # Get the TLS CA certificates, they're not provided by busybox.
 RUN apt-get update && apt-get install -y ca-certificates
 
+COPY go.* go.* $SRC_PATH/
```
kishansagathiya: why is same source path(`go.*`) passed twice?

PR #947, opened 2019-10-29
"Sync/Status/Recover" operations should not broadcast to peers with remote pins

### LINKED ISSUE #935: "Sync/Status/Recover" operations should not broadcast to peers with remote pins

Right now these operations do a full broadcast to all peers even though they know that for certain pins, the item is only allocated to some of them so only those can take any action.

**Additional context**

Double-check what happens when calling those and pins are remote (whether it effectively becomes a noop).

kishansagathiya:
Fix #935


kishansagathiya:
> Let's return an error when the pin is not part of the state for the moment.

Why? 

I have a different view because 
- That changes behaviour of status
- `ctl unpin` will through an error even when is succeeds, because we call status after unpin.

hsanjuan:
> > Let's return an error when the pin is not part of the state for the moment.
> 
> Why?
> 
> I have a different view because
> 
>     * That changes behaviour of status
> 
>     * `ctl unpin` will through an error even when is succeeds, because we call status after unpin.

ok, fair enough, let's leave things are they are

kishansagathiya:
There is one more thing that is wrong here. peers from which we could not get status because of rpc authorization are set as status remote. Previously, they were not added to global pin info. Shouldn't we add an RPC error status for it?

hsanjuan:
> There is one more thing that is wrong here. peers from which we could not get status because of rpc authorization are set as status remote. Previously, they were not added to global pin info. Shouldn't we add an RPC error status for it?

Do not change that behaviour here. This PR is not about that. If there's an improvement to be made -> issue for discussion -> new concrete PR.

kishansagathiya:
https://github.com/ipfs/ipfs-cluster/pull/951#discussion_r344444702

PR #950, opened 2019-11-06
Fix #949: Cluster peers should advertise their external earlier

### LINKED ISSUE #949: Cluster peers in Docker or behind NATs do not advertise their public IP addresses
**Describe the bug:**

Cluster peers in Docker or behind NATs do not advertise their public IP addresses.

https://discuss.ipfs.io/t/missing-public-multiaddresses-in-peerstore-docker/6588

https://godoc.org/github.com/libp2p/go-libp2p/p2p/protocol/identify#pkg-constants

Unfortunately ActivationThreshold is a constant. I'm not sure if #932 actually fixes this.

hsanjuan:
This lowers the activation threshold for an observed addressed
to 1 ~~and increases the OwnObservedAddressTTL to infinite~~.

This should let cluster peers (particularly in small clusters)
advertise their known external addresses asap so that other peers
can include them in their peerstores (and save them on shutdown).

By default this only happens when an external address has received 4
connections in the last 40 minutes. In NATed environments this may
result in cluster peers advertising right away their NAT-translated address/ports
which may not be dialable, but in other cases that address seems to be
indeed dialable (docker).

This should be complimented with the chance to let the user specify Announce, NoAnnounce and AddrFilters like IPFS.


hsanjuan:
Docker testing image: `ipfs/ipfs-cluster:949-test`

hsanjuan:
Perhaps we should expire OwnObservedAddresses in any case, as it seems peers will already learn and add the addresses to the peerstore during identify, there is no need to advertise them forever...

PR #951, opened 2019-11-06
Avoid out of index error in failover

kishansagathiya:


api/rest/client/lbclient.go
```diff
@@ -44,20 +45,23 @@ func (r *RoundRobin) SetClients(cl []Client) {
 	r.length = uint32(len(cl))
 }
 
-// Failover is a load balancing strategy that would try the local cluster peer first.
-// If the local call fail it would try other client in a round robin fashion.
+// Failover is a load balancing strategy that would try the local cluster peer
```
hsanjuan: it's not the local, it's the first client. Whether that's a local peer or not is not the LB client problem...

api/rest/client/lbclient.go
```diff
@@ -44,20 +45,23 @@ func (r *RoundRobin) SetClients(cl []Client) {
...
+// round robin fashion.
 type Failover struct {
 	clients []Client
+	length  int
 }
 
 // Next returns the next client to be used.
 func (f *Failover) Next(count int) Client {
-	return f.clients[count]
+	return f.clients[count%f.length]
```
hsanjuan: why can't you `count%len(f.clients)` and save a struct field?

api/rest/client/lbclient.go
```diff
@@ -77,7 +81,7 @@ func NewLBClient(strategy LBStrategy, cfgs []*Config, retries int) (Client, erro
 
 // retry tries the request until it is successful or tries `lc.retries` times.
 func (lc *loadBalancingClient) retry(count int, call func(Client) error) error {
-	logger.Debugf("retrying %d times", count+1)
+	logger.Infof("retrying %d times", count+1)
```
hsanjuan: this should still be a debug statement.

hsanjuan:
Add tests please.

api/rest/client/lbclient.go
kishansagathiya: it save us from calling len() several times.

hsanjuan: https://blog.golang.org/go-slices-usage-and-internals

calling `len` is precisely reading a length field in an struct.

kishansagathiya:
Hopefully latest commit would remove this 
```
--- FAIL: TestClustersPin (2.33s)

panic: runtime error: invalid memory address or nil pointer dereference [recovered]

	panic: runtime error: invalid memory address or nil pointer dereference

[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xfafaae]

goroutine 266740 [running]:

testing.tRunner.func1(0xc05163c500)

	/home/travis/.gimme/versions/go1.13.4.linux.amd64/src/testing/testing.go:874 +0x3a3

panic(0x11447a0, 0x1f8e0a0)

	/home/travis/.gimme/versions/go1.13.4.linux.amd64/src/runtime/panic.go:679 +0x1b2

github.com/ipfs/ipfs-cluster.(*Cluster).StateSync(0xc05acd23c0, 0x1609420, 0xc05b174330, 0x0, 0x0)

	/home/travis/gopath/src/github.com/ipfs/ipfs-cluster/cluster.go:968 +0x22e

github.com/ipfs/ipfs-cluster.(*Cluster).Join(0xc05acd23c0, 0x16093a0, 0xc0000d2010, 0x1622120, 0xc05acd5ac0, 0x0, 0x0)

	/home/travis/gopath/src/github.com/ipfs/ipfs-cluster/cluster.go:947 +0x97e

github.com/ipfs/ipfs-cluster.createClusters(0xc05163c500, 0xc001310e90, 0x44d217, 0x17942f0, 0x1e91300, 0x50e4dd, 0x0)

	/home/travis/gopath/src/github.com/ipfs/ipfs-cluster/ipfscluster_test.go:395 +0x120f

github.com/ipfs/ipfs-cluster.TestClustersPin(0xc05163c500)

	/home/travis/gopath/src/github.com/ipfs/ipfs-cluster/ipfscluster_test.go:647 +0x52

testing.tRunner(0xc05163c500, 0x147f8a0)

	/home/travis/.gimme/versions/go1.13.4.linux.amd64/src/testing/testing.go:909 +0xc9

created by testing.(*T).Run

	/home/travis/.gimme/versions/go1.13.4.linux.amd64/src/testing/testing.go:960 +0x350
```

consensus/crdt/consensus.go
```diff
@@ -428,7 +428,13 @@ func (css *Consensus) RmPeer(ctx context.Context, pid peer.ID) error {
 }
 
 // State returns the cluster shared state.
-func (css *Consensus) State(ctx context.Context) (state.ReadOnly, error) { return css.state, nil }
+func (css *Consensus) State(ctx context.Context) (state.ReadOnly, error) {
+	if css.state == nil {
+		return state.Empty(), nil
```
hsanjuan: I'd prefer if you set css.state to Empty in New during intialization. However this change seems to be completely unrelated to this PR?

hsanjuan: Also, I'm not even sure the WaitForSync in Join is necessary at this point. Maybe it can be removed in the stateless pintracker PR.

hsanjuan: Maybe an error should be returned. If someone calls State() once they get an empty state forever. The truth is the state is not setup so it should probably error.

kishansagathiya: > Maybe an error should be returned. If someone calls State() once they get an empty state forever. The truth is the state is not setup so it should probably error.

Both consensus should behave similarly if we want to use them interchangeably. Currently, if state is not set, raft returns an empty state without error, which is why I did that same for crdt.


> However this change seems to be completely unrelated to this PR?

My intention was to fix a tests from panicking, so that build would pass. But this clearly opens more problems than fix. So yes, better to do it separately.

hsanjuan: > Both consensus should behave similarly if we want to use them interchangeably.

ok

hsanjuan:
Can merge if you remove the redundant length field

PR #953, opened 2019-11-07
Fix/tests2

hsanjuan:


cmd/ipfs-cluster-ctl/graph_test.go
```diff
@@ -172,19 +162,10 @@ C2 -> I2
 /* The swarm peer connections among ipfs daemons in the cluster */
 I1 -> I0
 I1 -> I2
-I1 -> I3
```
hsanjuan: This is fishy, can you look into it @kishansagathiya ?

ipfscluster_test.go
```diff
@@ -382,6 +388,8 @@ func createClusters(t *testing.T) ([]*Cluster, []*test.IpfsMock) {
 	clusters[0] = createCluster(t, hosts[0], dhts[0], cfgs[0], stores[0], cons[0], apis[0], ipfss[0], trackers[0], mons[0], allocs[0], infs[0], tracers[0])
 	<-clusters[0].Ready()
 	bootstrapAddr := clusterAddr(clusters[0])
+
+	SetFacilityLogLevel("raftlib", "info")
```
hsanjuan: remove this

cmd/ipfs-cluster-ctl/graph_test.go
kishansagathiya: graph tests are failing because colours and colorschemes in tests are different than what we actually use. I am fixing that.

kishansagathiya: What's fishy?

Latest build is passing.

I went through a bunch of previous (5-6) builds. I did not find anything graph related except one which was failing because of some tls related error.

hsanjuan: I'm talking about having to remove some edges for the tests to pass. Is the graph not producing some edges that it should in the allIpfs mode?

hsanjuan:
please have a look to https://github.com/ipfs/ipfs-cluster/pull/953#discussion_r344066549 before approving

cmd/ipfs-cluster-ctl/graph_test.go
kishansagathiya: It shouldn't produce those edge. ipfs links that we get from connect graph are swarmpeers, but in graph we don't show them all. 
We only show links among ipfs nodes which are linked to cluster peers.
So, removing those edges from `allIpfs` was the appropriate thing to do.

hsanjuan: `makeDot` takes an `allIpfs` parameter that enables/disabled (or should enable/disable) showing edges from all ipfs peer in the swarm. This test is testing that links exist for ipfs nodes not linked to cluster peers.

hsanjuan: Therefore this must be broken as right now both test produce the same output but the TestIpfsAllGraph has more ipfs links in the input.

kishansagathiya: Ok. I see the problem. I will raise a PR against this.

PR #954, opened 2019-11-08
Fix graph not displaying all ipfs nodes when all-ipfs-peers enabled

hsanjuan:


hsanjuan:
It has been very long since I saw so many green things

PR #958, opened 2019-11-11
Instead of using a mock state, use dsstate

kishansagathiya:
- this covers only `pintracker_test.go` yet
- a similar exercise would be required in `stateless_test.go`


pintracker/pintracker_test.go
```diff
@@ -201,42 +157,24 @@ func TestPinTracker_Untrack(t *testing.T) {
...
 					Status: api.TrackerStatusPinned,
 				},
-				{
-					Cid:    test.Cid2,
-					Status: api.TrackerStatusRemote,
-				},
-				{
-					Cid:    test.Cid3,
-					Status: api.TrackerStatusPinned,
-				},
```
hsanjuan: ? unsure why removed

hsanjuan:
Ideally the only thing that I should see changing in all the pintracker tests is the initialization of things plus adjustments to tests expecting different outputs from List depending on whether the pin was fast or slow. The diff should be simpler than this.

kishansagathiya:
mockState is does not behave like a dsstate. It is returns value based on what tests need.

for example, in `slow`
`Pins` returns cid1 and cid3 with pinOpts (rmax and rmin -1), but
`PinGet(cid3)` would give `not found`  (contradicting behaviour that can't be done if using dsstate)

similarlu in normal mock rpc client,
`PinGet(cid2)` returns pin with rmax and rmin 1
`Pins` returns cid2 with rmax and rmin 0 

If you want to use a dsstate inside mockState, I am guessing, that would also require changing quite some tests like this. 

> initializing it with whatever content you need at the beginning of the tests rather than hardcoding it.

I mistook this as initializing content in the beginning of each test and not in the beginning of all tests. My bad. 

I will edit this and merge this with main PR.

hsanjuan:
> mockState is does not behave like a dsstate. It is returns value based on what tests need.
> 
> for example, in `slow`
> `Pins` returns cid1 and cid3 with pinOpts (rmax and rmin -1), but
> `PinGet(cid3)` would give `not found` (contradicting behaviour that can't be done if using dsstate)
> 

> similarlu in normal mock rpc client,
> `PinGet(cid2)` returns pin with rmax and rmin 1
> `Pins` returns cid2 with rmax and rmin 0
> 

I'm asking you to adjust tests where that special behaviour change is needed so that it is not needed anymore.

> If you want to use a dsstate inside mockState, I am guessing, that would also require changing quite some tests like this.
> 
> > initializing it with whatever content you need at the beginning of the tests rather than hardcoding it.
> 
> I mistook this as initializing content in the beginning of each test and not in the beginning of all tests. My bad.
> 
> I will edit this and merge this with main PR.

You might need to initialize differently on some tests, but since there was 1 mock, I would expect all test to work with a single pre-initialized state, once the things above are corrected.

hsanjuan:
Is this PR deprecated?

kishansagathiya:
> Is this PR deprecated?

Yes, changes added in https://github.com/ipfs/ipfs-cluster/pull/944

Issue #937, opened 2019-10-12
ipfs-cluster-service: print full working configuration at startup 


hsanjuan:
**Describe the feature you are proposing**

ipfs-cluster-service should print the working configuration on startup (as a debug log message). This could be the JSON form but:

* Sensitive information needs to be hidden
* What we call "hidden fields" (those with "omittempty" settings should appear). It might be better to print the actual Config objects

Therefore worth to write a stringer method for the actual Config structs. This can use struct tags to mark sensitive fields. A util function from `config` can be used for all ComponentConfig implentations.

**Additional context**

It should be explicit to know what configuration cluster is using after we have done all the parsing/replacing etc.

hsanjuan:
This is a `String()` method provided by the config manager that relies in `String()` methods in each ComponentConfig that use a helper method in the `config` module that prints the config structs, hiding the values for fields with a specific tag.

PR #959, opened 2019-11-13
print full working configuration at startup

kishansagathiya:
#937

### DIFF
--- a/api/ipfsproxy/config.go
+++ b/api/ipfsproxy/config.go
@@ -333,3 +333,13 @@ func (cfg *Config) toJSONConfig() (jcfg *jsonConfig, err error) {
 
 	return
 }
+
+// ToDisplayJSON returns JSON config as a string.
+func (cfg *Config) ToDisplayJSON() ([]byte, error) {
+	jcfg, err := cfg.toJSONConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	return config.DisplayJSON(jcfg)
+}
--- a/cluster_config.go
+++ b/cluster_config.go
@@ -162,8 +162,8 @@ type Config struct {
 type configJSON struct {
 	ID                   string             `json:"id,omitempty"`
 	Peername             string             `json:"peername"`
-	PrivateKey           string             `json:"private_key,omitempty"`
-	Secret               string             `json:"secret"`
+	PrivateKey           string             `json:"private_key,omitempty" hidden:"true"`
+	Secret               string             `json:"secret" hidden:"true"`
 	LeaveOnShutdown      bool               `json:"leave_on_shutdown"`
 	ListenMultiaddress   ipfsconfig.Strings `json:"listen_multiaddress"`
 	EnableRelayHop       bool               `json:"enable_relay_hop"`
@@ -524,6 +524,15 @@ func (cfg *Config) GetPeerstorePath() string {
 	}
 
 	return filepath.Join(cfg.BaseDir, filename)
+}
+
+// ToDisplayJSON returns JSON config as a string.
+func (cfg *Config) ToDisplayJSON() ([]byte, error) {
+	jcfg, err := cfg.toConfigJSON()
+	if err != nil {
+		return nil, err
+	}
+	return config.DisplayJSON(jcfg)
 }
 
 // DecodeClusterSecret parses a hex-encoded string, checks that it is exactly
--- a/cmd/ipfs-cluster-service/daemon.go
+++ b/cmd/ipfs-cluster-service/daemon.go
@@ -121,8 +121,11 @@ func createCluster(
 
 	cfgs := cfgHelper.Configs()
 	cfgMgr := cfgHelper.Manager()
+	cfgBytes, err := cfgMgr.ToDisplayJSON()
+	checkErr("getting configuration string", err)
+	logger.Debugf("Configuration:\n%s\n", cfgBytes)
 
-	ctx, err := tag.New(ctx, tag.Upsert(observations.HostKey, host.ID().Pretty()))
+	ctx, err = tag.New(ctx, tag.Upsert(observations.HostKey, host.ID().Pretty()))
 	checkErr("tag context with host id", err)
 
 	var apis []ipfscluster.API
--- a/config/config.go
+++ b/config/config.go
@@ -58,6 +58,8 @@ type ComponentConfig interface {
 	// Provides a channel to signal the Manager that the configuration
 	// should be persisted.
 	SaveCh() <-chan struct{}
+	// ToDisplayJSON returns a string representing the config excluding hidden fields.
+	ToDisplayJSON() ([]byte, error)
 }
 
 // These are the component configuration types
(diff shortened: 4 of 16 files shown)


kishansagathiya:
Not complete yet, because

- Yet to ignore hidden field from configs
- Yet to print it on the startup

I have used Config objects ( not jsonConfig objects), but because some of them have things like lock in them they get flagged in go-vet. and there is no way to ignore that go-vet suggestion (https://github.com/golang/go/issues/10752)

so probably using jsonConfigs would be better

kishansagathiya:
I will try with struct tags, see if there is a simpler. 
This feels like I am re-implementing json.Marshal.

cluster_config.go
```diff
@@ -519,6 +519,13 @@ func (cfg *Config) GetPeerstorePath() string {
 	return filepath.Join(cfg.BaseDir, filename)
 }
 
+func (cfg *Config) String() string {
+	jcfg, _ := cfg.toConfigJSON()
+
+	hidden := []string{"ID", "PrivateKey", "Secret"}
```
hsanjuan: ID is not secret

cmd/ipfs-cluster-service/daemon.go
```diff
@@ -122,6 +122,8 @@ func createCluster(
 	cfgs := cfgHelper.Configs()
 	cfgMgr := cfgHelper.Manager()
 
+	logger.Debugf("this cluster peer's configurations are as follows:\n%s", cfgMgr.String())
```
hsanjuan: `Configuration:`

config/config.go
```diff
@@ -547,6 +549,25 @@ func (cfg *Manager) IsLoadedFromJSON(t SectionType, name string) bool {
 	return !cfg.undefinedComps[t][name]
 }
 
+func (cfg *Manager) String() string {
+	var result string
```
hsanjuan: Use string.Builder

cluster_config.go
kishansagathiya: ID is not part of config. It has been moved to `identity.go`. ID is part of config only for compatibility,

api/ipfsproxy/config.go
```diff
@@ -317,7 +317,15 @@ func (cfg *Config) toJSONConfig() (jcfg *jsonConfig, err error) {
 	return
 }
 
-func (cfg *Config) String() string {
-	jcfg, _ := cfg.toJSONConfig()
-	return config.String(*jcfg, nil)
+func (cfg *Config) String() (string, error) {
```
hsanjuan: Either we keep the Stringer interface by not returning errors (maybe returning the error instead of the string) or we rename this to a different method (ToDisplayJSON).

api/ipfsproxy/config.go
```diff
@@ -317,7 +317,15 @@ func (cfg *Config) toJSONConfig() (jcfg *jsonConfig, err error) {
...
+	jcfg, err := cfg.toJSONConfig()
+	if err != nil {
+		return "", err
+	}
+
+	bytes, err := config.DefaultJSONMarshalWithoutHiddenFields(*jcfg)
+	if err != nil {
+		return "", err
+	}
```
hsanjuan: could `return string(bytes), err` directly.

api/ipfsproxy/config.go
```diff
@@ -317,7 +317,15 @@ func (cfg *Config) toJSONConfig() (jcfg *jsonConfig, err error) {
...
-func (cfg *Config) String() string {
-	jcfg, _ := cfg.toJSONConfig()
-	return config.String(*jcfg, nil)
+func (cfg *Config) String() (string, error) {
+	jcfg, err := cfg.toJSONConfig()
+	if err != nil {
+		return "", err
+	}
+
+	bytes, err := config.DefaultJSONMarshalWithoutHiddenFields(*jcfg)
```
hsanjuan: Rather than de-referencing everytime, it is best to fix the function to de-reference automatically when the value passed in is a pointer.

config/config.go
```diff
@@ -547,6 +550,30 @@ func (cfg *Manager) IsLoadedFromJSON(t SectionType, name string) bool {
...
+	result.WriteString(fmt.Sprintf("%-20s : %s\n\n", cfg.clusterConfig.ConfigKey(), clusterConfig))
+	for _, stype := range SectionTypes() {
+		for _, s := range cfg.sections[stype] {
+			cfgStr, err := s.ToDisplayJSON()
+			if err != nil {
+				return "", err
+			}
+			result.WriteString(fmt.Sprintf("%-20s : %s\n", s.ConfigKey(), cfgStr))
+		}
+		result.WriteString("\n")
```
hsanjuan: Rather than printing something by hand, you can build a `jsonConfig` using the DisplayJSONs provided by every component, and print the DefaultJSONMarshal of that. Similar to how `ToJSON`. In fact, the stuff in ToJSON  can probably be re-used by putting most into a parametrized function with an option to calls either component.ToJSON or component.ToDisplayJSON.

config/util.go
```diff
@@ -147,7 +149,21 @@ func DefaultJSONMarshalWithoutHiddenFields(cfg interface{}) ([]byte, error) {
...
+			Name:      f.Name,
+			PkgPath:   f.PkgPath,
+			Type:      f.Type,
+			Offset:    f.Offset,
+			Index:     f.Index,
+			Anonymous: f.Anonymous,
+		}
+		tagStr := strings.Replace(string(f.Tag), "omitempty", "", 1)
+		tagStr = strings.Replace(tagStr, ",\"", "\"", 1)
+		newF.Tag = reflect.StructTag(tagStr)
```
hsanjuan: Why not setting `f.Tag` directly?

config/util.go
```diff
@@ -147,7 +149,21 @@ func DefaultJSONMarshalWithoutHiddenFields(cfg interface{}) ([]byte, error) {
...
+		newF := reflect.StructField{
+			Name:      f.Name,
+			PkgPath:   f.PkgPath,
+			Type:      f.Type,
+			Offset:    f.Offset,
+			Index:     f.Index,
+			Anonymous: f.Anonymous,
+		}
+		tagStr := strings.Replace(string(f.Tag), "omitempty", "", 1)
+		tagStr = strings.Replace(tagStr, ",\"", "\"", 1)
```
hsanjuan: What if its `json:"something,omitempty,somethingelse"`. Does it all break to have a tag like `json:"name," or is the need for this Replace an assumption?

config/config.go
```diff
@@ -499,8 +498,12 @@ func (cfg *Manager) ToJSON() ([]byte, error) {
...
 	if cfg.clusterConfig != nil {
 		cfg.clusterConfig.SetBaseDir(dir)
-		raw, err := cfg.clusterConfig.ToJSON()
-
+		var raw []byte
+		if display {
+			raw, err = cfg.clusterConfig.ToDisplayJSON()
+		} else {
+			raw, err = cfg.clusterConfig.ToJSON()
+		}
```
hsanjuan: This does not work when Source is set. It will not show the config then.

config/config.go
```diff
@@ -477,7 +479,7 @@ func (cfg *Manager) SaveJSON(path string) error {
 
 // ToJSON provides a JSON representation of the configuration by
 // generating JSON for all componenents registered.
-func (cfg *Manager) ToJSON() ([]byte, error) {
+func (cfg *Manager) ToJSON(display bool) ([]byte, error) {
```
hsanjuan: I'm not happy about this. What is `display`? Anyone reading the code will find that ToJSON() is taking true, or false. So there's an extra step about figuring out what this true or false means.

Issue #938, opened 2019-10-12
ipfs-cluster-service: allow to select logging facilities and level


hsanjuan:
**Describe the feature you are proposing**

Rework --loglevel flag so that specific component logging can be selected. Or leave this flag as it is and provide new flags/ways of selecting logging components. Allow the user to inspect the existing logging facilities and the default log-level that they are using.

**Additional context**

Current approach is pretty much log all or nothing.

kishansagathiya:
> Allow the user to inspect the existing logging facilities and the default log-level that they are using.

`log.json` with identifier to logleve map in `~/.ipfs-cluster` will give user ability to customize logs for multiple cluster runs

To weed out unwanted for just current run a `--skip-log-fac` could be provided.

hsanjuan:
Thinking more of exposing `/health/logs` or the sort. Not going to handle another configuration file for this.

There are two parts:

* Letting ipfs-cluster-service set the right log levels and facilities on boot (flags can also be mapped to env vars)
* Allowing to inspect/change at runtime.

Can be done in separate PRs.

PR #960, opened 2019-11-14
Environment variable to set loglevels for individual identifiers

kishansagathiya:
- Example: export LOG_FACS="cluster:debug,raft:debug"

`kelseyhightower/envconfig` works only with structs. I will look at this again and see if there a way to do this like `export LOG_CLUSTER=debug`

#938

### DIFF
--- a/cmd/ipfs-cluster-service/main.go
+++ b/cmd/ipfs-cluster-service/main.go
@@ -21,7 +21,6 @@ import (
 
 	semver "github.com/blang/semver"
 	logging "github.com/ipfs/go-log"
-	"github.com/kelseyhightower/envconfig"
 	cli "github.com/urfave/cli"
 )
 
@@ -191,9 +190,15 @@ func main() {
 			Usage: "enable full debug logging (very verbose)",
 		},
 		cli.StringFlag{
-			Name:  "loglevel, l",
-			Value: defaultLogLevel,
-			Usage: "set the loglevel for cluster components only [critical, error, warning, info, debug]",
+			Name:   "loglevel, l",
+			Value:  defaultLogLevel,
+			EnvVar: "LOG_LEVEL",
+			Usage:  "set the loglevel for cluster components only [critical, error, warning, info, debug]",
+		},
+		cli.StringSliceFlag{
+			Name:   "component-loglevel, cl",
+			EnvVar: "COMP_LOG_LEVEL",
+			Usage:  "set separate loglevel for individual cluster components",
 		},
 	}
 
@@ -206,13 +211,8 @@ func main() {
 		configPath = filepath.Join(absPath, DefaultConfigFile)
 		identityPath = filepath.Join(absPath, DefaultIdentityFile)
 
-		err = setupLogLevel(c.String("loglevel"))
-		if err != nil {
-			return err
-		}
-		if c.Bool("debug") {
-			setupDebug()
-		}
+		err = setupLogLevel(c.Bool("debug"), c.String("loglevel"), c.StringSlice("component-loglevel"))
+		checkErr("setting log levels", err)
 
 		locker = &lock{path: absPath}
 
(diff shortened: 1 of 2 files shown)


hsanjuan:
Add or fix the flags that control the logging. Flags can be aliased by env vars as supported by the urfave/cli library, but the main point is that you propose flags for ipfs-cluster-service that are versatile enough for this.

For example, having something like `--loglevel error --loglevel swarm2,debug --loglevel crdt,info` working.

kishansagathiya:
environment variable should have loglevels comma separated, like this
```
export COMP_LOG_LEVEL=crdt:debug,cluster:debug
```

cmd/ipfs-cluster-service/main.go
```diff
@@ -190,9 +190,15 @@ func main() {
 			Usage: "enable full debug logging (very verbose)",
 		},
 		cli.StringFlag{
-			Name:  "loglevel, l",
-			Value: defaultLogLevel,
-			Usage: "set the loglevel for cluster components only [critical, error, warning, info, debug]",
+			Name:   "loglevel, l",
+			Value:  defaultLogLevel,
+			EnvVar: "LOG_LEVEL",
```
hsanjuan: CLUSTER_LOG_LEVEL

cmd/ipfs-cluster-service/main.go
```diff
@@ -190,9 +190,15 @@ func main() {
...
+			Usage:  "set the loglevel for cluster components only [critical, error, warning, info, debug]",
+		},
+		cli.StringSliceFlag{
+			Name:   "component-loglevel, cl",
+			EnvVar: "COMP_LOG_LEVEL",
+			Usage:  "set separate loglevel for individual cluster components",
```
hsanjuan: We can use a single flag for everything right?

"debug"  means all cluster
"debug,swarm2:info" means all facilities to debug and swarm2 to info
"raft:error", means all facilities with default logging except raft.

cmd/ipfs-cluster-service/main.go
kishansagathiya: `LOG_LEVEL` matches the flag name.
 One could get confused thinking `CLUSTER_LOG_LEVEL` affects only cluster component.

kishansagathiya: `LOG_LEVEL` matches the flag name.
 One could get confused thinking `CLUSTER_LOG_LEVEL` affects only cluster component.

cmd/ipfs-cluster-service/main.go
kishansagathiya: I have pushed changes to do everything together.

But I find the previous way better. Now if someone uses "swarms=info" or "swam2-info" it would say that it is not a valid log level, which is less informative than saying `use the formate identifier:loglevel`

cmd/ipfs-cluster-service/main.go
```diff
@@ -189,10 +189,10 @@ func main() {
 			Name:  "debug, d",
 			Usage: "enable full debug logging (very verbose)",
 		},
-		cli.StringFlag{
-			Name:  "loglevel, l",
-			Value: defaultLogLevel,
-			Usage: "set the loglevel for cluster components only [critical, error, warning, info, debug]",
+		cli.StringSliceFlag{
+			Name:   "loglevel, l",
+			EnvVar: "LOG_LEVEL",
```
hsanjuan: IPFS_CLUSTER_LOG_LEVEL

Cluster environment variables should have CLUSTER_ in them. `env | grep CLUSTER_` should show everything affecting cluster.

cmd/ipfs-cluster-service/main.go
```diff
@@ -189,10 +189,10 @@ func main() {
...
+			Usage:  "set loglevels for all or separate loglevels for individual log identifiers [loglevel, identifier:loglevel]. Valid loglevels are critical, error, warning, notice, info and debug.",
```
hsanjuan: Add a paragraph and/or an example to the description explaining how to set log levels, keep this line short.

cmd/ipfs-cluster-service/main.go
```diff
@@ -580,15 +579,69 @@ func run(c *cli.Context) error {
...
+	// get overall log level and component-wise log levels from arguments
+	for _, cll := range compLogLevel {
+		if cll == "" {
+			continue
+		}
+		identifierToLevel := strings.Split(cll, ":")
+
+		switch len(identifierToLevel) {
+		case 1:
+			lvl := strings.ToUpper(identifierToLevel[0])
```
hsanjuan: I think it does not matter if the log level is uppercased.

cmd/ipfs-cluster-service/main.go
```diff
@@ -189,10 +189,10 @@ func main() {
 			Name:  "debug, d",
 			Usage: "enable full debug logging (very verbose)",
 		},
-		cli.StringFlag{
-			Name:  "loglevel, l",
-			Value: defaultLogLevel,
-			Usage: "set the loglevel for cluster components only [critical, error, warning, info, debug]",
+		cli.StringSliceFlag{
```
hsanjuan: Let's make this a StringFlag and parse the `--loglevel` value in the same way the environment value is parsed. A format which needs to be explained.

cmd/ipfs-cluster-service/main.go
```diff
@@ -580,15 +579,69 @@ func run(c *cli.Context) error {
...
-	for f := range ipfscluster.LoggingFacilities {
-		ipfscluster.SetFacilityLogLevel(f, lvl)
+func setupLogLevel(debug bool, compLogLevel []string) error {
+	// if debug is set to true, log everything in debug level
+	if debug {
+		ipfscluster.SetFacilityLogLevel("*", "DEBUG")
+		return nil
+	}
+
+	var logLevel string
```
hsanjuan: This is overcomplicated from this point.

0. Make the compLogFacs maps and set "all" -> defaultLogLevel
1. Loop and parse the logLevel flags into additional entries in the map or overwrite "all" when not component-specific.
2. for f := range ipfscluster.LoggingFacilities ... -> set to compLogFacs["all"]
3. Loop compLogFacs and set the rest of things.

The fact that 2 of your case entries are nearly identical should raise the flag that a refactor is needed.

logging.go
```diff
@@ -66,3 +66,14 @@ func SetFacilityLogLevel(f, l string) {
...
+// IsLogLevel checks if given string is a valid log level.
+func IsLogLevel(s string) bool {
+	for _, level := range []string{"DEBUG", "CRITICAL", "ERROR", "WARNING", "NOTICE", "INFO"} {
+		if s == level {
+			return true
+		}
+	}
+
+	return false
+}
```
hsanjuan: What a valid log level is is defined by the underlying library. We should not define that. It already provides methods to check whether a level exists: https://godoc.org/github.com/ipfs/go-log#LevelFromString

kishansagathiya: This requires to update go-log. Studying go-log and zap to make sure formatting stays the same.

kishansagathiya: > Studying go-log and zap to make sure formatting stays the same.

I could not find a way to keep the formatting same. It seems that new version(version in which LevelFromString exist) provide only two formatting option colorful or plain. What we use right now, logging.LogFormats, is removed.
New version uses zap internally. I think using zap.Config and zapcore.EncodeConfig it would be possible to achieve the right formatting, but it is not possible to provide that using go-log.

hsanjuan: As discussed, we do not need to verify that the log level is valid since the library does that and errors accordingly: https://github.com/ipfs/go-log/blob/v0.0.1/oldlog.go#L129

PR #961, opened 2019-11-16
`repo gc` in ipfs proxy

kishansagathiya:


api/ipfsproxy/ipfsproxy.go
```diff
@@ -647,6 +651,50 @@ func (proxy *Server) repoStatHandler(w http.ResponseWriter, r *http.Request) {
...
 
+type ipfsRepoGCResp struct {
+	Key   cid.Cid `json:",omitempty"`
+	Error string  `json:",omitempty"`
+}
+
+func (proxy *Server) repoGCHandler(w http.ResponseWriter, r *http.Request) {
+	queryValues := r.URL.Query()
+	streamErrors := queryValues.Get("stream-errors")
+	// TODO: Does `quiet` really quiet anything?
```
hsanjuan: Yes, but only for text output https://github.com/ipfs/go-ipfs/blob/master/core/commands/repo.go#L121

hsanjuan: So we can ignore

api/ipfsproxy/ipfsproxy.go
```diff
@@ -647,6 +651,50 @@ func (proxy *Server) repoStatHandler(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		ipfsErrorResponder(w, err.Error(), -1)
+	}
+
+	w.WriteHeader(http.StatusOK)
+	enc := json.NewEncoder(w)
+	var ipfsRepoGC ipfsRepoGCResp
+	for _, gc := range repoGC.PeerMap {
+		for _, key := range gc.Keys {
+			if streamErrors == "true" {
```
hsanjuan: declare streamErrors as a boolean (` := queryValues.Get("stream-errors") == "true"`)

api/ipfsproxy/ipfsproxy.go
```diff
@@ -647,6 +651,50 @@ func (proxy *Server) repoStatHandler(w http.ResponseWriter, r *http.Request) {
...
+				ipfsRepoGC = ipfsRepoGCResp{Key: key.Key}
+			}
+
+			// Cluster tags start with small letter, but IPFS tags with capital letter.
+			if err := enc.Encode(ipfsRepoGC); err != nil {
+				logger.Error(err)
+			}
+		}
+	}
+	return
```
hsanjuan: If stream errors is false and there were errors, these need to be collected and the X-Stream-Error Trailer needs to be set to their value (ipfs has a multierror type that is returned here (https://github.com/ipfs/go-ipfs/blob/master/core/commands/repo.go#L97). Similar to /add. The logic of how ipfs does this last part is around go-ipfs-cmds/http/responseEmitter:closeWithError().

hsanjuan: Worth testing.

kishansagathiya:
Looking at https://github.com/ipfs/go-ipfs-cmds/blob/master/http/responseemitter.go
all responses would have these headers
```
< Access-Control-Allow-Headers: X-Stream-Output, X-Chunked-Output, X-Content-Length
< Access-Control-Expose-Headers: X-Stream-Output, X-Chunked-Output, X-Content-Length
< Trailer: X-Stream-Error
```

worth adding these in proxy responses as well?

api/ipfsproxy/ipfsproxy.go
```diff
@@ -658,10 +660,12 @@ type ipfsRepoGCResp struct {
...
 func (proxy *Server) repoGCHandler(w http.ResponseWriter, r *http.Request) {
 	queryValues := r.URL.Query()
-	streamErrors := queryValues.Get("stream-errors")
-	// TODO: Does `quiet` really quiet anything?
+	streamErrors := queryValues.Get("stream-errors") == "true"
+	// ignoring `quiet` since it only affects text output
 
 	proxy.setHeaders(w.Header(), r)
+
+	w.Header().Set("Trailer", "X-Stream-Error")
```
hsanjuan: We will need to add this to the `add` handler, I think. It's a bug (new PR).

api/ipfsproxy/ipfsproxy.go
```diff
@@ -692,6 +701,11 @@ func (proxy *Server) repoGCHandler(w http.ResponseWriter, r *http.Request) {
 			}
 		}
 	}
+
+	if !streamErrors && len(errs) != 0 {
+		w.Header().Set("X-Stream-Error", ipfsrepo.NewMultiError(errs...).Error())
```
hsanjuan: We should not require all go-ipfs module to re-use a tiny method. Cluster should not depend on go-ipfs.

api/ipfsproxy/ipfsproxy.go
hsanjuan: Worth adding to all requests.

kishansagathiya:
Seems like you can't check names of Trailers in w.Header() like every other headers.

api/ipfsproxy/ipfsproxy.go
```diff
@@ -703,7 +702,7 @@ func (proxy *Server) repoGCHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if !streamErrors && len(errs) != 0 {
-		w.Header().Set("X-Stream-Error", ipfsrepo.NewMultiError(errs...).Error())
+		w.Header().Set("X-Stream-Error", newMultiError(errs...).Error())
```
hsanjuan: Instead of building a slice of errors, which you then loop to generate a string, build the final error string directly as you receive the errors. Note how the last error is also becoming the "summary" even though it is not a summary.

kishansagathiya:
> I'm not sure, is this a problem here?

Nope, It is not a problem. You can check if trailer header was set by checking response.Trailer

hsanjuan:
@kishansagathiya can you send a PR for https://cluster.ipfs.io/documentation/reference/proxy/

PR #965, opened 2019-11-27
Use index endpoint with a trailing slash

kishansagathiya:
On using `/debug/pprof` as index endpoint, profiles contained by it
would point to wrong url. For example, link for `allocs` would be
`/debug/allocs` (results in 404), whereas it should actually be
`/debug/pprof/allocs`.

Use a trailing slash to get correct links. `/debug/pprof` will
automatically be redirected to `/debug/pprof/`


hsanjuan:
Thanks

PR #967, opened 2019-12-02
config manager: Improve source-fetch error handling

hsanjuan:
* Allow detection of source-download errors by adding IsSourceDownloadError method.
* Treat non-success response codes as a failure to fetch the source.
* Set the Source url before errors happens so it can be consulted even in that case.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -21,7 +21,9 @@ var logger = logging.Logger("config")
 
 var (
 	// Error when downloading a Source-based configuration
-	errFetchingSource = errors.New("could not fetch configuration source")
+	errFetchingSource = errors.New("could not fetch configuration from source")
+	// Error when remote source points to another remote-source
+	errSourceRedirect = errors.New("a sourced configuration cannot point to another source")
 )
 
 // IsErrFetchingSource reports whether this error happened when trying to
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -137,6 +149,9 @@ func NewManager() *Manager {
 // Shutdown makes sure all configuration save operations are finished
 // before returning.
 func (cfg *Manager) Shutdown() {
+	if cfg == nil {
```
kishansagathiya: I skimmed through some of the code. I did not find where a nil cfg could call Shutdown.

If that happens, if would help to not silence this error.

You could also see if you should make this method with value receiver so that Shutdown doesn't get called at all.

config/config.go
```diff
@@ -346,16 +361,22 @@ func (cfg *Manager) LoadJSONFromFile(path string) error {
 // LoadJSONFromHTTPSource reads a Configuration file from a URL and parses it.
```
kishansagathiya: line 357-8
return cfg.LoadJSON(file)

config/config.go
```diff
@@ -346,16 +361,22 @@ func (cfg *Manager) LoadJSONFromFile(path string) error {
...
 func (cfg *Manager) LoadJSONFromHTTPSource(url string) error {
 	logger.Infof("loading configuration from %s", url)
+	cfg.Source = url
 	resp, err := http.Get(url)
 	if err != nil {
-		return err
+		logger.Error(err)
```
kishansagathiya: this will get logged twice. Once here and once in service/main.go

config/config.go
hsanjuan: True, this is not necessary.

This method could not be made a value receiver though, as this would copy the WaitGroup.

PR #968, opened 2019-12-02
Fix: disable default libp2p API endpoint with CRDT consensus

hsanjuan:


cmd/ipfs-cluster-service/daemon.go
```diff
@@ -127,9 +127,19 @@ func createCluster(
 
 	var apis []ipfscluster.API
 	if cfgMgr.IsLoadedFromJSON(config.API, cfgs.Restapi.ConfigKey()) {
-		rest, err := rest.NewAPIWithHost(ctx, cfgs.Restapi, host)
+		var api *rest.API
+		// Do NOT enable default Libp2p API endpoint on CRDT
+		// clusters. Collaborative clusters are likely to share the
+		// secret with untrusted peers, thus the API would be open for
+		// anyone.
```
kishansagathiya: Do we expose ourselves to any risk by opening this API to anyone?

hsanjuan: I don't fully understand your question.. I guess yes? That's the point of this PR...

PR #970, opened 2019-12-02
Service: extract some methods to cmdutils

hsanjuan:
And some small renaming.

### DIFF
--- a/cmd/ipfs-cluster-service/main.go
+++ b/cmd/ipfs-cluster-service/main.go
@@ -25,7 +25,7 @@ import (
 )
 
 // ProgramName of this application
-const programName = `ipfs-cluster-service`
+const programName = "ipfs-cluster-service"
 
 // flag defaults
 const (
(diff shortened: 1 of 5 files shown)


cmd/ipfs-cluster-service/main.go
```diff
@@ -628,23 +628,14 @@ func yesNoPrompt(prompt string) bool {
...
-		cfgHelper.Configs(),
+	cfgHelper, err := cmdutils.NewLoadedConfigHelper(
+		configPath,
+		identityPath,
 	)
-	checkErr("creating state manager", err)
+	checkErr("loading configurations", err)
+	cfgHelper.Manager().Shutdown()
+	mgr, err := cmdutils.NewStateManagerWithHelper(cfgHelper)
+	checkErr("loading state manager", err)
```
kishansagathiya:
```suggestion
	checkErr("creating state manager", err)
```

cmdutils/cmdutils.go
```diff
@@ -56,3 +63,67 @@ func getPort(ln net.Listener, code int) int {
 	}
 	return 0
 }
+
+// HandleSignals orderly shutsdown an IPFS Cluster peer
```
kishansagathiya:
```suggestion
// HandleSignals orderly shuts down an IPFS Cluster peer
```

cmdutils/state.go
```diff
@@ -26,6 +26,7 @@ type StateManager interface {
 	ImportState(io.Reader) error
 	ExportState(io.Writer) error
 	GetStore() (ds.Datastore, error)
+	GetOfflineState(ds.Datastore) (state.State, error)
```
kishansagathiya: This method is not used anywhere outside this package. Why export it?

hsanjuan: It will be used.

PR #972, opened 2019-12-04
Add metrics based on types of tracker status

### LINKED ISSUE #897: Add metric to expose number of cids pinning

kishansagathiya:
- Also fixes locks related bug in operations

(TODO: This doesn't produce the right count yet somehow)

Fixes #897


pintracker/optracker/operationtracker.go
```diff
@@ -209,18 +204,17 @@ func (opt *OperationTracker) GetAll(ctx context.Context) []*api.PinInfo {
...
+	defer opt.mu.Unlock()
 
-	if errop.Phase() != PhaseError {
+	errOp, ok := opt.operations[cidStr]
+	if !ok || errOp.Phase() != PhaseError {
 		return
 	}
 
-	opt.Clean(ctx, errop)
+	delete(opt.operations, cidStr)
```
hsanjuan: As far as I can see, this is changing behaviour and introducing a bug.

hsanjuan: There seems to be a double-locking issue, but it should only remove the operations when the pointer is the same.

kishansagathiya: I changed this because of the double locking issue. How is this introducing a bug?

hsanjuan: You're right, sorry, I misread. In the pintracker PR I fully removed this method as it is not called from anywhere.

pintracker/optracker/operation.go
```diff
@@ -160,16 +172,32 @@ func (op *Operation) Error() string {
...
 		op.ts = time.Now()
 	}
 	op.mu.Unlock()
+
+	op.recordStatuses(prevStatus, op.ToTrackerStatus())
 	span.End()
 }
 
+func (op *Operation) recordStatuses(prevStatus api.TrackerStatus, newStatus api.TrackerStatus) {
+	if !prevStatus.Match(newStatus) {
```
hsanjuan: Isn't it better to do `prevStatus == newStatus` ? Using match implies we expect a status may match several other statuses, but here we are just checking for equality right?

pintracker/optracker/operation.go
```diff
@@ -160,16 +172,32 @@ func (op *Operation) Error() string {
...
+		stats.Record(op.ctx, observations.GetMeasureFromStatus(prevStatus).M(-1))
+		stats.Record(op.ctx, observations.GetMeasureFromStatus(newStatus).M(1))
+	}
+
+	if newStatus.Match(api.TrackerStatusUnpinned) {
+		stats.Record(op.ctx, observations.GetMeasureFromStatus(api.TrackerStatusPinned).M(-1))
+	}
+}
```
hsanjuan: If prevStatus is pinned. And new status is unpinned. Won't this will do M(-1) twice on the Pinned metric? Also, there is no Unpinned view anyway... the code seems a bit shaky here. Why not having a view for unpinned items?

Also, doing -1 on the old metrics and +1 in the new metrics also works if both are the same, so we can do that on every case and skip checking if they are the same or not.

observations/metrics.go
```diff
@@ -66,11 +111,88 @@ var (
...
+	}
+
 	DefaultViews = []*view.View{
 		PinsView,
 		TrackerPinsView,
 		PeersView,
 		AlertsView,
+		&StatusClusterErrorView,
+		&StatusPinErrorView,
+		&StatusUnpinErrorView,
```
hsanjuan: For consistency, do the &reference in the previous views as well.

pintracker/optracker/operation.go
kishansagathiya: There are status change paths based on Operations. One is UnpinQueued to Unpinned, other is PinQueued to Pinned. Pinned and Unpinned marking end of an operation. So, even though a pinned item can be unpinned or vice-versa, phase changes will not happen from Pinned to Unpinned, because they are separate operation.

kishansagathiya: > Why not having a view for unpinned items

Anything that is not in the shared state is unpinned. That way number of unpinned items are infinite.
We can have more lenient definition (only for the purpose of metrics), where we consider only those unpinned things that were pinned before. There are no states after unpinned. So, unpinned will always increase. If same item is pinned and unpinned multiple times, that would increase unpinned count.
Moreover, unlike other statuses, I fail to see value of measuring unpinned_count

If this convinces you I need to remove int64Measure meant to track this status.

kishansagathiya: > Also, doing -1 on the old metrics and +1 in the new metrics also works if both are the same, so we can do that on every case and skip checking if they are the same or not.

Checking this condition reduce unnecessary function calls

PR #976, opened 2019-12-07
Feat: ipfs-cluster-follow

hsanjuan:
This adds a new cluster command: ipfs-cluster-follow.

This command allows initializing and running follower peers as configured by a
remote-source configuration. The command can list configured peers
and obtain information for each of them.

Peers are launched with the rest API listening on a local unix socket. The
command can be run to list the items in the cluster pinset using this
endpoint. Alternatively, if no socket is present, the peer will be assumed to
be offline and the pin list will be directly read from the datastore.

Cluster peers launched with this command (and their configurations) are
compatible with ipfs-cluster-ctl and ipfs-cluster-service. We purposely do not
support most configuration options here. Using ipfs-cluster-ctl or launching
the peers using ipfs-cluster-service is always an option when the usecase
deviates from that supported by ipfs-cluster-follow.

Examples:

$ ipfs-cluster-follow -> list configured peers
$ ipfs-cluster-follow --help
$ ipfs-cluster-follow \<clusterName\> init <url>
$ ipfs-cluster-follow \<clusterName\> info
$ ipfs-cluster-follow \<clusterName\> run
$ ipfs-cluster-follow \<clusterName\> list


cmd/ipfs-cluster-follow/main.go
```diff
@@ -0,0 +1,286 @@
...
+			},
+		}
+		return clusterApp.RunAsSubcommand(c)
+	}
+
+	app.Run(os.Args)
+}
+
+// build paths returns the path to the configuration folder,
+// the identify and the service.json files.
```
kishansagathiya:
```suggestion
// the identity.json and the service.json files.
```

cmd/ipfs-cluster-follow/main.go
```diff
@@ -0,0 +1,286 @@
...
+				Action: infoCmd,
+			},
+			{
+				Name:      "init",
+				Usage:     "initializes the follower peer",
+				ArgsUsage: "<template_URL>",
+				Description: fmt.Sprintf(`
+This command initializes a follower peer for the cluster named "%s". You
+will need to pass the peer configuration URL. The command will generate a new
+peer identity and leave things readyto run "%s %s run".
```
kishansagathiya:
```suggestion
peer identity and leave things ready to run "%s %s run".
```

cmd/ipfs-cluster-follow/commands.go
```diff
@@ -0,0 +1,434 @@
...
+
+	for f := range ipfscluster.LoggingFacilities {
+		ipfscluster.SetFacilityLogLevel(f, "critical")
+	}
+
+	absPath, configPath, identityPath := buildPaths(c, clusterName)
+	cfgHelper, err := cmdutils.NewLoadedConfigHelper(configPath, identityPath)
+	var url string
+	if err != nil {
+		if config.IsErrFetchingSource(err) {
```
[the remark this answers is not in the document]
hsanjuan: I prefer this. The error is a wrapped one.

cmd/ipfs-cluster-follow/commands.go
```diff
@@ -0,0 +1,434 @@
...
+		alloc,
+		[]ipfscluster.Informer{informer},
+		tracer,
+	)
+	if err != nil {
+		store.Close()
+		return cli.Exit(errors.Wrap(err, "error creating cluster peer"), 1)
+	}
+
+	return cmdutils.HandleSignals(ctx, cancel, cluster, host, dht)
```
hsanjuan: Note to self, Shutdown() closes the store. Document this.

cmd/ipfs-cluster-follow/commands.go
```diff
@@ -0,0 +1,434 @@
...
+
+	absPath, configPath, identityPath := buildPaths(c, clusterName)
+	cfgHelper, err := cmdutils.NewLoadedConfigHelper(configPath, identityPath)
+	if err != nil {
+		return cli.Exit(errors.Wrap(err, "error loading configurations"), 1)
+	}
+	cfgHelper.Manager().Shutdown()
+
+	err = printStatusOnline(absPath, clusterName)
+	if err != nil {
```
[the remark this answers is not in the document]
hsanjuan: It tries online and then offline. It is not an error if node is offline. Can try to make sure it is actually that error and not something else though.

cmd/ipfs-cluster-follow/main.go
```diff
@@ -0,0 +1,286 @@
...
+	programName,
+	programName,
+)
+
+var logger = logging.Logger("clusterfollow")
+
+// Default location for the configurations and data
+var (
+	// DefaultFolder is the name of the cluster folder
+	DefaultFolder = ".ipfs-cluster-follow"
```
hsanjuan: Would be nice to have someone test how this plays on windows for a default folder.

cmd/ipfs-cluster-follow/main.go
```diff
@@ -0,0 +1,286 @@
...
+
+func getClient(absPath, clusterName string) (client.Client, error) {
+	endp, err := socketAddress(absPath, clusterName)
+	if err != nil {
+		return nil, err
+	}
+	cfg := client.Config{
+		APIAddr: endp,
+	}
+	return client.NewDefaultClient(&cfg)
```
hsanjuan: will need adapting if we support the api running on a tcp port..

cmd/ipfs-cluster-follow/commands.go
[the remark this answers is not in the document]
hsanjuan: fixed

hsanjuan:
>     * Need to not timeout overly early when fetching the configuration

Any timeouts would come from the gateway, we Get without timeout.

cmd/ipfs-cluster-follow/commands.go
```diff
@@ -0,0 +1,486 @@
...
+	if len(filteredDirs) == 0 {
+		printFirstStart()
+		return nil
+	}
+
+	fmt.Printf("Configurations found for %d follower peers. For info and help, try running:\n\n", len(filteredDirs))
+	for _, d := range filteredDirs {
+		fmt.Printf("%s \"%s\"\n", programName, d)
+	}
+	fmt.Printf("\nTip: \"%s --help\" for help and examples.", programName)
```
kishansagathiya:
```suggestion
	fmt.Printf("\nTip: \"%s --help\" for help and examples.\n", programName)
```

cmd/ipfs-cluster-follow/commands.go
```diff
@@ -0,0 +1,486 @@
...
+	crdtcons, err := crdt.New(
+		host,
+		dht,
+		pubsub,
+		cfgs.Crdt,
+		store,
+	)
+	if err != nil {
+		store.Close()
+		return cli.Exit(errors.Wrap(err, "creating CRDT component"), 1)
```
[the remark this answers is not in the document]
hsanjuan: There is no persistent state to worry about with `host` and `dht`. Also the host has no connections and the context used to create them is cancelled on return. So no real need to explicitly Close().

hsanjuan:
> Users different peers with same IPFS node, it is possible in both `ipfs-cluster-service` and `ipfs-cluster-follow`. Data could be allocated to two different cluster peers, satisfying a replication factor of 2 when in reality there is only one replication because it's the same ipfs node.
> 
> Malicious users could mislead the network using this.

Cluster does not protect against peers faking  (pinning something that they are not pinning). Recommendation is to use replication factor -1 in public clusters. Otherwise, Filecoin does protect against peers faking what they are storing.

PR #977, opened 2019-12-12
[WIP] Introduce epoch in metric so that when a peer is restarted new metrics

kishansagathiya:
series can be used and old one can be discarded


kishansagathiya:
If number of metrics are not enough, I am relying on expiry. I am facing some bug where metrics are often expired. Investigating.

PR #979, opened 2019-12-13
 Fix: stateless: cluster should pin items that are in the state but not on ipfs

hsanjuan:
StateSync() used to take care of this by issuing Track() calls. But this
functionality was removed.

This starts returning items that are in the state but not on IPFS as
PIN_ERRORs. It ensures that the Recover methods see them so that they can
trigger repinnings for missing items. This covers cases where the user
modifies the ipfs state manually, or resets the ipfs daemon but keeps the
cluster state, and cases where cluster was stopped half-way through a pinning.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -614,10 +614,10 @@ This might be due to one or several causes:
 		c.Shutdown(ctx)
 		return
 	case <-c.consensus.Ready(ctx):
-		// Consensus ready means the state is up to date so we can sync
-		// it to the tracker. We ignore errors (normal when state
-		// doesn't exist in new peers).
-		c.StateSync(ctx)
+		// Consensus ready means the state is up to date. Every item
+		// in the state that is not pinned will appear as PinError so
+		// we can proceed to recover all of those in the tracker.
+		c.RecoverAllLocal(ctx)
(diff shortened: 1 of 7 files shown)


hsanjuan:
@kishansagathiya I merged #944 too fast. The removal of functionality in StateSync was not replaced anywhere else so cluster would never try to pin things that had not made it to IPFS or had disappeared.

I took the opportunity to clean up the tests a bit too.

Please review..

hsanjuan:
> LGTM.
> 
> Though I have one concern.
> 
> RecoverAll calls StatusAll.
> With stateless, quite some work load was moved from StateSync to StatusAll.
> 
> Now with, RecoverAll running every few minutes, we might run into the same problem again (cpu spike every few minutes when RecoverAll gets called)

Can't do much there fore now... if we are to check that the things we want pinned are pinned we need an ipfs `pin/ls`. Alternative would be to run individual `pin/ls/cid` commands, which for really big pinsets is even worse. Until we have streaming we will have to load that response on memory. At least there will be one less copy of the pinset in memory overall. If the user does not worry about auto-recover so much, the interval can be increased.

PR #980, opened 2019-12-13
Added Announce/NoAnnounce/FilterAddrs swarm options

### LINKED ISSUE #952: Announce/NoAnnounce/FilterAddrs swarm options

Same as ipfs options.

Maybe a config option:

```
"addresses": {
  "announce": [],
  "no_announce": [],
  "filter": []
}
```

kishansagathiya:
Same as corresponding IPFS Options
announce - the swarm addresses to announce to the network
no_announce - swarm addresses not to announce to the network
addrs_filters - addresses (multiaddr netmasks) to not dial

Fixes #952 

Code inspired/copied from https://github.com/ipfs/go-ipfs/blob/master/core/node/libp2p/addrs.go
I guess that should be fine because these options are same as IPFS and IPFS functions used to set this options cannot be reused nicely in cluster since those functions return `func() (opts Libp2pOpts, err error) ` instead of the options.

### DIFF
--- a/util.go
+++ b/util.go
@@ -166,3 +166,26 @@ func peersSubtract(a []peer.ID, b []peer.ID) []peer.ID {
 
 	return result
 }
+
+func toMultiAddrs(addrs []string) ([]ma.Multiaddr, error) {
+	var mAddrs []ma.Multiaddr
+	for _, addr := range addrs {
+		mAddr, err := ma.NewMultiaddr(addr)
+		if err != nil {
+			//err = fmt.Errorf("error parsing a listen_multiaddress: %s", err)
+			return nil, err
+		}
+		mAddrs = append(mAddrs, mAddr)
+	}
+
+	return mAddrs, nil
+}
+
+func multiAddrstoStrings(mAddrs []ma.Multiaddr) []string {
+	var addrs []string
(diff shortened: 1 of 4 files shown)


hsanjuan:
See if you can get this extracted from go-ipfs:

* Since AddrsFactory is in the end a libp2p thing in host/basic, might be worth having a DefaultAddrFactory builder in there.
* Otherwise, might be a good candidate for go-ipfs-util
* -> Open issue on go-ipfs to gather feedback about whether it is possible to extract this and what would be the best place to do so from the relevant stakeholders.

util.go
```diff
@@ -166,3 +166,26 @@ func peersSubtract(a []peer.ID, b []peer.ID) []peer.ID {
 
 	return result
 }
+
+func toMultiAddrs(addrs []string) ([]ma.Multiaddr, error) {
+	var mAddrs []ma.Multiaddr
+	for _, addr := range addrs {
+		mAddr, err := ma.NewMultiaddr(addr)
+		if err != nil {
+			//err = fmt.Errorf("error parsing a listen_multiaddress: %s", err)
```
kishansagathiya:
```suggestion
			
```

PR #994, opened 2020-01-01
Introducing `add` for tar.gz files

kishansagathiya:
Given a tar.gz file for adding, add its contents to cluster

This is super-WIP
- It doesn't work (yet),  if tar contains files directly instead of wrapping them in a directory
- I have tried this with tars containing some quite nested and decently big (say 20 files excluding directories) file trees and was able to achieve the same cid as if you would add them directly. But it failed to match cids tars containing some relatively large file trees (say codebase of this project)


kishansagathiya:
TODO: Ignore dot files.

Issue #1000, opened 2020-02-02
default cluster.listen_multiaddress is legacy IP only


RubenKelevra:
**Additional information:**
 - OS: Linux
 - IPFS Cluster version: 0.12.1
 - Installation method: binary

**Describe the bug:**

The default configuration of ipfs-cluster-service only creates IPv4 entries for the cluster connections on port 9096 (+quic). The IPv6 entries have to be added manually.

Why?

hsanjuan:
It is the safest default. For this, I think it is reasonable that the user configures it the way it works for them... at least for the time being

RubenKelevra:
@hsanjuan what's unsafe about using IPv6? 🤔

In many countries, you only get IPv4 with very bad carrier-grade nat for private internet connections. But you get a public IPv6 which can let you establish reliable connections.

If you configure only IPv4 by default on the servers, you end up with a lot of use of relay hopping inside the cluster, or connection issues.

hsanjuan:
Well, the first question I have is whether it blows up whenever it starts on a device without ipv6 interfaces/config.

RubenKelevra:
I don't think there's a OS out there which doesn't have IPv6 enabled by default.

The worst case is, that you open 1-2 ports on ::1, depending if you have quic enabled or not, and never use them.

If you got an idea which OS have no IPv6 out of the box, let me know and I'll set up a VM and test it.

Also: Just for this one weird edge case I wouldn't disable IPv6 by default. We don't have new IPv4 to allocate in nearly all regions. Which will lead to more and more crude workarounds by providers.

hsanjuan:
Fair enough, mind sending a PR? cluster_config.go, api/restapi/config.go, api/ipfsproxy/config.go have default listen addresses. The last two need a small adaptation so we can parse both a string and an array from the json, this was already fixed in cluster_config.go -> using `ipfsconfig.Strings`.

Luflosi:
@RubenKelevra did #1010 resolve this issue? If so, it can be closed.

PR #1002, opened 2020-02-03
add ipv6 listening addresses to the default config

RubenKelevra:
Fixes https://github.com/ipfs/ipfs-cluster/issues/1000

@hsanjuan on ```Dockerfile-test``` and ```docker/start-daemons.sh``` I'm not sure how the syntax works. I really don't want to break things ;)

Since I don't know no go, take the change on ```cmd/ipfs-cluster-service/main_test.go``` also with a grain of salt.


hsanjuan:
Will work on #1010

PR #1010, opened 2020-02-20
add ipv6 listening addresses to the default config

hsanjuan:
Taking over from #1002 

@RubenKelevra

### DIFF
--- a/api/ipfsproxy/config.go
+++ b/api/ipfsproxy/config.go
@@ -137,7 +137,7 @@ func (cfg *Config) ConfigKey() string {
 
 // Default sets the fields of this Config to sensible default values.
 func (cfg *Config) Default() error {
-	var proxy []ma.Multiaddr
+	proxy := make([]ma.Multiaddr, 0, len(DefaultListenAddrs))
 	for _, def := range DefaultListenAddrs {
 		a, err := ma.NewMultiaddr(def)
 		if err != nil {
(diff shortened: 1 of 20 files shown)


hsanjuan:
@RubenKelevra travis ci, and docker don't have ipv6 enabled by default. I have disabled it again. I don't want peers failing to start on default configurations because the platform does not happen to be configured for ipv6.

RubenKelevra:
> @RubenKelevra travis ci, and docker don't have ipv6 enabled by default. I have disabled it again. I don't want peers failing to start on default configurations because the platform does not happen to be configured for ipv6.

Right, valid points.

How about removing the explicit default listening address configuration completely from the config file, like many daemons do (sshd for example).

If the user choose to explicitly configure an IPv6 or IPv4 listening address in the configuration, we can still fatally fail, since this is clearly a missing precondition.

If there's no configuration for listening addresses we just attach to ::1/127.0.0.1 for command and ::/0.0.0.0 for communication like originally proposed. If something of this fails, we can print a warning.

---

Apart from this it sounds pretty crazy that Docker and Travis CI don't support IPv6 by default, since this is pretty modern technology 🤔

hsanjuan:
I'll merge this as is. Enabling ipv6 with this as an addition to ipv4 is very easy and I don't see we should default to it or complicates things more right now, and now it would be just matter of updating the defaults which should be super easy.

PR #1016, opened 2020-03-06
Adder: fix #1006: Specify block format, hash and length when adding

### LINKED ISSUE #1006: add file with hash blake2b-256 does not work
**Additional information:**
 - OS: Linux
 - IPFS Cluster version: 0.12.1
 - Installation method: dist.ipfs.io

**Describe the bug:**

I'm running the command on the cluster:

`ipfs-cluster-ctl add --hash blake2b-256 --raw-leaves --quieter --name file --local --replication-min=1 --replication-max=10 /path/to/file`

I like to create a local directory entry with this file afterwards, so I use the CID (bafk2bzacecwvphs5yu7lpay66oksr7tx4ts6q6znldma5cbm53umuq76o6m3i) on the same machine to run an ipfs command:

`ipfs files cp /ipfs/bafk2bzacecwvphs5yu7lpay66oksr7tx4ts6q6znldma5cbm53umuq76o6m3i /cool/path/for/file`

This command does not return within 2-3 minutes since ipfs cannot find the hash locally...

`ipfs-cluster-ctl pin ls bafk2bzacecwvphs5yu7lpay66oksr7tx4ts6q6znldma5cbm53umuq76o6m3i`

Returns with a normal listing of the CID in the cluster.

`ipfs pin ls bafk2bzacecwvphs5yu7lpay66oksr7tx4ts6q6znldma5cbm53umuq76o6m3i`

Returns with an error:

`Error: path 'bafk2bzacecwvphs5yu7lpay66oksr7tx4ts6q6znldma5cbm53umuq76o6m3i' is not pinned`

---

The same file, commands, and cluster work fine, without specifying the "--hash" in the command line.

hsanjuan:
Currently we were only specifying the block format. When adding with
a custom hash function, even though we produced the right cids, IPFS
did not know the hash function and ended up storing them using SHA256.

Additionally, since NodeWithMeta serializes the CID, we do not need
to carry a Format parameter (which specifies the Codec): it is already
embedded.

Tests have been added and BlockPut in ipfshttp now checks that the
response's CID matches the data sent. This will catch errors like
what was happening, but also any data corruption between cluster and
IPFS during the block upload.

### DIFF
--- a/adder/util.go
+++ b/adder/util.go
@@ -93,20 +93,10 @@ func ipldNodeToNodeWithMeta(n ipld.Node) *api.NodeWithMeta {
 		logger.Warning(err)
 	}
 
-	format, ok := cid.CodecToStr[n.Cid().Type()]
-	if !ok {
-		format = ""
-		logger.Warning("unsupported cid type, treating as v0")
-	}
-	if n.Cid().Prefix().Version == 0 {
-		format = "v0"
-	}
-
 	return &api.NodeWithMeta{
 		Cid:     n.Cid(),
 		Data:    n.RawData(),
 		CumSize: size,
-		Format:  format,
 	}
 }
 
--- a/api/types.go
+++ b/api/types.go
@@ -883,7 +883,6 @@ type NodeWithMeta struct {
 	Data    []byte  `codec:"d,omitempty"`
 	Cid     cid.Cid `codec:"c,omitempty"`
 	CumSize uint64  `codec:"s,omitempty"` // Cumulative size
-	Format  string  `codec:"f,omitempty"`
 }
 
 // Size returns how big is the block. It is different from CumSize, which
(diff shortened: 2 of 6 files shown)


adder/util.go
```diff
@@ -93,20 +93,10 @@ func ipldNodeToNodeWithMeta(n ipld.Node) *api.NodeWithMeta {
...
 
-	format, ok := cid.CodecToStr[n.Cid().Type()]
-	if !ok {
-		format = ""
-		logger.Warning("unsupported cid type, treating as v0")
-	}
-	if n.Cid().Prefix().Version == 0 {
-		format = "v0"
-	}
-
```
hsanjuan: This is removed because all the information is part of the CID that we are adding.

PR #1023, opened 2020-03-28
Fixes #996 pin expiry is updated if set in options

### LINKED ISSUE #996: 'pin update' with new timeout for new pin
It would be nice if we could set the expire time of a new pin which is pushed with `pin update`, while the old pin's expire date is kept unmodified.

Follow-up to:

https://github.com/ipfs/ipfs-cluster/pull/923#issuecomment-577365064

gargdeepak:
1. Ran `go test` in `ipfs-cluster-ctl` go module


gargdeepak:
needs review

hsanjuan:
Closing as stale. Might continue in different PR.

PR #1024, opened 2020-03-28
Fixes #998 Added object expiry time to pin ls cmd.

### LINKED ISSUE #998: showing the 'expire-in' time in pin ls

Currently there's no way to see if there's a expire-in time set or on which date/time the pin will be removed from the cluster.

I think it would be nice if the expire-in time can be listed, or just a 'no' (like for metadata) if there's no expire time set.

gargdeepak:
Tested manually by firing an ipfs-cluster and building the ctl tool


hsanjuan:
Closing as stale. Might continue in different PR.

PR #1025, opened 2020-03-28
Fixes #981 closing channel after 1st message & Ready() returns if context is Done

### LINKED ISSUE #981: Improve handling of Ready() channels
The `Ready()` functions in consensus and main components return a channel which receives a single element when the cluster is ready. This means only one call can be made to them, as there will be only one notification.

The channel should be closed when the cluster/consensus is ready, and the call should return when the contexts are cancelled (either the one passed or the component's main context).

gargdeepak:
Tested with bringing up ipfs-cluster and running `ipfs-cluster-ctl` tests

### DIFF
--- a/consensus/raft/consensus.go
+++ b/consensus/raft/consensus.go
@@ -179,6 +179,7 @@ func (cc *Consensus) finishBootstrap() {
 	logger.Debug("Raft state is now up to date")
 	logger.Debug("consensus ready")
 	cc.readyCh <- struct{}{}
+	close(cc.readyCh)
 }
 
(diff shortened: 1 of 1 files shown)


gargdeepak:
Running failed tests locally to figure out the issue. Apologies for the branch name, it was supposed to be temporary.

gargdeepak:
The test failed in the Travis run seem to be intermittent. I ran it a few times locally, it failed once. What do you recommend should be done ?
https://travis-ci.com/github/ipfs/ipfs-cluster/jobs/306700218

PR #1065, opened 2020-04-11
Fixes #996 pin expiry is updated if set in options

### LINKED ISSUE #996: 'pin update' with new timeout for new pin

gargdeepak:
Running `$ go test `in ipfs-cluster repo shows the following error:
```
23:39:46.480 ERROR     engine: datastore closed engine.go:291
23:39:46.481 ERROR     engine: datastore closed engine.go:291
```
Am I missing a step ?

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -1496,7 +1496,9 @@ func (c *Cluster) PinUpdate(ctx context.Context, from cid.Cid, to cid.Cid, opts 
 	if opts.Name != "" {
 		existing.Name = opts.Name
 	}
-
+	if !opts.ExpireAt.IsZero() && opts.ExpireAt.After(time.Now()) {
+		existing.ExpireAt = opts.ExpireAt
+	}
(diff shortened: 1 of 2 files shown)


hsanjuan:
The tests are not passing and they are showing the reason:

```
TestClustersPinUpdate: ipfscluster_test.go:750: Expiry time didn't match. Expected: 2021-04-11 13:50:30.979525015 +0000 UTC. Got: 2021-04-11 13:50:30 +0000 UTC
```

It may be that we round the dates somewhere. Try fixing the comparison.

gargdeepak:
Thanks @hsanjuan 
I have fixed the test. I'm rounding it to 2s as the test was failing intermittently with the expiry time from cmd being always even.

PR #1083, opened 2020-04-20
Fix #1009: support pinning in direct mode

### LINKED ISSUE #1009: a non-recursive option for the pin-command

I like to be able to pin objects non-recursively to the cluster, like `ipfs pin add --recursive=0 ...` does.

**Additional context**

I like release files in a directory listing earlier than the directory itself. So if no client in the cluster or the network holds a file anymore, the file won't be accessible, but my directory listing (which is non-recursively pinned) still works, to list other files, which are still accessible.

hsanjuan:
This exposes pinning modes (direct/recursive) with a new `mode` API parameter and a new flag in pin/add.

Implementation-wise, this is a new PinOption, which is made to match the MaxDepth field when creating the Pin object. Since Max Depth is already stored we do not need to change the disk serialization.

Additionally some improvements have been added around pinning and unpinning, for speed and resilience.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -462,11 +462,63 @@ func (pT PinType) String() string {
 
 var pinOptionsMetaPrefix = "meta-"
 
+// PinMode is a PinOption that indicates how to pin something on IPFS,
+// recursively or direct.
+type PinMode int
+
+// PinMode values
+const (
+	PinModeRecursive PinMode = 0
+	PinModeDirect    PinMode = 1
+)
+
+// PinModeFromString converst a string to PinMode.
+func PinModeFromString(s string) PinMode {
+	switch s {
+	case "recursive", "":
+		return PinModeRecursive
+	case "direct":
+		return PinModeDirect
+	default:
+		logger.Warn("unknown pin mode. Defaulting to recursive")
+		return PinModeRecursive
+	}
+}
+
+// String returns a human-readable value for PinMode.
+func (pm PinMode) String() string {
+	switch pm {
+	case PinModeRecursive:
+		return "recursive"
+	case PinModeDirect:
+		return "direct"
+	default:
+		return "recursive"
+	}
+}
+
+// MarshalJSON converts the PinMode into a readable string in JSON.
+func (pm PinMode) MarshalJSON() ([]byte, error) {
+	return json.Marshal(pm.String())
+}
+
+// UnmarshalJSON takes a JSON value and parses it into PinMode.
+func (pm *PinMode) UnmarshalJSON(b []byte) error {
+	var s string
+	err := json.Unmarshal(b, &s)
+	if err != nil {
+		return err
+	}
+	*pm = PinModeFromString(s)
+	return nil
+}
+
 // PinOptions wraps user-defined options for Pins
 type PinOptions struct {
 	ReplicationFactorMin int               `json:"replication_factor_min" codec:"rn,omitempty"`
 	ReplicationFactorMax int               `json:"replication_factor_max" codec:"rx,omitempty"`
 	Name                 string            `json:"name" codec:"n,omitempty"`
+	Mode                 PinMode           `json:"mode" codec:"o,omitempty"`
 	ShardSize            uint64            `json:"shard_size" codec:"s,omitempty"`
 	UserAllocations      []peer.ID         `json:"user_allocations" codec:"ua,omitempty"`
 	ExpireAt             time.Time         `json:"expire_at" codec:"e,omitempty"`
@@ -486,6 +538,10 @@ func (po *PinOptions) Equals(po2 *PinOptions) bool {
 	}
 
 	if po.Name != po2.Name {
+		return false
+	}
+
+	if po.Mode != po2.Mode {
 		return false
 	}
 
@@ -538,6 +594,7 @@ func (po *PinOptions) ToQuery() (string, error) {
 	q.Set("replication-min", fmt.Sprintf("%d", po.ReplicationFactorMin))
 	q.Set("replication-max", fmt.Sprintf("%d", po.ReplicationFactorMax))
 	q.Set("name", po.Name)
+	q.Set("mode", po.Mode.String())
 	q.Set("shard-size", fmt.Sprintf("%d", po.ShardSize))
 	q.Set("user-allocations", strings.Join(PeersToStrings(po.UserAllocations), ","))
 	if !po.ExpireAt.IsZero() {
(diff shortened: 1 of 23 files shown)


api/types.go
```diff
@@ -462,11 +462,76 @@ func (pT PinType) String() string {
...
+// UnmarshalJSON takes a JSON value and parses it into PinMode.
+func (pm *PinMode) UnmarshalJSON(b []byte) error {
+	var s string
+	err := json.Unmarshal(b, &s)
+	if err != nil {
+		return err
+	}
+	*pm = PinModeFromString(s)
+	return nil
+}
```
hsanjuan, line 514: For things like PinType, the API shows simply an integer which has not meaning whatsoever for the user. JSON serialization can use meaningful values though, so we do this here.

api/types.go
```diff
@@ -631,6 +704,23 @@ func (po *PinOptions) FromQuery(q url.Values) error {
...
+	switch pd {
+	case -1:
+		return PinModeRecursive
+	case 0:
+		return PinModeDirect
+	default:
+		logger.Warnf("bad pin depth: %d", pd)
+		return PinModeRecursive
+	}
+}
```
hsanjuan, line 722: Since PinMode and Depth are quite related, we need easy conversion between the two. Therefore PinDepth can use it's own type.

api/types.go
```diff
@@ -729,10 +821,11 @@ func (pin *Pin) ProtoMarshal() ([]byte, error) {
 		ReplicationFactorMax: int32(pin.ReplicationFactorMax),
 		Name:                 pin.Name,
 		ShardSize:            pin.ShardSize,
+		Metadata:             pin.Metadata,
+		PinUpdate:            pin.PinUpdate.Bytes(),
+		ExpireAt:             expireAtProto,
+		// Mode:                 pin.Mode,
```
hsanjuan, line 827: Don't need to store this option as we store MaxDepth.

cluster.go
```diff
@@ -1308,8 +1308,8 @@ func checkPinType(pin *api.Pin) error {
 
 // setupPin ensures that the Pin object is fit for pinning. We check
 // and set the replication factors and ensure that the pinType matches the
-// metadata consistently.
-func (c *Cluster) setupPin(ctx context.Context, pin *api.Pin) error {
+// metadata consistently. Returns the existing Pin object for this Cid if it exists.
```
hsanjuan, line 1311: fixme

cluster.go
```diff
@@ -1322,18 +1322,24 @@ func (c *Cluster) setupPin(ctx context.Context, pin *api.Pin) error {
...
 		msg += "clear state with pin rm to proceed. "
 		msg += "New: %s. Was: %s"
 		return fmt.Errorf(msg, pin.Type, existing.Type)
 	}
 
+	if existing.Mode == api.PinModeRecursive && pin.Mode != api.PinModeRecursive {
+		msg := "cannot repin a CID which is already pinned in "
+		msg += "recursive mode (new pin is pinned as %s). Unpin it first."
+		return fmt.Errorf(msg, pin.Mode)
+	}
```
hsanjuan, line 1341: The pinning operation would fail otherwise, as ipfs doesn't let us do this.

cmd/ipfs-cluster-ctl/formatters.go
```diff
@@ -167,7 +167,12 @@ func textFormatPrintVersion(obj *api.Version) {
 }
 
 func textFormatPrintPin(obj *api.Pin) {
-	fmt.Printf("%s | %s | %s | ", obj.Cid, obj.Name, strings.ToUpper(obj.Type.String()))
+	t := strings.ToUpper(obj.Type.String())
+	if obj.Mode == api.PinModeDirect {
+		t = t + "-DIRECT"
+	}
```
hsanjuan, line 173: We signal that a pin is in `direct` mode by showing `PIN-DIRECT`. Recursive pins will show just the cluster `pin` type as before.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -433,31 +446,30 @@ func (ipfs *Connector) Unpin(ctx context.Context, hash cid.Cid) error {
...
 			return err
 		}
-		logger.Info("IPFS Unpin request succeeded:", hash)
-		stats.Record(ctx, observations.Pins.M(-1))
+		logger.Debug("IPFS object is already unpinned: ", hash)
 	}
 
-	logger.Debug("IPFS object is already unpinned: ", hash)
+	logger.Info("IPFS Unpin request succeeded:", hash)
+	stats.Record(ctx, observations.Pins.M(-1))
```
hsanjuan, line 472: For unpins, we stopped checking if the item is pinned with PinLsCid. This means 2 calls for every unpin, makes IPFS load the whole recursive pinset list to check. We just trigger unpin and check the error message, succeeding if it's an ErrNotPinned. It is not awesome that we have to parse an error message, but IPFS gives little change (Error code is set to 0 and http status code is 500, so it's indistinguisable from other errors).

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -491,34 +503,21 @@ func (ipfs *Connector) PinLs(ctx context.Context, typeFilter string) (map[string
...
 }
 
-// PinLsCid performs a "pin ls <hash>" request. It first tries with
-// "type=recursive" and then, if not found, with "type=direct". It returns an
-// api.IPFSPinStatus for that hash.
-func (ipfs *Connector) PinLsCid(ctx context.Context, hash cid.Cid) (api.IPFSPinStatus, error) {
+// PinLsCid performs a "pin ls <hash>" request. It will use "type=recursive" or
+// "type=direct" (or other) depending on the given pin's MaxDepth setting.
+// It returns an api.IPFSPinStatus for that hash.
+func (ipfs *Connector) PinLsCid(ctx context.Context, pin *api.Pin) (api.IPFSPinStatus, error) {
```
hsanjuan, line 509: I changed the signature of PinLsCid so that we can get a full Pin object and we can check `direct` or `recursive` direclty, without having to trigger two requests for checking if a direct pin is pinned.

This affects the RPC api, but since this is a local-only endpoint, we don't need to bump the RPC protocol version.

ipfsconn/ipfshttp/ipfshttp.go
```diff
@@ -575,12 +574,9 @@ func checkResponse(path string, res *http.Response) ([]byte, error) {
...
 		if err := json.Unmarshal(body, &ipfsErr); err == nil {
-			return body, fmt.Errorf(
-				"IPFS request unsuccessful (%s). Code: %d. Message: %s",
-				path,
-				res.StatusCode,
-				ipfsErr.Message,
-			)
+			ipfsErr.code = res.StatusCode
+			ipfsErr.path = path
+			return body, ipfsErr
```
hsanjuan, line 579: Take advantage of making the ipfsError type a Go error itself.

PR #1131, opened 2020-05-17
Fixes to BlockPut

hsanjuan:


hsanjuan:
The first issue broke adding completely, if it wasn't because the second issue completely ignored the errors.

PR #1160, opened 2020-06-08
config: reduce default pin timeout to 2 minutes

RubenKelevra:


hsanjuan:
@RubenKelevra can you follow up with a PR to the ipfs/ipfs-cluster-website (configuration reference needs to change the default).

RubenKelevra:
@hsanjuan wrote:

> @RubenKelevra can you follow up with a PR to the ipfs/ipfs-cluster-website (configuration reference needs to change the default).

Done: https://github.com/ipfs/ipfs-cluster-website/pull/116

Issue #1085, opened 2020-04-21
Make sure dockerfiles work well in any architectures


hsanjuan:
**Describe the feature you are proposing**

Our docker containers should be able to run in different architectures and be built for them.

**Additional context**

Go-ipfs is probablly a good example on how to do this. This needs porting of things like https://github.com/krallin/tini/releases/download .

omkarprabhu-98:
Link: https://github.com/krallin/tini/releases/

Docker has an experimental CLI plugin for multi-arch builds
- Buildx: https://docs.docker.com/buildx/working-with-buildx/
- Blog: https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

hsanjuan:
Usually we can copy most of the things from how go-ipfs does it.

omkarprabhu-98:
Ah okay!
I'll make an attempt, any suggestions on how to test it to make sure it works?

hsanjuan:
Here's an easy dockerfile which I think supports multiarch and includes some other nice things we haven't used yet: https://github.com/ipfs/rainbow/blob/main/Dockerfile

hsanjuan:
https://github.com/ipfs/kubo/blob/master/.github/workflows/docker-image.yml too as reference.

PR #1196, opened 2020-07-22
Feat 1085: Dockerfile to support multiple architectures

omkarprabhu-98:
#1085 

Changes: 
- Downloads appropriate tini binary based on the machine architecture https://github.com/ipfs/go-ipfs/pull/7187
- Builds a static su-exec binary to make it robust (no dependency on libraries) and cross-platform support https://github.com/ipfs/go-ipfs/pull/7505

For uploading images with different machine architectures to DockerHub, [Buildx](https://docs.docker.com/buildx/working-with-buildx/) or some other tool may be required

Any suggestions on how to test this for architectures other than amd64? 
Other libraries required would have to be imported as shown here  https://github.com/ipfs/go-ipfs/pull/6854

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,15 +9,20 @@ ENV GO111MODULE on
 ENV GOPROXY     https://proxy.golang.org
 
 ENV SUEXEC_VERSION v0.2
-ENV TINI_VERSION v0.16.1
-RUN set -x \
-  && cd /tmp \
+ENV TINI_VERSION v0.19.0
+RUN set -eux; \
+    dpkgArch="$(dpkg --print-architecture)"; \
+    case "${dpkgArch##*-}" in \
+        "amd64" | "armhf" | "arm64") tiniArch="tini-static-$dpkgArch" ;;\
+        *) echo >&2 "unsupported architecture: ${dpkgArch}"; exit 1 ;; \
+    esac; \
+  cd /tmp \
   && git clone https://github.com/ncopa/su-exec.git \
   && cd su-exec \
   && git checkout -q $SUEXEC_VERSION \
-  && make \
+  && make su-exec-static \
   && cd /tmp \
-  && wget -q -O tini https://github.com/krallin/tini/releases/download/$TINI_VERSION/tini \
+  && wget -q -O tini https://github.com/krallin/tini/releases/download/$TINI_VERSION/$tiniArch \
   && chmod +x tini
 
 # Get the TLS CA certificates, they're not provided by busybox.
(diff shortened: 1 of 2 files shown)


hsanjuan:
> Any suggestions on how to test this for architectures other than amd64?

Maybe I can test in a raspberry Pi. Change looks OK, will leave unmerged until I can test it.

Jorropo (changes requested):
Can you rebase on master pls ?
Also this :
https://github.com/ipfs/ipfs-cluster/blob/3ef09eec8f33bb0606a6c96e3b16da4c735df5c9/Dockerfile-test#L54
Need to be changed, `x86_64-linux-gnu` has to be changed depending of the configuration.

omkarprabhu-98:
@Jorropo please take a look now

Jorropo:
> @Jorropo please take a look now

@omkarprabhu-98 error like this on an arm64 manjaro:
```
Step 36/40 : RUN mkdir -p $IPFS_CLUSTER_PATH &&     chown 1000:100 $IPFS_CLUSTER_PATH
 ---> Running in 3951667724d7
standard_init_linux.go:219: exec user process caused: exec format error
The command '/bin/sh -c mkdir -p $IPFS_CLUSTER_PATH &&     chown 1000:100 $IPFS_CLUSTER_PATH' returned a non-zero code: 1
make: *** [Makefile:62 : docker] Erreur 1
```
I'll maybe dig it later.

Jorropo:
@omkarprabhu-98 I've found the issue, we are pulling the `ipfs/go-ipfs:master` images :
https://github.com/ipfs/ipfs-cluster/blob/3ef09eec8f33bb0606a6c96e3b16da4c735df5c9/Dockerfile-test#L23
But IPFS doesn't support ARM docker build : you would need to fix ipfs/go-ipfs#7541 and pull the right image to work.
PS: You could also rebuild go-ipfs here but why not doing this in the go-ipfs repo (plus this would be awsome at the same time) ? Plus this will very likely get merged in the go-ipfs repo (see the help wanted tag in the issue).

Jorropo:
If you want some more insight about what to do check this :
https://github.com/ipfs/go-ipfs/issues/7541#issuecomment-661209028
Basicaly the docker hub image is built using the build made on circle CI host machine (amd64) :
https://github.com/ipfs/go-ipfs/blob/480defab689610550ee3d346e31441a2bb881fcb/.circleci/config.yml#L320

hsanjuan:
Note that `Dockefile-test` is not really useful to anyone anymore. If that's the blocker, it can be fully ignore (we definitely don't need to worry about arm builds for this one).

Jorropo (approved):
> Note that `Dockefile-test` is not really useful to anyone anymore. If that's the blocker, it can be fully ignore (we definitely don't need to worry about arm builds for this one).

@hsanjuan 
In this case it works on my arm64 manjaro, LGTM.

PR #1227, opened 2020-08-28
Deprecate secio to noise and fix docker build.

### LINKED ISSUE #1214: Deprecate secio
Related: https://github.com/libp2p/go-libp2p/pull/972

Cluster nodes should be backwards compatible because we are using TLS preferentially to secio, so removing it should be no problem. We should add noise as first option though (`clusterhost.go`).

Jorropo:
Closes #1214

### DIFF
--- a/clusterhost.go
+++ b/clusterhost.go
@@ -17,10 +17,10 @@ import (
 	routing "github.com/libp2p/go-libp2p-core/routing"
 	dht "github.com/libp2p/go-libp2p-kad-dht"
 	dual "github.com/libp2p/go-libp2p-kad-dht/dual"
+	noise "github.com/libp2p/go-libp2p-noise"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	libp2pquic "github.com/libp2p/go-libp2p-quic-transport"
 	record "github.com/libp2p/go-libp2p-record"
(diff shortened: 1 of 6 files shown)


Jorropo:
@hsanjuan this should also fix the docker build (it was calling an non existent image).

clusterhost.go
```diff
@@ -118,7 +118,7 @@ func baseOpts(psk corepnet.PSK) []libp2p.Option {
 		libp2p.PrivateNetwork(psk),
 		libp2p.EnableNATService(),
 		libp2p.Security(libp2ptls.ID, libp2ptls.New),
-		libp2p.Security(secio.ID, secio.New),
+		libp2p.Security(noise.ID, noise.New),
```
hsanjuan, line 121: Noise should be first (move before tls).

hsanjuan (changes requested):
Small ask. Thanks for contributing this!

Jorropo:
@hsanjuan done

hsanjuan:
@Jorropo it seems docker is still borked

Jorropo:
@hsanjuan it works now (on my machine at least) :)

Jorropo:
It finally builds !

hsanjuan (approved):

PR #1237, opened 2020-09-01
Support connecting to the IPFS API via Unix domain sockets

AluisioASG:
Same as ipfs/go-ipfs#6678. I haven't tested it with tracing on, would be nice if someone using that did it.

I'm looking at how to add tests to it.

### DIFF
--- a/ipfsconn/ipfshttp/ipfshttp.go
+++ b/ipfsconn/ipfshttp/ipfshttp.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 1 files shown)


hsanjuan (changes requested):
Thanks for sending this.

I don't think this works. There is a `apiURL()` method that is used in many places and that does:

```go
fmt.Sprintf("http://%s/api/v0", ipfs.nodeAddr)
```

and I guess it should instead print something like `unix://path-to-socket` (?).

About testing, I'd suggest:

1. Modifying https://github.com/ipfs/ipfs-cluster/blob/master/test/ipfs_mock.go#L135 adding a `unixSocketPath string` parameter to the function perhaps.
2. When set, configure the `httptest.Server` (https://golang.org/pkg/net/http/httptest/#Server) with a unix listener. (Needs to do NewUnstartedServer() first and then Start after replacing the Listener).
3. Update the function signature wherever is called in the codebase
4. Add an additional simple test that uses a `NewIpfsMock` with a unix listener instead of the normal. Make sure that something simple like a call to .ID() works.

All that said, this is a very good idea and glad to support it.

hsanjuan:
Closing this as it is stale. Can re-open at a later point.

Issue #1222, opened 2020-08-24
Integrate the failover client in ipfs-cluster-ctl


hsanjuan:
**Describe the feature you are proposing**

ipfs-cluster-ctl should use the restapi LBClient with a Failover strategy.

Particularly, it should also be possible to create a Failover and a RoundRobin client strategy from a /dnsaddr. If the record resolves to multiple multiaddresses, each client should be initialized to one of them.

This needs a method to turn a single Config into multiple ones that can be used for `NewLBClient`, each of them with one multiaddress as resolved from the `/dnsaddr`. i.e. `config.AsTemplateFor([]multiaddresses) []Config` and `config.AsTemplateForResolvedAddresses(multiaddress) []Config`.

This should in the end enable `ipfs-cluster-ctl` to  perform auto-failover to other resolved addresses in a multiaddress.

omkarprabhu-98:
So the input through the command line would be a list of client config attributes that would be converted to a list of configs and then be passed to create the LBClient?

hsanjuan:
The input form the command line would be `ipfs-cluster-ctl --host /dnsaddr/my.host ...`.

Initializing the LB/Failover clients requires passing each 1 configuration for each endpoint in it. Thus, we need helper methods to, given a template configuration, produce multiple configs only changing the client multiaddress. As mentioned above, an additional method can actually perform resolving:

AsTemplateForResolvedAddresses(address) -> resolves -> calls AsTemplateFor(resolved_addresses).

Once you have these two helper methods, it should be easy to update ipfs-cluster-ctl to use the Failover client.

PR #1250, opened 2020-10-04
Feat #1222: Integrates the failover client in ipfs-cluster-ctl

omkarprabhu-98:
#1222 

Changes:
- adds methods to create client configs from a dnsaddr host input
- uses the LBClient in ipfs-cluster-ctl

### DIFF
--- a/api/rest/client/client.go
+++ b/api/rest/client/client.go
@@ -166,21 +166,21 @@ type Config struct {
 }
 
 // AsTemplateFor creates client configs from resolved multiaddresses
-func (c *Config) AsTemplateFor(resolvedAddrs []ma.Multiaddr) ([]*Config) {
+func (c *Config) AsTemplateFor(addrs []ma.Multiaddr) ([]*Config) {
 	var cfgs []*Config
-	for _, multiaddress := range resolvedAddrs{
+	for _, addr := range addrs {
 		cfg := *c
-		cfg.APIAddr = multiaddress
+		cfg.APIAddr = addr
 		cfgs = append(cfgs, &cfg)
 	}
 	return cfgs
 }
 
 // AsTemplateForResolvedAddress creates client configs from a multiaddress
-func (c *Config) AsTemplateForResolvedAddress(addr ma.Multiaddr) ([]*Config, error) {
-	resolvedAddrs, err := resolveAddr(context.Background(), addr)
+func (c *Config) AsTemplateForResolvedAddress(ctx context.Context, addr ma.Multiaddr) ([]*Config, error) {
+	resolvedAddrs, err := resolveAddr(ctx, addr)
 	if err != nil {
-		return []*Config{}, err
+		return nil, err
 	}
 	return c.AsTemplateFor(resolvedAddrs), nil
 }
@@ -386,11 +386,11 @@ func resolveAddr(ctx context.Context, addr ma.Multiaddr) ([]ma.Multiaddr, error)
 	defer cancel()
 	resolved, err := madns.Resolve(resolveCtx, addr)
 	if err != nil {
-		return []ma.Multiaddr{}, err
+		return nil, err
 	}
 	
 	if len(resolved) == 0 {
-		return []ma.Multiaddr{}, fmt.Errorf("resolving %s returned 0 results", addr)
+		return nil, fmt.Errorf("resolving %s returned 0 results", addr)
 	}
 
 	return resolved, nil
--- a/api/rest/client/client.go
+++ b/api/rest/client/client.go
@@ -414,8 +414,7 @@ func resolveDNSAddr(addr ma.Multiaddr) ([]ma.Multiaddr, error) {
 	for len(currResolved) != len(prevResolved) {
 		var tobeResolved []ma.Multiaddr
 		for _, addr := range currResolved {
-			ctx, cancel := context.WithCancel(context.Background())
-			resolveCtx, cancel := context.WithTimeout(ctx, ResolveTimeout)
+			resolveCtx, cancel := context.WithTimeout(context.Background(), ResolveTimeout)
 			defer cancel()
 			resolved, err := madns.Resolve(resolveCtx, addr)
 			if err != nil {
--- a/api/rest/client/client.go
+++ b/api/rest/client/client.go
@@ -165,6 +165,26 @@ type Config struct {
 	LogLevel string
 }
 
+// AsTemplateFor creates client configs from resolved multiaddresses
+func (c *Config) AsTemplateFor(resolvedAddrs []ma.Multiaddr) ([]*Config) {
+	var cfgs []*Config
+	for _, multiaddress := range resolvedAddrs{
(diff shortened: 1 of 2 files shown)


api/rest/client/client.go
```diff
@@ -366,3 +386,45 @@ func isUnixSocketAddress(addr ma.Multiaddr) bool {
...
+	return (dnsaddr != "" && err == nil)
+}
+
+// resolve dnsaddr
+func resolveDNSAddr(addr ma.Multiaddr) ([]ma.Multiaddr, error) {
+	var prevResolved []ma.Multiaddr
+	var currResolved []ma.Multiaddr
+
+	// Only resolve dnsaddr
+	if !isDNSAddress(addr) {
```
hsanjuan, line 406: Comment above. Also, resolveAPIAddr() should call this helper function.

api/rest/client/client.go
```diff
@@ -366,3 +386,45 @@ func isUnixSocketAddress(addr ma.Multiaddr) bool {
...
+			resolveCtx, cancel := context.WithTimeout(context.Background(), ResolveTimeout)
+			defer cancel()
+			resolved, err := madns.Resolve(resolveCtx, addr)
+			if err != nil {
+				return []ma.Multiaddr{}, err
+			}
+			tobeResolved = append(tobeResolved, resolved...)
+		}
+		prevResolved = currResolved
+		currResolved = tobeResolved
```
hsanjuan, line 426: Resolve once. This is a bit confusing. Even if recursive resolution would bring something we may just want to load balance on the first-resolved addresses only.

hsanjuan (changes requested):
Thanks! Some comments below.

api/rest/client/client.go
omkarprabhu-98: oh alright. I was looking at cases like this https://github.com/multiformats/multiaddr/blob/master/protocols/DNSADDR.md where the root dnsaddr resolves into other dnsaddrs

hsanjuan: Let's take the simple approach now. dnsaddr addresses that resolve to other dnsaddr addresses is something I haven't seen used in reality, and if it was perhaps it would be `madns` problem to handle recursive resolution.

api/rest/client/client.go
```diff
@@ -366,3 +379,19 @@ func isUnixSocketAddress(addr ma.Multiaddr) bool {
...
 	return (value != "" && err == nil)
 }
+
+// resolve addr
+func resolveAddr(ctx context.Context, addr ma.Multiaddr) ([]ma.Multiaddr, error) {
+	resolveCtx, cancel := context.WithTimeout(ctx, ResolveTimeout)
+	defer cancel()
+	resolved, err := madns.Resolve(resolveCtx, addr)
+	if err != nil {
+		return []ma.Multiaddr{}, err
```
hsanjuan, line 389:
```suggestion
		return nil, err
```

api/rest/client/client.go
```diff
@@ -366,3 +379,19 @@ func isUnixSocketAddress(addr ma.Multiaddr) bool {
...
+	}
+	
+	if len(resolved) == 0 {
+		return []ma.Multiaddr{}, fmt.Errorf("resolving %s returned 0 results", addr)
```
hsanjuan, line 393:
```suggestion
		return nil, fmt.Errorf("resolving %s returned 0 results", addr)
```

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -89,10 +89,10 @@ func main() {
 	app.Description = Description
 	app.Version = Version
 	app.Flags = []cli.Flag{
-		cli.StringFlag{
+		cli.StringSliceFlag{
 			Name:  "host, l",
-			Value: defaultHost,
-			Usage: "Cluster's HTTP or LibP2P-HTTP API endpoint",
+			Value: &cli.StringSlice{defaultHost}, 
+			Usage: "Cluster's HTTP or LibP2P-HTTP API endpoint. To provide multiple hosts: --host a --host b",
```
hsanjuan, line 95: There is a non-written style rule that we provide lists to arguments as comma-separated values rather than using the StringSlice way. This means you keep the String flag, but do a `strings.Split` with the value when processing it.

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -174,7 +166,26 @@ requires authorization. implies --https, which you can disable with --force-http
...
+			cfgs, err := cfg.AsTemplateForResolvedAddress(multiaddr)
+			checkErr("creating configs", err)
+			configs = append(configs, cfgs...)
+		}
+
+		for _, cfg := range configs {
+			if client.IsPeerAddress(cfg.APIAddr) && c.Bool("https") {
+				logger.Warn("Using libp2p-http. SSL flags will be ignored")
+				break;
+			}
```
hsanjuan, line 184: I think we can include this warning in the loop above but updating the warning message to `Using libp2p-http for %s. The https flag will be ignored for this connection`. It is ok if it appears multiple times.

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -174,7 +166,26 @@ requires authorization. implies --https, which you can disable with --force-http
...
 		}
 
-		globalClient, err = client.NewDefaultClient(cfg)
+		var configs []*client.Config
+		var err error
+		for _, addr := range c.StringSlice("host") {
+			multiaddr, err := ma.NewMultiaddr(addr)
+			checkErr("parsing host multiaddress", err)
+
+			cfgs, err := cfg.AsTemplateForResolvedAddress(multiaddr)
```
hsanjuan, line 175: This would receive a context. `c` is a context already, I think it can be passed directly.

hsanjuan (changes requested):
Thanks, after changes here it should be ready to merge.

cmd/ipfs-cluster-ctl/main.go
omkarprabhu-98: I think c (`c *cli.Context`) is a different context struct provided by the library. I can use the `ctx` variable created as soon as `main` function begins.

hsanjuan: You're right, I was looking at urfave/cli/v2 documentation where it actually implements `context.Context`. Ok, use that global context for the moment.

omkarprabhu-98:
@hsanjuan let me know if any other changes are required. The build is failing due to cluster-ctl help's 120 char limit should I increase it or try to reduce the help output size

hsanjuan:
> @hsanjuan let me know if any other changes are required. The build is failing due to cluster-ctl help's 120 char limit should I increase it or try to reduce the help output size

You should reduce your strings to fit the limit.

hsanjuan:
Sorry, super busy this week, will eventually get to this.

hsanjuan (approved):
Thank you very much for this contribution!

PR #1301, opened 2021-01-21
feat: add --wait flag to ipfs-cluster-ctl add

### LINKED ISSUE #1285: Support --wait in ipfs-cluster-ctl add

Support the `--wait` flag just like in `pin add`. This makes mostly sense when adding with `--local`, since that is when the process of adding might be finished and the process of pinning may take a while longer (otherwise it should be pretty quick).

olizilla:
A "simplest thing that could work" implementation of adding a `--wait` flag to the ipfs-cluster-ctl add command. Allows CI to wait for cluster to fully replicate the files just added before continuting, or fail if replication fails.

A dull thing with this change is that `add -w` is already used as the alias for `add --wrap-with-directory`, so the meaning of `-w` is inconsistent between the `add` and `pin` commands. It seemed unwise to change the existing meaning of `add -w` so I have not updated it here.

Fixes #1285

License: MIT
Signed-off-by: Oli Evans <oli@tableflip.io>

### DIFF
--- a/cmd/ipfs-cluster-ctl/main.go
+++ b/cmd/ipfs-cluster-ctl/main.go
@@ -385,6 +385,15 @@ content.
 					Name:  "nocopy",
 					Usage: "Add the URL using filestore. Implies raw-leaves. (experimental)",
 				},
+				cli.BoolFlag{
+					Name:  "wait",
+					Usage: "Wait for all nodes to report a status of pinned before returning",
+				},
+				cli.DurationFlag{
+					Name:  "wait-timeout, wt",
+					Value: 0,
+					Usage: "How long to --wait (in seconds), default is indefinitely",
+				},
 				// TODO: Uncomment when sharding is supported.
 				// cli.BoolFlag{
 				//	Name:  "shard",
@@ -490,6 +499,12 @@ content.
 					if lastBuf == nil || lastBuf.AddedOutput == nil {
 						return // no elements at all
 					}
+
+					if c.Bool("wait") {
+						var _, cerr = waitFor(lastBuf.AddedOutput.Cid, api.TrackerStatusPinned, c.Duration("wait-timeout"))
+						checkErr("waiting for pin status", cerr)
+					}
+
 					if bufferResults { // we buffered.
 						if qq { // [last elem]
 							formatResponse(c, []*addedOutputQuiet{lastBuf}, nil)
--- a/cmd/ipfs-cluster-ctl/main.go
+++ b/cmd/ipfs-cluster-ctl/main.go
@@ -499,12 +499,6 @@ content.
 					if lastBuf == nil || lastBuf.AddedOutput == nil {
 						return // no elements at all
 					}
-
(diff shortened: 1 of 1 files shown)


olizilla:
I'm a novice gopher, so this is very much a first pass, copy-paste and tweak PR, so that I can get some understanding of how cluster works.

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -490,6 +499,12 @@ content.
 					if lastBuf == nil || lastBuf.AddedOutput == nil {
 						return // no elements at all
 					}
+
+					if c.Bool("wait") {
+						var _, cerr = waitFor(lastBuf.AddedOutput.Cid, api.TrackerStatusPinned, c.Duration("wait-timeout"))
+						checkErr("waiting for pin status", cerr)
+					}
+
```
hsanjuan, line 507: I have the feeling this should happen after the `if.. else `. So that the user can see the added CID already, even though the command will not return until pinned.

hsanjuan (changes requested):
Happy taking this patch (with the comment addressed).

I am wondering if `--wait` should return a status report at the end (like pin-add). But I'm not sure, so we can just do nothing for the moment.

olizilla:
Done!

cmd/ipfs-cluster-ctl/main.go
```diff
@@ -501,6 +510,10 @@ content.
 						formatResponse(c, lastBuf, nil)
 						return
 					}
+					if c.Bool("wait") {
```
hsanjuan, line 513: There are `return` statements above that need to be removed so that the code-flow always hits the `wait` part. Right now it will disregard the wait flag if --quieter or --no-stream are passed, for example.

hsanjuan (changes requested):

cmd/ipfs-cluster-ctl/main.go
olizilla: uff. I should not push code in the evening. fixed!

gmasgras:
I'm probably missing something but how is this different behaviour from not passing `--local` to `add` ?

olizilla:
`add` will add the blocks to ipfs, and return the CID. It then replicates them across the cluster. Either way, until we add `--wait`, the `add` command won't wait for replication to complete.

hsanjuan (approved):

hsanjuan:
> I'm probably missing something but how is this different behaviour from not passing `--local` to `add` ?

In cluster, pinning is by default async. You send the pin and the call returns when cluster grabbed it. The process of pinning it on IPFS happens in the background. The --wait flags are a workaround to hold the pin/add commands until the item has actually reached "pinned" state.

The --local flag means BlockPuts calls happen locally on the peer receiving an add request, as opposed to somewhere else (via broadcasting). If using a `ipfs-cluster-ctl --host <remote node> add --local`, things happen on `<remote node>` (which is confusing). Most options are server-side (with --wait being an exception).

PR #1343, opened 2021-04-19
Feat: support adding CAR files

hsanjuan:
This commit adds a new add option: "format".

This option specifies how IPFS Cluster is expected to build the DAG when
adding content. By default, it takes a "unixfs", which chunks and DAG-ifies as
it did before, resulting in a UnixFSv1 DAG.

Alternatively, it can be set to "car". In this case, Cluster will directly
read blocks from the CAR file and add them.

Adding CAR files or doing normal processing is independent from letting
cluster do sharding or not. If sharding is ever enabled, Cluster could
potentially shard a large CAR file among peers.

Currently, importing CAR files is limited to a single CAR file with a single
root (the one that is pinned). Future iterations may support multiple CARs
and/or multiple roots by transparently wrapping them.

### DIFF
--- a/adder/adder.go
+++ b/adder/adder.go
@@ -4,15 +4,19 @@ package adder
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"mime/multipart"
 	"strings"
 
 	"github.com/ipfs/ipfs-cluster/adder/ipfsadd"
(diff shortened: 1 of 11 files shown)


hsanjuan (commented):
LGTM

PR #1346, opened 2021-04-28
crdt: Add batching support

### LINKED ISSUE #1008: CRDT-Batching support

go-ds-crdt supports batching. Multiple updates can be included in the same DAG node. With batching, we do not need to issue one block per update.

Cluster with heavy pinning/unpinning intake can take advantage of batching to greatly reduce (I think around ~20000 pins can go in a batch) the DAG size.

Batching can be two ways:

* Time based: A batch will be published every X seconds, if any updates have been made. This allows to group all updates every 5 seconds in a single batch.
* Size based: A batch will be published as soon as X updates have been made.

Both approaches can be combined (whatever condition is hit first issues the batch). i.e. A batch should be issued every 100 updates, but in should be sent in any case if 10 minutes have passed from the previous batch.

Note that in-principle batch support is in-memory. Uncommitted batches will be lost.

Batching should be configurable in the `crdt` config section:

```json
"batching" : {
  "max_batch_size": number_of_updates,
  "max_batch_age": "duration"
}
```

0 values disable batching.

hsanjuan:
This adds batching support to crdt-consensus per #1008 . The crdt component can now take
advantage of the BatchingState, which uses the batching-crdt datastore. In
batching mode, the crdt datastore groups any Add and Delete operations
in a single delta (instead of just 1, as it does by default).

Batching is enabled in the crdt configuration section by setting MaxBatchSize
**and** MaxBatchAge. These two settings control when a batch is committed,
either by reaching a maximum number of pin/unpin operations, or by reaching a
maximum age.

Batching unlocks large pin-ingestion scalability for clusters, but should be
set according to expected work loads. An additional, hidden MaxQueueSize
parameter provides the ability to perform backpressure on Pin/Unpin
requests. When more than MaxQueueSize pin/unpins are waiting to be included in
a batch, the LogPin/LogUnpin operations will fail. If this happens, it is
means cluster cannot commit batches as fast as pins are arriving. Thus,
MaxQueueSize should be increase (to accommodate bursts), or the batch size
increased (to perform less commits and hopefully handle the requests faster).

Note that the underlying CRDT library will auto-commit when batch deltas reach
1MB of size.

Fixes #1008.

### DIFF
--- a/consensus/crdt/config.go
+++ b/consensus/crdt/config.go
@@ -25,7 +25,7 @@ var (
 	DefaultRebroadcastInterval  = time.Minute
 	DefaultTrustedPeers         = []peer.ID{}
 	DefaultTrustAll             = true
-	DefaultBatchingMaxQueueSize = 1000
+	DefaultBatchingMaxQueueSize = 50000
 )
 
 // BatchingConfig configures parameters for batching multiple pins in a single
--- a/consensus/crdt/config.go
+++ b/consensus/crdt/config.go
@@ -19,13 +19,31 @@ var envConfigKey = "cluster_crdt"
 
 // Default configuration values
 var (
-	DefaultClusterName         = "ipfs-cluster"
-	DefaultPeersetMetric       = "ping"
-	DefaultDatastoreNamespace  = "/c" // from "/crdt"
-	DefaultRebroadcastInterval = time.Minute
-	DefaultTrustedPeers        = []peer.ID{}
-	DefaultTrustAll            = true
+	DefaultClusterName          = "ipfs-cluster"
+	DefaultPeersetMetric        = "ping"
+	DefaultDatastoreNamespace   = "/c" // from "/crdt"
+	DefaultRebroadcastInterval  = time.Minute
+	DefaultTrustedPeers         = []peer.ID{}
+	DefaultTrustAll             = true
+	DefaultBatchingMaxQueueSize = 1000
 )
+
+// BatchingConfig configures parameters for batching multiple pins in a single
+// CRDT-put operation.
+//
+// MaxBatchSize will trigger a commit whenever the number of pins in the batch
+// reaches the limit.
+//
+// MaxBatchAge will trigger a commit when the oldest update in the batch
+// reaches it. Setting both values to 0 means batching is disabled.
+//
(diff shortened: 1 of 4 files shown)


hsanjuan:
Tests missing.

consensus/crdt/config.go
```diff
@@ -87,6 +116,10 @@ func (cfg *Config) Validate() error {
 	if cfg.RebroadcastInterval <= 0 {
 		return errors.New("crdt.rebroadcast_interval is invalid")
 	}
+
+	if cfg.Batching.MaxQueueSize <= 0 {
+		return errors.New("crdt.batching.max_queue_size is invalid")
+	}
```
alanshaw, line 122: Also validate `MaxBatchSize` is less than `MaxQueueSize`?

consensus/crdt/config.go
```diff
@@ -19,14 +19,32 @@ var envConfigKey = "cluster_crdt"
...
+// CRDT-put operation.
+//
+// MaxBatchSize will trigger a commit whenever the number of pins in the batch
+// reaches the limit.
+//
+// MaxBatchAge will trigger a commit when the oldest update in the batch
+// reaches it. Setting both values to 0 means batching is disabled.
+//
+// MaxQueueSize specifies how many items can be waiting to be batched before
+// the LogPin/Unpin operations block.
```
alanshaw, line 41:
```suggestion
// the LogPin/Unpin operations fail.
```

consensus/crdt/config.go
hsanjuan: They are not related. There is a goroutine building the batches from the information that is put on a channel and commiting when the time comes. MaxQueueSize is the buffer size of that channel. Regardless of the MaxBatchSize, the queue will only get filled if the goroutine is unable to batch-add+commit things as fast as the requests arrive.

I can think of MaxBatchSize = 5000 while MaxQueueSize is left at the default of 1000 and that should work, I guess.

hsanjuan: This made me think and I bumpbed the default MaxQueueSize to 50000. I don't think this limit should be hit during normal operation regardless of maxBatchSize.

alanshaw: Makes sense!

PR #1359, opened 2021-05-19
Fix #1358: Set RawLeaves to true when unset and CidVersion=1

### LINKED ISSUE #1358: Inconsistent hash when using v1 CID
When adding content with CID v1 `ipfs add file --cid-version=1` I get a different CID to when I add content with the IPFS Cluster HTTP API, passing `?cid-version=1`.

In IPFS, if you set `--cid-version=1` it implicitly sets `--raw-leaves=true` unless you explicitly pass `--raw-leaves=false`.

I believe that Cluster explicitly passes `?raw-leaves=false` on to IPFS even if you do not explicitly call the cluster API with this parameter and this is the reason for this discrepancy.

The workaround is to explicitly pass `?cid-version=1&raw-leaves=true`, but the current behaviour may be a surprise for people expecting to get the same CID from cluster as they get on the command line when using v1 CIDs.

hsanjuan:
Fixes #1358. Mimics go-ipfs defaults.

### DIFF
--- a/api/add.go
+++ b/api/add.go
@@ -154,10 +154,6 @@ func AddParamsFromQuery(query url.Values) (*AddParams, error) {
 		return nil, err
 	}
 
-	err = parseBoolParam(query, "raw-leaves", &params.RawLeaves)
-	if err != nil {
-		return nil, err
-	}
 	err = parseBoolParam(query, "hidden", &params.Hidden)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


api/add.go
```diff
@@ -181,6 +177,19 @@ func AddParamsFromQuery(query url.Values) (*AddParams, error) {
...
 	}
 
+	// This mimics go-ipfs behaviour.
+	if params.CidVersion > 0 {
+		params.RawLeaves = true
+	}
+
+	// If the raw-leaves param is empty, the default RawLeaves value will
+	// take place (which may be true or false depending on
+	// CidVersion). Otherwise, it will be explicitally set.
```
alanshaw, line 187:
```suggestion
	// CidVersion). Otherwise, it will be explicitly set.
```

api/add_test.go
```diff
@@ -28,6 +28,53 @@ func TestAddParams_FromQuery(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	p, err = AddParamsFromQuery(q)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !p.RawLeaves {
+		t.Error("RawLeaves should be true when explicitally set")
```
alanshaw, line 74:
```suggestion
		t.Error("RawLeaves should be true when explicitly set")
```

alanshaw (approved):
LGTM, just a couple of typos.

PR #1372, opened 2021-06-28
Fix #1362: Add "size" to response when adding CARs with a unixfs file root

### LINKED ISSUE #1362: Document why add --format=car returns `bytes` not `size`
I'm working on CAR import for nft.storage, and it took me a while to figure out why switching the format to CAR cause the response body to have a `bytes` propoerty rathe than a `size` property. I'm assuming that it's because we don't know if the CAR contains a unixfs root, so we just return the raw byte size of the blocks.

As a nice to have, could cluster return the `size` property if the root block is unixfs

```console
$ curl -X POST -F 'file=@/Users/oli/Pictures/pinpie.jpg' 'http://127.0.0.1:9094/add?chunker=size-262144&cid-version=1&format=unixfs&hash=sha2-256&hidden=false&layout=&local=false&mode=recursive&name=&nocopy=false&progress=false&raw-leaves=true&recursive=false&replication-max=0&replication-min=0&shard=false&shard-size=104857600&stream-channels=true&user-allocations=&wrap-with-directory=false'
{"name":"pinpie.jpg","cid":{"/":"bafkreiajkbmpugz75eg2tmocmp3e33sg5kuyq2amzngslahgn6ltmqxxfa"},"size":47874}

$ curl -X POST -F 'file=@/Users/oli/Pictures/pinpie.car' 'http://127.0.0.1:9094/add?chunker=size-262144&cid-version=1&format=car&hash=sha2-256&hidden=false&layout=&local=false&mode=recursive&name=&nocopy=false&progress=false&raw-leaves=true&recursive=false&replication-max=0&replication-min=0&shard=false&shard-size=104857600&stream-channels=true&user-allocations=&wrap-with-directory=false'
{"name":"pinpie.car","cid":{"/":"bafkreiajkbmpugz75eg2tmocmp3e33sg5kuyq2amzngslahgn6ltmqxxfa"},"bytes":47874}
```

hsanjuan:
This sets the "size" value to the FileSize() value when the CAR file's root
is a unixfs file. (Folders are not files, size will still be 0).

### DIFF
--- a/adder/adder.go
+++ b/adder/adder.go
@@ -10,6 +10,7 @@ import (
 	"mime/multipart"
 	"strings"
 
+	"github.com/ipfs/go-unixfs"
 	"github.com/ipfs/ipfs-cluster/adder/ipfsadd"
 	"github.com/ipfs/ipfs-cluster/api"
 	"github.com/ipld/go-car"
@@ -282,6 +283,7 @@ func (ca *carAdder) Add(name string, fn files.Node) (cid.Cid, error) {
 
 	root := carReader.Header.Roots[0]
 	bytes := uint64(0)
+	size := uint64(0)
 
 	for {
 		block, err := carReader.Next()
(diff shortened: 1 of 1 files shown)


hsanjuan:
@olizilla do you like this?

PR #1377, opened 2021-07-06
Fix #1360: Efficient pinset status with filters

### LINKED ISSUE #1360: Implement status filtering in PinTracker/Status
Currently, the REST API allows pin status filtering using TrackerStatus flags like:

https://github.com/ipfs/ipfs-cluster/blob/bfe179e94343321484f423d882b8770a6ecd4354/api/types.go#L84-L85

This happens on the API after the `StatusAll` RPC call to the pintracker has happened.

The StatusAll implementation lists the local pinset, calls `pin ls --type recursive` on IPFS to tell if things in the local pinset are pinned or not, and finally incorporates ongoing operations to the response.

For very big pinsets, the `pin ls` call is very expensive and, in particular, would not be necessary if we are filtering for things like `queued` or `pinning`, which affect ongoing operations only so the information could be obtained directly from the operation tracker.

Thus, doing the filtering directly on the pintracker would allow for big improvements for things like "show me all queued items" or "show me all items that are pinning".

hsanjuan:
This commit modifies the pintracker StatusAll call to take a status filter.

This allows to skip a PinLs call to ipfs when checking status for items that
are queued, pinning, unpinning or in error. Those status come directly from
the operation tracker. This should result in a significant performance
increase for those calls, particularly in nodes with several hundred thousand
pins and more, where the call to IPFS is very expensive.

A new TrackerStatusUnexpectedlyUnpinned status has been introduce to
differentiate between pin errors (tracked by the operation tracker) and "lost"
items (which before were pin errors too). This new status is handled by the
Recover() operation as before.

Fixes #1360 .

### DIFF
--- a/api/rest/restapi.go
+++ b/api/rest/restapi.go
@@ -891,30 +891,6 @@ func (api *API) allocationHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-// filterGlobalPinInfos takes a GlobalPinInfo slice and discards
-// any item in it which does not carry a PinInfo matching the
-// filter (OR-wise).
-func filterGlobalPinInfos(globalPinInfos []*types.GlobalPinInfo, filter types.TrackerStatus) []*types.GlobalPinInfo {
-	if filter == types.TrackerStatusUndefined {
-		return globalPinInfos
-	}
-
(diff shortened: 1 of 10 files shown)


hsanjuan:
Note: this is an RPC-breaking change. StatusAll calls will no longer work among peers with different versions.

PR #1428, opened 2021-07-28
Fix data races

iand:
Fixes three data races caused by incorrect or missing locks. See commits for individual details.


Stebalien (approved):

hsanjuan:
Thankity thanks

Issue #1430, opened 2021-07-30
Feature request: ipfs-cluster-service that acts as consensus arbiter


etam:
**Describe the feature you are proposing**

Allow to run a ipfs-cluster-service, that
- Is part of the cluster and runs the "consensus component"
- Doesn't take part in pinning the data and doesn't need an ipfs node running

Why:
- Running additional raft nodes allows for better resilience against failures
- An arbiter could be run on small devices (like Raspberry Pi), which are too small for ipfs node.

**Additional context**

https://docs.gluster.org/en/latest/Administrator-Guide/Split-brain-and-ways-to-deal-with-it/#2-arbiter-volume

lanzafame:
> Doesn't take part in pinning the data and doesn't need an ipfs node running

Currently, the `pintracker` component actually uses the underlying IPFS node as storage for pinset metadata, so this isn't possible. 

A possible option, and I will ask that @hsanjuan confirm this is will work, is that you could run a node that has a really small IPFS RepoSize, just make sure that it is bigger than the predicted size of the pinset. This way the node will always be last in the list of nodes suggested by the `informer` for pinning the files to. This requires that the `informer` uses the `disk` `freespace` configuration. 

JFYI, you can comfortably run IPFS and IPFS Cluster on a Raspberry Pi.

hsanjuan:
A dummy IPFSConnector component that does nothing on pin/unpin and reports 0 size available would do, I think.

PR #1443, opened 2021-08-10
feat: add arbiter ipfsconn component

lanzafame:
This is a first pass at solving issue [#1430](https://github.com/ipfs/ipfs-cluster/issues/1430). @hsanjuan I have only done basic tests, `numpin` is incompatible with this dummy ipfshttp connector. I have left the config as is for now but a lot of it is not needed now, i.e. timeouts, disableunpin.


hsanjuan:
This should be a very simple mock of the ipfsconnector interface. Doesn't need config, most code can go away. Probably faster to write from scratch than to copy/paste+remove.

ipfscluster_test.go
```diff
@@ -59,6 +60,7 @@ var (
 
 	consensus = "crdt"
 	datastore = "badger"
+	connector = "arbiter"
```
lanzafame, line 63: @hsanjuan I will change this back to the normal ipfshttp connector, once you are happy with the PR.

hsanjuan:
I am not sure I understand this PR. The idea is to have a peer that does not really pin or anything, using a mock ipfsconn that does nothing. Why is it proxying things to the actual connector?

lanzafame:
@hsanjuan because other parts of cluster still rely on the connector interface? To clarify, are just suggesting a complete empty ipfsconnector interface implementation?

PR #1444, opened 2021-08-11
Change --wait behaviour to return on replication-factor-min

### LINKED ISSUE #1427: Relaxed --wait <number-of-confirmations> for use on CI
## Current behavior

Current behavior  of `pin add --wait` seems to be waiting for "pinned" status everywhere (no matter what replication factor was set? I tried to pass `--replication-min|max` and those seem to be ignored as well).

This effectively breaks CI build if any of the cluster peers is slow/unresponsive, and even when all of them are online, makes build unnecessarily long. 

In recent days, it was causing issues for both project teams and stewards (dist.ipfs.io).

I ended up not using `--wait` and [decided to poll for status in userland instead](https://github.com/ipfs/distributions/blob/3e9a8e209801f443586a647654aaf5fe9e39ca16/.github/workflows/main.yml#L263-L286) 

## Proposed  enhancement

I believe the most useful feature for CI would be to allow for "relaxed" `--wait` where we block only until "n" replica confirmations  are received. 

I propose we add a feature where a number of replica confirmations can be passed, and then we wait until that number of "pinned" confirmations is received:

```
ipfs-cluster-ctl pin add --wait 2 --replication 4 <CID>
```

The above would pin with replication factor 4 but wait only for 2 confirmations before unblocking CI job.

I imagine that when the number is missing, we would default to either   `--replication[-min]`  (if provided), and then fallback to number of peers when no replication is passed. 

### Alternative approach

Alternative is to make `--wait` just follow  implicit/explicit `--replication[-min]`, but I believe decoupling waiting from replication factor makes it more useful / flexible.

cc @olizilla @hsanjuan  @dholms (https://github.com/ipfs/ipfs-cluster/issues/1311)

hsanjuan:
This commit updates the WaitFor helper code and the --wait flag behaviour.

Currently, when waiting for a CID to be pinned, it would wait until all
statuses report Pinned or Remote. If a peer is lagging behind and shows
"unpinned" for whatever reason, the wait will get stuck even if the item was
correctly pinned in the places it needs to.

The logic is now re-worked to allow waiting until replication_factor_min is fulfilled.

Fixes #1427

### DIFF
--- a/api/rest/client/methods.go
+++ b/api/rest/client/methods.go
@@ -350,14 +350,21 @@ func (c *defaultClient) RepoGC(ctx context.Context, local bool) (*api.GlobalRepo
 	return &repoGC, err
 }
 
-// WaitFor is a utility function that allows for a caller to wait for a
-// particular status for a CID (as defined by StatusFilterParams).
-// It returns the final status for that CID and an error, if there was.
+// WaitFor is a utility function that allows for a caller to wait until a CID
+// status target is reached (as given in StatusFilterParams).
+// It returns the final status for that CID and an error, if there was one.
 //
-// WaitFor works by calling Status() repeatedly and checking that all
-// peers have transitioned to the target TrackerStatus or are Remote.
-// If an error of some type happens, WaitFor returns immediately with an
-// empty GlobalPinInfo.
+// WaitFor works by calling Status() repeatedly and checking that returned
+// peers have transitioned to the target TrackerStatus. It immediately returns
+// an error when the an error is among the statuses (and an empty
+// GlobalPinInfo).
+//
+// A special case exists for TrackerStatusPinned targets: in this case,
+// TrackerStatusRemote statuses are ignored, so WaitFor will return when
+// all Statuses are Pinned or Remote by default.
+//
+// The Limit parameter allows to specify finer-grained control to, for
+// example, only wait until a number of peers reaches a status.
 func WaitFor(ctx context.Context, c Client, fp StatusFilterParams) (*api.GlobalPinInfo, error) {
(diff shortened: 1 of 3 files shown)


api/rest/client/methods.go
```diff
@@ -354,10 +354,11 @@ func (c *defaultClient) RepoGC(ctx context.Context, local bool) (*api.GlobalRepo
...
 //
-// WaitFor works by calling Status() repeatedly and checking that all
-// peers have transitioned to the target TrackerStatus or are Remote.
-// If an error of some type happens, WaitFor returns immediately with an
-// empty GlobalPinInfo.
+// WaitFor works by calling Status() repeatedly and checking that all peers
+// have transitioned to the target TrackerStatus. When the target is
+// TrackerStatusPinned, TrackerStatusRemote will be considered as Pinned.
+// If an error of some type happens, WaitFor returns immediately with an empty
+// GlobalPinInfo.
```
hsanjuan, line 361: Fix desc.

api/rest/client/methods.go
```diff
@@ -463,18 +484,35 @@ func (sf *statusFilter) pollStatus(ctx context.Context, c Client, fp StatusFilte
...
+	if target == api.TrackerStatusPinned {
+		pinned := 0
+		for _, pinInfo := range gblPinInfo.PeerMap {
+			if pinInfo.Status == api.TrackerStatusPinned {
+				pinned++
 			}
-			return false, nil
+		}
+		return pinned >= rplMin, nil
+	}
```
hsanjuan, line 509: In the case of local, rplMin = 1. This will mean blocking until the local peer reports pinned.

PR #1462, opened 2021-08-28
Adder: try to make test less flaky

### LINKED ISSUE #1461: TestAdder_ContextCancelled seems to be flaky
**Additional information:**
 - OS: Linux (NixOS)
 - IPFS Cluster version: 0.14.1
 - Installation method: built from source

**Describe the bug:**

I tried updating the version of ipfs-cluster in Nixpkgs from 0.14.0 to 0.14.1: https://github.com/NixOS/nixpkgs/pull/134342. A bot built the new version and ran the test suite but the test TestAdder_ContextCancelled failed. The logs are in https://gist.github.com/r-rmcgibbo/4120129668b85b2bc51dc6f70c64b651. The bot built ipfs-cluster twice, for two different architectures and it failed once. I also can't reproduce the failure on any of my machines, so the failure seems to be non-deterministic. I noticed, that it failed before in https://github.com/ipfs/ipfs-cluster/pull/613#issue-235170657.

hsanjuan:
Fixes #1461

### DIFF
--- a/adder/adder_test.go
+++ b/adder/adder_test.go
@@ -95,7 +95,7 @@ func TestAdder_ContextCancelled(t *testing.T) {
 	sth := test.NewShardingTestHelper()
 	defer sth.Clean(t)
 
-	lg, closer := sth.GetRandFileReader(t, 50000) // 50 MB
+	lg, closer := sth.GetRandFileReader(t, 100000) // 50 MB
 	st := sth.GetTreeSerialFile(t)
 	defer closer.Close()
 	defer st.Close()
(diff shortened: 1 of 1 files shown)


Luflosi:
Instead of making it "less flaky", would it be possible to make it absolutely 100% reliable?

hsanjuan:
Everything is possible. But I don't have time for everything so I'm happy with the 99.9%.

Luflosi:
You should update the comment as well.

AMRo-Crypto (approved):

AMRo-Crypto:
- nice

PR #1468, opened 2021-09-13
Add tags informer and enable partition-based peer allocations for intelligent distribution

### LINKED ISSUE #159: IPFS Node Pining choice strategy
In some cases you would want to choice specific nodes to pin replication. Think the case you use JBOD and have 1 IPFS Daemon per disk, and also have multiple machines.
You would want to avoid ipfs cluster replicate a file in same machine diferent disks, you would want always to pin the file in diferent hosts and even maybe diferent region hosts.
I was looking into the documentation but I dind't found anything and could be awesome, because I am afraid to use a overlayFS over multiple drives as a single IPFS Repository, if one disk fail you could lose all the reopsitory located in the machine and you will need to redownload it again. In the case I say, you should only need to redownload the failed drive.

hsanjuan:
This PR fixes #159 .

It does it with two new things:

* There is now a `tags` informer. This allows users to manually provide metrics that they want their peer to broadcast just like it broadcasts "freespace" metrics. Thus, we are able to broadcast metrics like "region: europe" along with "freespace: 10GB".

* There is a new "metrics" allocator. This allocator can be configured to "group" things that have the same metric values. Thus, I can choose to group peers by whatever tags I defined and in whatever order (region, availability zone, <insert anything>). Once things are partitioned, it will make a priority list from peers by never taking two peers from the same place (if possible). So for example, if there are peers in several regions, the two first allocations will never be from the same region. Ultimately, it can choose by freespace. This is configured with an `allocate_by` allocator option that looks like `["region", "az", "freespace"]`, for example.

This required an amount of refactoring around informers/allocator components. Only peers that have all the necessary metrics per allocate_by will be considered. By default, nothing changes: it will still just allocate by the freespace metric.


cluster.go
```diff
@@ -282,28 +283,48 @@ func (c *Cluster) watchPinset() {
...
+	if len(metrics) == 0 {
+		logger.Errorf("informer %s produced no metrics", informer.Name())
+		return minTTL, nil
+	}
+
+	for _, metric := range metrics {
+		metric.Peer = c.id
+		ttl := metric.GetTTL()
+		if minTTL == 0 {
+			minTTL = metric.GetTTL()
```
hsanjuan, line 302: minTTL = ttl

hsanjuan: rewrite 

if ttl > 0 && (ttl < minTTL || minTTL = 0) {
      minTTL = ttl
}

cluster.go
```diff
@@ -282,28 +283,48 @@ func (c *Cluster) watchPinset() {
 	}
 }
 
-func (c *Cluster) sendInformerMetric(ctx context.Context, informer Informer) (*api.Metric, error) {
+func (c *Cluster) sendInformerMetrics(ctx context.Context, informer Informer) (time.Duration, error) {
```
hsanjuan, line 286: comment that returns the smallest ttl from the metrics pushed by this informer.

cluster.go
```diff
@@ -331,20 +352,24 @@ func (c *Cluster) pushInformerMetrics(ctx context.Context, informer Informer) {
 			// wait
 		}
 
-		metric, err := c.sendInformerMetric(ctx, informer)
+		minTTL, err := c.sendInformerMetrics(ctx, informer)
+		if minTTL == 0 {
+			logger.Errorf("informer %s reported metric ttl 0. This must be a bug. Aborting this informer", informer.Name())
```
hsanjuan, line 357: Is it a bug though? What happens with invalid disk metrics?

hsanjuan: Let's just wait 15 secs and try again?

informer/disk/disk.go
```diff
@@ -74,9 +92,9 @@ func (disk *Informer) Shutdown(ctx context.Context) error {
 	return nil
 }
 
-// GetMetric returns the metric obtained by this
-// Informer.
-func (disk *Informer) GetMetric(ctx context.Context) *api.Metric {
+// GetMetrics returns the metric obtained by this Informer. It must always
+// return at least one metric.
```
hsanjuan, line 96: write this in the interface definition.

informer/disk/disk.go
```diff
@@ -85,10 +103,10 @@ func (disk *Informer) GetMetric(ctx context.Context) *api.Metric {
 	disk.mu.Unlock()
 
 	if rpcClient == nil {
-		return &api.Metric{
+		return []*api.Metric{&api.Metric{
```
hsanjuan, line 106: Metric is expired. How did the pusher handle this before without resetting the timer to 0?

informer/numpin/numpin.go
```diff
@@ -56,17 +62,17 @@ func (npi *Informer) Name() string {
...
+// GetMetrics contacts the IPFSConnector component and requests the `pin ls`
+// command. We return the number of pins in IPFS. It must always return at
+// least one metric.
+func (npi *Informer) GetMetrics(ctx context.Context) []*api.Metric {
 	ctx, span := trace.StartSpan(ctx, "informer/numpin/GetMetric")
 	defer span.End()
 
 	if npi.rpcClient == nil {
-		return &api.Metric{
+		return []*api.Metric{&api.Metric{
```
hsanjuan, line 73: metric expired.

informer/tags/tags.go
```diff
@@ -0,0 +1,102 @@
...
+	for k := range cfg.Tags {
+		metrics.RegisterInformer("tag:"+k, sorter.SortText, true)
+	}
+
+	return &Informer{
+		config: cfg,
+	}, nil
+}
+
+// Name returns the name of the metric issued by this informer.
```
hsanjuan, line 49: Returns the name of the informer. It actually issues more custom metric names.

informer/tags/tags.go
```diff
@@ -0,0 +1,102 @@
...
+	tags.mu.Lock()
+	defer tags.mu.Unlock()
+	tags.rpcClient = c
+}
+
+// Shutdown is called on cluster shutdown. We just invalidate
+// any metrics from this point.
+func (tags *Informer) Shutdown(ctx context.Context) error {
+	_, span := trace.StartSpan(ctx, "informer/tags/Shutdown")
+	defer span.End()
```
hsanjuan, line 66: span remove

hsanjuan (commented):
self-review

hsanjuan:
One shortcoming here is that the "tags" are ordered alphabetically, and not by the sum of the freespace at the bottom. To order by this we will need to assign weights as we partition the metrics, and then do a last pass ordering the partitions by weight (rather than pre-ordering?).

Otherwise, if replication factor = 1, then something will always get allocated to the same region regardless of the free-space usage below.

informer/disk/disk.go
hsanjuan: A bug probably?

PR #1471, opened 2021-09-15
API: Refactor REST API to enable re-use of functionality.

hsanjuan:
This is a preparatory PR to add additional APIs (Pinning Service API) easily
to cluster.

Instead of copy-pasting most of what the REST API does, I have refactored so
that the whole configuration, routing and request-handling utilities can be
re-used.

The worst part has been to divide the test between tests that test core
(common.API) functionality and tests that test specific REST API endpoint
functionality. I could not get away without an additional common/test package
to provide functions that are used from both places. This is a side effect of
testing both http and libp2p endpoints for every request etc.


api/common/api.go
```diff
@@ -0,0 +1,622 @@
...
+	// no HTTPListenAddr, nor libp2p configuration fields, nor a libp2p
+	// Host are provided.
+	ErrNoEndpointsEnabled = errors.New("neither the libp2p nor the HTTP endpoints are enabled")
+
+	// ErrHTTPEndpointNotEnabled is returned when trying to perform
+	// operations that rely on the HTTPEndpoint but it is disabled.
+	ErrHTTPEndpointNotEnabled = errors.New("the HTTP endpoint is not enabled")
+)
+
+// When passed to sendResponse, it will figure out which http status
```
hsanjuan, line 69: SendResponse

api/rest/restapi.go
```diff
@@ -785,13 +360,13 @@ func (api *API) peerRemoveHandler(w http.ResponseWriter, r *http.Request) {
...
 		)
-		api.sendResponse(w, autoStatus, err, nil)
+		api.SendResponse(w, common.SetStatusAutomatically, err, nil)
 	}
 }
 
 func (api *API) pinHandler(w http.ResponseWriter, r *http.Request) {
-	if pin := api.parseCidOrError(w, r); pin != nil {
-		logger.Debugf("rest api pinHandler: %s", pin.Cid)
+	if pin := api.API.ParseCidOrFail(w, r); pin != nil {
```
hsanjuan, line 368: Can't I just `api.ParseCidOrFail` ?

Issue #1212, opened 2020-08-17
Status information for in-flight pins does not set the name field


hsanjuan:
The status objects include the pin names now, however this is not the case for pins that are currently being tracked (pinning, queued).

This is an issue in the `pintracker` module. The `optracker` module returns `PinInfo` objects that do not have the name set. This should probably be added as soon as they are tracked by checking the state (or when they are returned from the optracker).

omkarprabhu-98:
we can get the pin name through the operation, add it in the PinInfo returned it here? please correct me if I missed something
https://github.com/ipfs/ipfs-cluster/blob/78b3111ed688bb88a16dfcd6f955bdceb91c89c0/pintracker/optracker/operationtracker.go#L155

hsanjuan:
Probably, not 100% sure that there are no more subtleties involved. It would need a PinName() method with a read lock like Cid().

Issue #939, opened 2019-10-12
Monitor: make accrual detection useful (or remove)


hsanjuan:
**Describe the feature you are proposing**

Accrual failure detection exists in the `monitor` module. However, two metrics coming unusually close (like in a re-start), would skew the interval between metrics and then trigger alerts because the next one would come at a normal interval.

As a quick mitigation, safeguards were added to never alert if the last metric was still valid per it's TTL. But since metrics are valid 2x the interval in which they are produced, this means that accrual detection only kicks in on a full degradation (a metric is not only coming in a bit late, but very late), and just alerts right away. The edge cases in which it does not alert right away are very narrow (and honestly, at that point there should be an alert!).

In practice, metrics validity should not depend on their hardcoded TTL but rely on accrual detection which effectively handles some degradation. If a peer is re-started it could send metrics with a new epoch or in a way that accrual can detach from previous series.

In short, accrual is a complicated thing and we have it for no value right now, we should take full advantage of it or drop it as useless complexity if it turns out it can't really help: it should reduce the number of false positives when making alert decisions purely based on TTL (and not add more false positives).

kishansagathiya:
We could also ignore metrics received at anomalous intervals from the distribution

hsanjuan:
> We could also ignore metrics received at anomalous intervals from the distribution

No, metrics are sent for a reason. For example, to inform of a new freespace value when the storage is decreasing rapidly.

kishansagathiya:
I think that came out wrong. Let me make that clear.
Don't ignore the metric, but do not count it in the distribution if it came at an anomalous interval. (sort of like how data scientists clear the data before running models on them)

PR #1556, opened 2022-01-28
Pintracker improvements

### LINKED ISSUE #1554: Expose PeerId of IPFS node pinning a given CID

Calling `HTTP GET /pins/{cid}` will give us a Map keyed by PeerId of the cluster and its metadata as value. However, if someone wants to parse this response and establish a direct connection with the Peer pinning content it is not possible unless we do a follow up request to the Cluster ( `HTTP GET /peers` ) and get the PeerId from there.

I think it would be important information to be provided in the `/pins/{cid}` route. Probably a property within the response object as `ipfsPeerId`. This would easily unblock issues like this: https://github.com/web3-storage/web3.storage/issues/414

The alternative solution for us is to perform a follow up request per upload, which is not great.

**Additional context**

N/A

cc @hsanjuan

hsanjuan:
Adds IPFS field to pin status containing the IPFS Peer ID and fixes #1554, #1212 , #939 along the way.


test/rpc_api_mock.go
```diff
@@ -52,7 +52,11 @@ func NewMockRPCClientWithHost(t testing.TB, h host.Host) *rpc.Client {
 	if err != nil {
 		t.Fatal(err)
 	}
-	err = s.RegisterName("PeerMonitor", &mockPeerMonitor{})
+	var pid peer.ID
+	if h != nil {
+		pid = h.ID()
+	}
+	err = s.RegisterName("PeerMonitor", &mockPeerMonitor{pid: pid})
```
hsanjuan, line 59: we don't need this in the peermonitor anymore.

test/rpc_api_mock.go
```diff
@@ -64,7 +68,9 @@ type mockCluster struct{}
 type mockPinTracker struct{}
 type mockIPFSConnector struct{}
 type mockConsensus struct{}
-type mockPeerMonitor struct{}
+type mockPeerMonitor struct {
+	pid peer.ID
```
hsanjuan, line 72: remove

hsanjuan:
The `failure_threshold` option for the monitor dissappears with this PR.

PR #1607, opened 2022-03-22
Pintracker: streaming methods

hsanjuan:
This commit continues the work of taking advantage of the streaming
capabilities in go-libp2p-gorpc by improving the ipfsconnector and pintracker
components.

StatusAll and RecoverAll methods are now streaming methods, with the REST API
output changing accordingly to produce a stream of GlobalPinInfos rather than
a json array.

pin/ls request to the ipfs daemon now use ?stream=true and avoid having to
load the full pinset map on memory. StatusAllLocal and RecoverAllLocal
requests to the pin tracker stream all the way and no longer store the full
pinset, and the full PinInfo status slice before sending it out.

We have additionally switched to a pattern where streaming methods receive the
channel as an argument, allowing the caller to decide on whether to launch a
goroutine, do buffering etc.

### DIFF
--- a/api/common/api.go
+++ b/api/common/api.go
@@ -54,6 +54,9 @@ func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
+// StreamChannelSize is used to define buffer sizes for channels.
+const StreamChannelSize = 1024
+
 // Common errors
 var (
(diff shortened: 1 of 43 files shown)


hsanjuan (commented):
Reviewed

PR #1614, opened 2022-03-24
Adders: stream blocks to destinations

### LINKED ISSUE #810: Streaming blocks on /add

Blocks generated on the adding process should be streamed into ipfs on a single request.

**Additional context**

IPFS supports this since https://github.com/ipfs/go-ipfs/pull/6331.

This issue is blocked on RPC streaming.

hsanjuan:
This commit fixes #810 and adds block streaming to the final destinations when
adding. This should add major performance gains when adding data to clusters.

Before, everytime cluster issued a block, it was broadcasted individually to
all destinations (new libp2p stream), where it was block/put to IPFS (a single
block/put http roundtrip per block).

Now, blocks are streamed all the way from the adder module to the ipfs daemon,
by making every block as it arrives a single part in a multipart block/put
request.

Before, block-broadcast needed to wait for all destinations to finish in order
to process the next block. Now, buffers allow some destinations to be faster
than others while sending and receiving blocks.

Before, if a block put request failed to be broadcasted everywhere, an error
would happen at that moment.

Now, we keep streaming until the end and only then report any errors. The
operation succeeds as long as at least one stream finished successfully.

Errors block/putting to IPFS will not abort streams. Instead, subsequent
blocks are retried with a new request, although the method will return an
error when the stream finishes if there were errors at any point.

### DIFF
--- a/add_test.go
+++ b/add_test.go
@@ -32,7 +32,7 @@ func TestAdd(t *testing.T) {
 		mfr, closer := sth.GetTreeMultiReader(t)
 		defer closer.Close()
 		r := multipart.NewReader(mfr, mfr.Boundary())
-		ci, err := clusters[0].AddFile(r, params)
+		ci, err := clusters[0].AddFile(context.Background(), r, params)
 		if err != nil {
 			t.Fatal(err)
 		}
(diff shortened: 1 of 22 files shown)


hsanjuan:
oh wow tests pass on first attempt.

hsanjuan:
A test of BlockStream in ipfs is missing.

PR #1664, opened 2022-05-05
Fix #1547: Store pin metadata in a sorted array

### LINKED ISSUE #1547: State exports can be deserialized in multiple ways
A state export file contain pins. Each pin object may potentially have a user-defined metadata map.

These maps can potentially be serialized in any order. When the pin is stored as part of a delta, the binary representation of it may differ randomnly. This causes the created CRDT deltas to have different CIDs.

This means that multiple peers do not obtain a consistent state when the same state file is imported.

The serialization of a Pin to protobuf should be reproducible and only occur in one way for equivalent pin objects.

hsanjuan:
This deprecates the Metadata protobuf map and starts serializing metadata as an
array that is always sorted by the metadata key.

This should resolve the issue that an state export file can be imported
resulting in two different CRDT dags because the binary representation of the
pin can arbitrarily change depending on how the keys in the map are listed,
and thus the CID of a delta is impacted.

So with this, a state export should result in exactly the same DAG regardless
of where it is imported.


api/types.go
```diff
@@ -1106,17 +1106,36 @@ func (pin Pin) ProtoMarshal() ([]byte, error) {
...
+
 	opts := &pb.PinOptions{
 		ReplicationFactorMin: int32(pin.ReplicationFactorMin),
 		ReplicationFactorMax: int32(pin.ReplicationFactorMax),
 		Name:                 pin.Name,
 		ShardSize:            pin.ShardSize,
-		Metadata:             pin.Metadata,
-		PinUpdate:            pin.PinUpdate.Bytes(),
-		ExpireAt:             expireAtProto,
+		// Metadata:             pin.Metadata,
```
hsanjuan, line 1132: Peers on older versions will not be able to see metadata  for pins submitted on newer versions.

PR #1667, opened 2022-05-07
improve language; fix typos; move breaking change under own headline

RubenKelevra:


RubenKelevra:
I also committed that 1.0.1 release under GitHub with the old version of the changelog, as the tag existed, but the release was missing. :)

hsanjuan (approved):

PR #1676, opened 2022-05-24
improve expire-in docs

RubenKelevra:
- Make is more clear which (the old or the new) pin is beeing flagged as to be expired on the `pin update` command.
- Adjust the other expire-in docs to match this change above.

### DIFF
--- a/cmd/ipfs-cluster-ctl/main.go
+++ b/cmd/ipfs-cluster-ctl/main.go
@@ -360,7 +360,7 @@ content.
 				},
 				cli.StringFlag{
 					Name:  "expire-in",
-					Usage: "Duration after which the pin should be unpinned automatically",
+					Usage: "Duration after which the pin will be automatically removed from the cluster-pinset again",
 				},
 				cli.StringSliceFlag{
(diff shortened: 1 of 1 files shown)


hsanjuan:
@RubenKelevra could you somehow rebase this?

hsanjuan:
@RubenKelevra there's a test that checks that ipfs-cluster-ctl --help outputs are less than 120 char, and it fails.

RubenKelevra:
@hsanjuan and? The console should normally just make a line break on the output at the end of the view.

hsanjuan:
We keep things under 120 char so that things are readable.

RubenKelevra:
@hsanjuan and? The console should normally just make a line break on the output at the end of the view. Have a look:

Screenshot_20220620_214754-1

RubenKelevra:
@hsanjuan I mean it's a table, is there a way to make a line break in the table printing itself aka create two lines for the right side?

Or should the text be shortened?

hsanjuan:
Try shortening the text...

PR #1704, opened 2022-06-17
PinSVC API: fix bugs and increase compliance

hsanjuan:
These changes fix a few bugs in the PinSVC API and have been discovered by
running the compliance tool at https://github.com/ipfs-shipyard/pinning-service-compliance.

In particular, the error objects did not respect the spec, filters and counts
were not done right. An additional PR will follow with fixes to the pintracker
because some pin status information was not correctly set.

### DIFF
--- a/api/pinsvcapi/config.go
+++ b/api/pinsvcapi/config.go
@@ -67,7 +67,10 @@ func NewConfig() *Config {
 	cfg.DefaultFunc = defaultFunc
 	cfg.APIErrorFunc = func(err error, status int) error {
 		return pinsvc.APIError{
-			Reason: err.Error(),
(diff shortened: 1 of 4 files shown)


api/pinsvcapi/config.go
```diff
@@ -67,7 +67,10 @@ func NewConfig() *Config {
 	cfg.DefaultFunc = defaultFunc
 	cfg.APIErrorFunc = func(err error, status int) error {
 		return pinsvc.APIError{
-			Reason: err.Error(),
+			Details: pinsvc.APIErrorDetails{
+				Reason:  err.Error(),
+				Details: err.Error(),
```
hsanjuan: Remove this and ensure details is "omitempty".

PR #1705, opened 2022-06-17
pintracker: fix status objects missing or having wrong fields

hsanjuan:
The operation tracker was not setting some fields correctly when producing
PinInfo objects. Additionally, recover operations were submitted with empty
pin objects, which resulted in the status for pins sent on recover operations
to be missing fields.


hsanjuan:
This has broken the `TestAttemptCountAndPriority` test, somehow.

PR #1719, opened 2022-06-21
crdt: Commit batches on shutdown

### LINKED ISSUE #1697: Commit batches on shutdown
When shutting down, we should do it orderly and first of all let the crdt component commit any open batches (and reject any new pins).

hsanjuan:
This attempt to commit any pending batches when the crdt component is being
shutudown. A commit should succeed if the new DAG node is created, heads are
replaced and broadcast.

The latest version of CRDT ensure that the datastore does not unnecessarily
gets marked as dirty when a broadcasted head cannot be fetched/processed, so
the side effect of publishing a head before shutting down should be under
control at least.

Fixes #1697

### DIFF
--- a/consensus/crdt/consensus.go
+++ b/consensus/crdt/consensus.go
@@ -58,8 +58,10 @@ type batchItem struct {
 // and remove pins from the Cluster shared state. It uses a CRDT-backed
 // implementation of go-datastore (go-ds-crdt).
 type Consensus struct {
-	ctx    context.Context
-	cancel context.CancelFunc
+	ctx            context.Context
+	cancel         context.CancelFunc
+	batchingCtx    context.Context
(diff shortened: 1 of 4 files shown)


consensus/crdt/consensus.go
```diff
@@ -330,9 +336,19 @@ func (css *Consensus) Shutdown(ctx context.Context) error {
 		logger.Debug("already shutdown")
 		return nil
 	}
+	css.shutdown = true
 
 	logger.Info("stopping Consensus component")
 
+	// Cancel
+	css.batchingCancel()
+	if css.config.batchingEnabled() {
```
hsanjuan, line 345: this does not really work, it is still racy, we would need to wait until the batch worker has existed to be sure nothing else will be committed.

Issue #1702, opened 2022-06-17
pinqueue and pinning metrics can get negative


hsanjuan:
These metrics that are recorded in the operation tracker somehow end up with negative values sometimes, which should not be possible I think:

https://github.com/ipfs-cluster/ipfs-cluster/blob/master/pintracker/optracker/operationtracker.go#L362-L365

RubenKelevra:
@hsanjuan Got a suspicion: Have a look at #1699 it looks like the `concurrency` is completely ignored atm, as there are 10 pins started even if I set it to `1` in the service file (as well as environmental variable).

This may be related.

PR #1729, opened 2022-07-04
Fix: operationtracker metrics go negative

hsanjuan:
By substracing 1 on every cancel we are double-counting.


hsanjuan:
Related to #1702 .

PR #1732, opened 2022-07-08
Fix: leaking goroutines on aborted /add requests

hsanjuan:
It has been observed that some peers have a growing number of goroutines,
usually stuck in go-libp2p-gorpc.MultiStream() function, which is waiting to
read items from the arguments channel.

We suspect this is due to aborted /add requests. In situations when the add
request is aborted or fails, Finalize() is never called and the blocks channel
stays open, so MultiStream() can never exit, and the BlockStreamer can never
stop streaming etc.

As a fix, we added the requirement to call Close() we stop using a
ClusterDAGService (error or not). This should ensure that the blocks channel
is always closed and not just on Finalize().


RubenKelevra:
I think I ran into this on one of my servers, which just got 4 GB of memory before. It mainly happens when QUIC is enabled. Maybe a 1 to 100 reduction when QUIC is disabled.

I thought its an issue in Kubo itself (still weird to use this name :D).

hsanjuan:
This is not related to QUIC memory usage.

PR #1736, opened 2022-08-15
fix panic: unaligned 64-bit atomic operation on Linux@armv7

cldy309:
According to https://pkg.go.dev/sync/atomic#pkg-note-BUG , the panic below might be produced by struct unaligned problem.


hsanjuan:
@cldy309 thanks! Looks good! Do you mind leaving a comment on those structs that those fields are there because of alignment?

Issue #1796, opened 2022-11-13
Crash on startup - signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xf13031


mitchds:
```
/usr/bin/ipfs-cluster-service daemon 
2022-11-13T17:31:16.106+0400	INFO	service	ipfs-cluster-service/daemon.go:49	Initializing. For verbose output run with "-l debug". Please wait...
2022-11-13T17:31:16.126+0400	INFO	badger	badger@v1.6.2/logger.go:46	All 0 tables opened in 0s

2022-11-13T17:31:16.126+0400	INFO	badger	badger@v1.6.2/logger.go:46	Replaying file id: 0 at offset: 0

2022-11-13T17:31:16.126+0400	INFO	badger	badger@v1.6.2/logger.go:46	Replay took: 5.866µs

2022-11-13T17:31:16.127+0400	INFO	service	ipfs-cluster-service/daemon.go:267	Datastore backend: badger
2022-11-13T17:31:16.161+0400	INFO	cluster	ipfs-cluster@v1.0.4/cluster.go:137	IPFS Cluster v1.0.4 listening on:
        /ip4/10.0.0.10/tcp/9096/p2p/12D3KooWGtzjXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
        /ip4/127.0.0.1/tcp/9096/p2p/12D3KooWGtzjjXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX


2022-11-13T17:31:16.831+0400	INFO	pinsvcapi	common/api.go:460	PINSVCAPI (HTTP): /ip4/127.0.0.1/tcp/9097
2022-11-13T17:31:16.831+0400	INFO	restapi	common/api.go:460	RESTAPI (HTTP): /ip4/127.0.0.1/tcp/9094
2022-11-13T17:31:16.831+0400	INFO	ipfsproxy	ipfsproxy/ipfsproxy.go:331	IPFS Proxy: /ip4/127.0.0.1/tcp/9095 -> /ip4/10.0.0.10/tcp/5001
2022-11-13T17:31:16.833+0400	INFO	crdt	go-ds-crdt@v0.3.9/set.go:122	Tombstones have bloomed: 0 tombs. Took: 170.972µs
[...]

PR #1840, opened 2023-01-27
Dependency upgrades

hsanjuan:


hsanjuan:
Main breakage due to #1796 , fix comes later.

Issue #1885, opened 2023-03-09
Missing newline in JSON stream from proxy API pin/ls


StrathCole:
**Additional information:**
 - OS: Linux
 - IPFS Cluster version: 1.0.6
 - Installation method: binary

**Describe the bug:**

When calling the proxy API for `pin/ls` on port 9095, the returned format is not a valid newline-separated JSON.
Entries are returned like:
`curl -X POST "http://127.0.0.1:9095/api/v0/pin/ls?type=recursive&stream=true"`
```
{"Cid":"QmRcYpbiapVNstWVWvLRgvkvWXm5zKHawMFy7cFpRYfMFz","Type":"recursive"}{"Cid":"QmcVrQjK7FLFzefChixfBdTaCqjdr5hPb5yp44m7NUti1m","Type":"recursive"}{"Cid":"QmUfyBTHzrdGCrnPnHf6Fxhw4CkFDUKSKNUW6PcjMUxdwh","Type":"recursive"}{"Cid":"QmWRnA8LATY96xK7kAL9mW7hoABHvNXSCcQmrtSUKmvLw5","Type":"recursive"}{"Cid":"QmeXqon4zfscwoRCeebCcMLRA7xrHM6bhF4xa8Fd5oZ9s5","Type":"recursive"}{"Cid":"QmTt3B3qe1PcjdVjNh7pWJb1hrotacNYoQn63aXcxS4JJJ","Type":"recursive"}
```

While the original API of the ipfs on port 5001 returns data like this:
`curl -X POST "http://127.0.0.1:5001/api/v0/pin/ls?type=recursive&stream=true"`
```
{"Cid":"QmRg2YqDzwUrCQYpcuU69YcJZ9X2j2WyixBRi1YVbzetcZ","Type":"recursive"}
{"Cid":"QmTVmr6QhX1kkZ7YYJR8qebxb5XdwfJNexwZTvXuiAWt7b","Type":"recursive"}
{"Cid":"QmVVZ9B3hyYgDHHQLr3b79nHd37udq3NfkxaUYiRV8vEm4","Type":"recursive"}
{"Cid":"QmdJmmC5iyU17vSduESHkiDG5SvPSy785papnKvdZRftSZ","Type":"recursive"}
{"Cid":"QmRxGPcA5SYk2bUYpkczGCM7uEVwY8YgNPD4cY4HG5HRXC","Type":"recursive"}
[...]

PR #1893, opened 2023-03-21
Add newline to json stream to pin/ls

StrathCole:
I am new to go, but for me this solves issue #1885 including missing pin status display in ipfs-go webui.

### DIFF
--- a/api/ipfsproxy/ipfsproxy.go
+++ b/api/ipfsproxy/ipfsproxy.go
@@ -476,7 +476,7 @@ func (proxy *Server) pinLsHandler(w http.ResponseWriter, r *http.Request) {
 				}
 				resBytes, _ := json.Marshal(ipinfo)
 				w.Write(resBytes)
-				w.Write("\n")
+				w.Write([]byte("\n"))
 			}
 			wg.Wait()
 			if err != nil {
--- a/api/ipfsproxy/ipfsproxy.go
+++ b/api/ipfsproxy/ipfsproxy.go
@@ -475,8 +475,8 @@ func (proxy *Server) pinLsHandler(w http.ResponseWriter, r *http.Request) {
 					Type: pin.Mode.ToIPFSPinStatus(),
 				}
 				resBytes, _ := json.Marshal(ipinfo)
-				resBytes = append(resBytes, "\n"...)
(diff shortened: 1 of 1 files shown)


hsanjuan (approved):

api/ipfsproxy/ipfsproxy.go
```diff
@@ -475,6 +475,7 @@ func (proxy *Server) pinLsHandler(w http.ResponseWriter, r *http.Request) {
 					Type: pin.Mode.ToIPFSPinStatus(),
 				}
 				resBytes, _ := json.Marshal(ipinfo)
+				resBytes = append(resBytes, "\n"...)
 				w.Write(resBytes)
```
hsanjuan, line 479:
```suggestion
				w.Write(resBytes)
				w.Write("\n")
```

hsanjuan: I think this is slightly better to avoid having to extend the resBytes slice if it's at capacity.

api/ipfsproxy/ipfsproxy.go
```diff
@@ -476,6 +476,7 @@ func (proxy *Server) pinLsHandler(w http.ResponseWriter, r *http.Request) {
 				}
 				resBytes, _ := json.Marshal(ipinfo)
 				w.Write(resBytes)
+				w.Write("\n")
```
hsanjuan, line 479:
```suggestion
				w.Write([]byte("\n"))
```

api/ipfsproxy/ipfsproxy.go
hsanjuan: yeah, sorry, editing in Github. Trying again.

hsanjuan: > Just thought it wouldn't matter here as it always only contains a single entry. But as said, I don't know much about go, yet.

I'm not 100% sure, but the slice might have the same size as the number of bytes of the json entry. Appending one byte to it might mean increasing capacity and causing a larger allocation. Not sure. It just feels more idiomatic this way.

StrathCole: Sure. For me this looks good now. :-)

PR #1898, opened 2023-03-31
Send readiness notification to systemd

max-privatevoid:
Uses the [sd_notify API](https://www.freedesktop.org/software/systemd/man/sd_notify.html) to send a readiness notification to systemd, suitable for `Type=notify` services.

Reuses the existing (previously indirect) dependency on `github.com/coreos/go-systemd`

### DIFF
--- a/cmd/ipfs-cluster-service/daemon.go
+++ b/cmd/ipfs-cluster-service/daemon.go
@@ -31,6 +31,7 @@ import (
 
 	ma "github.com/multiformats/go-multiaddr"
 
+	sddaemon "github.com/coreos/go-systemd/v22/daemon"
 	errors "github.com/pkg/errors"
 	cli "github.com/urfave/cli"
 )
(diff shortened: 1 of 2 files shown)


cluster.go
```diff
@@ -761,6 +762,7 @@ This might be due to one or several causes:
 	c.shutdownLock.Lock()
 	c.readyB = true
 	c.shutdownLock.Unlock()
+	daemon.SdNotify(false, daemon.SdNotifyReady)
```
hsanjuan, line 765: Hmm I don't think this is the best place, as cluster is supposed to work as a library included in some other daemon.

I think you should launch a goroutine that waits for "Ready()" here: https://github.com/ipfs-cluster/ipfs-cluster/blob/master/cmd/ipfs-cluster-service/daemon.go#L107

and do the notification then.

hsanjuan (changes requested):

cluster.go
max-privatevoid: I turned it into a goroutine as you suggested. Please let me know if my implementation looks appropriate, I'm not particularly experienced in Go development.

hsanjuan (approved):

cluster.go
hsanjuan: Yeah looks good to me now!

PR #1936, opened 2023-06-05
feat: update go-ipld-format and boxo to remove globals

aschmahmann:
This is related to ipfs/boxo#291 and depends on https://github.com/ipfs/go-ds-crdt/pull/201 and https://github.com/hsanjuan/ipfs-lite/pull/296


go.mod
```diff
@@ -17,20 +17,20 @@ require (
...
-	github.com/ipfs/go-ds-crdt v0.4.0
+	github.com/ipfs/go-ds-crdt v0.4.1-0.20230605203104-b879ef19ed1e
 	github.com/ipfs/go-ds-leveldb v0.5.0
 	github.com/ipfs/go-ds-pebble v0.2.3
 	github.com/ipfs/go-fs-lock v0.0.7
 	github.com/ipfs/go-ipfs-api v0.6.0
 	github.com/ipfs/go-ipfs-cmds v0.9.0
 	github.com/ipfs/go-ipld-cbor v0.0.6
-	github.com/ipfs/go-ipld-format v0.4.0
+	github.com/ipfs/go-ipld-format v0.4.1-0.20230530195241-c3da01c74a06
```
aschmahmann, line 33: TODO: use released versions

adder/sharding/dag.go
```diff
@@ -54,7 +45,7 @@ func CborDataToNode(raw []byte, format string) (ipld.Node, error) {
 	if err != nil {
 		return nil, err
 	}
-	shardNode, err := ipld.Decode(shardBlk)
+	shardNode, err := cbor.DecodeBlock(shardBlk)
```
aschmahmann, line 48: Looks like this only uses dag-cbor decoding so just went with directly decoding here

adder/adder.go
```diff
@@ -19,20 +19,33 @@ import (
...
-	ipld.Register(cid.DagCBOR, cbor.DecodeBlock)
+	mcReg := multicodec.Registry{}
+	mcReg.RegisterDecoder(cid.DagProtobuf, dagpb.Decode)
+	mcReg.RegisterDecoder(cid.Raw, raw.Decode)
+	mcReg.RegisterDecoder(cid.DagCBOR, dagcbor.Decode)
+	ls := cidlink.LinkSystemUsingMulticodecRegistry(mcReg)
+
+	ipldDecoder = ipldlegacy.NewDecoderWithLS(ls)
+	ipldDecoder.RegisterCodec(cid.DagProtobuf, dagpb.Type.PBNode, merkledag.ProtoNodeConverter)
+	ipldDecoder.RegisterCodec(cid.Raw, basicnode.Prototype.Bytes, merkledag.RawNodeConverter)
```
aschmahmann, line 48: Needs a decision on the type of registry to use here:
- go-ipld-prime codecs with wrappers for dag-pb + raw ( e.g. the initial proposal here )
- go-ipld-prime globals (e.g. including other codecs) with dag-pb + raw wrappers
- go-ipld-format only registry

cc @hsanjuan

hsanjuan: What do you recommend?

aschmahmann: I'm not sure how the performance of the codecs compares (e.g. the two dag-cbor codecs). I feel like I've heard both are better from different parties.

I'd stay away from global variables though, since less magic makes it more obvious what you do/don't support.

---

Whether you go with the go-ipld-format codecs and then add go-ipld-legacy wrappers for go-ipld-prime codecs, or use go-ipld-prime codecs with the couple necessary wrappers for dag-pb/raw doesn't make a big difference. I've heard people say the old codecs are faster, and I've heard the inverse too. Haven't had the motivation to benchmark and see though 🤷.

hsanjuan: I guess we can take what you did here then :shrug:

hsanjuan:
Including this in a separate PR with general dependency upgrades.

PR #2001, opened 2023-11-23
Fix #1999: Support talking to unix sockets in the ipfshttp connector

### LINKED ISSUE #1999: Connection to Kubo via Unix domain socket fails with HTTP error
**Additional information:**
 - OS: NixOS 23.05
 - IPFS Cluster version: 1.0.6
 - Installation method: NixOS repo

**Describe the bug:**

When setting `ipfs_connector.ipfshttp.node_multiaddr` to the Multiaddr of the Unix domain socket exposed by Kubo Nix package by default (`"/unix/run/ipfs.sock"`) IPFS Cluster Service appears to successfully connect to the socket but fails to make any HTTP requests instead printing the following error for each request:

> error posting to IPFS:Post "http:///run/ipfs.sock/api/v0/id": http: no Host in request URL

This is probably an encoding issue when generating the URL before passing it to whatever HTTP library is used by ipfs-cluster-service, since the encoded “request URL” should read as “http://%2Frun%2Fipfs.sock/api/v0/id” according to that URL RFC and hence have a hostname (even if unusual) set.

hsanjuan:
Kubo can listen on a unix socket instead of tcp. This allows talking to that in the same way that the API client already supported talking to a cluster daemon listening on unix sockets.

### DIFF
--- a/ipfsconn/ipfshttp/ipfshttp.go
+++ b/ipfsconn/ipfshttp/ipfshttp.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/ipfs-cluster/ipfs-cluster/api"
 	"github.com/ipfs-cluster/ipfs-cluster/observations"
+	"github.com/tv42/httpunix"
 
 	files "github.com/ipfs/boxo/files"
 	gopath "github.com/ipfs/boxo/path"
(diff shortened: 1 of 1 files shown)


ntninja:
Just built this and this version works as expected! Checked both having ipfs-cluster connect to `/unix/run/ipfs.sock` (NixOS default IPFS API address) and having ipfs-cluster expose its own API to `/unix/run/ipfs-cluster.sock` and connecting to that using `ipfs-cluster-ctl --host /unix/run/ipfs-cluster.sock id`. Number of “arbitrary remote control on localhost”-services just went down by 2!

Issue #1728, opened 2022-06-29
Leave cluster and remove all pins in the process


RubenKelevra:
**Describe the feature you are proposing**

It would be neat if there's a flag for ipfs-cluster-follow start it up and connect to the cluster, but then send a leave cluster message, remove all locally pins related to the cluster and then delete all files related to the cluster.

PR #2003, opened 2023-11-28
feat #1728: add stop command for ipfs-cluster-follow

loomts:


loomts:
I am a newbie of ipfs-cluster and interested in this project. So I pick up this issue to learn the basic knowledge of ipfs-cluster. Can anyone please help me review :D

cmd/ipfs-cluster-follow/commands.go
```diff
@@ -486,6 +486,42 @@ func listCmd(c *cli.Context) error {
...
+	}
+	cfgHelper.Manager().Shutdown()
+	mgr, err := cmdutils.NewStateManagerWithHelper(cfgHelper)
+	if err != nil {
+		return cli.Exit(errors.Wrap(err, "error creating state manager"), 1)
+	}
+	err = mgr.Clean()
+	if err != nil {
+		return cli.Exit(errors.Wrap(err, "error cleaning datastore, close ipfs-cluster follow daemon first"), 1)
+	}
```
hsanjuan, line 514: Clean datastore only if `cleanup`. `stop` should just cleanly shutdown the follower.

cmd/ipfs-cluster-follow/main.go
```diff
@@ -271,6 +278,25 @@ as obtained from the internal state on disk.
...
+This command stops following %s and leaves %s. It also removes any
+persisted consensus data, including the current pinset (state).
+
+The next start of this follower will be like a first start to all effects.
+`, clusterName, clusterName),
+				Action: stopCmd,
+				Flags: []cli.Flag{
+					&cli.BoolFlag{
+						Name:  "cleanup",
+						Usage: fmt.Sprintf("delete files in ~/.ipfs-cluster-follow/%s", clusterName),
```
hsanjuan, line 296: You would have to use `buildPaths()` or join `c.String("config")` with `clusterName` to obtain the right path. It may not be `~/.ipfs-cluster-follow` if it was customized.

cmd/ipfs-cluster-follow/commands.go
```diff
@@ -486,6 +486,42 @@ func listCmd(c *cli.Context) error {
...
+
+	cfgHelper, err := cmdutils.NewLoadedConfigHelper(
+		configPath,
+		identityPath,
+	)
+	if err != nil {
+		return cli.Exit(errors.Wrap(err, "error creating config helper"), 1)
+	}
+	cfgHelper.Manager().Shutdown()
```
hsanjuan, line 506: So this is just stopping the configuration manager, not the follower node itself, fyi

loomts:
Thanks for your tips!

1. I check `api-socket` file exist first, then `unpinEverything` and `killFollower`. As you said 

> For that to work IPFS would need to be running, and the command should fail cleanly if it is not, without doing anything.

if ipfs or ipfs-cluster-follow daemon all running, the `api-socket` will exist, so I just check this file for ensure IFPS daemon running.

2.  After referencing [signal issue](https://github.com/golang/go/issues/27613), the `killFollower` use `syscall.GetsockoptUcred`.

It run as expected in my wsl ubuntu, but I not really sure in MacOS.

:D

cmd/ipfs-cluster-follow/commands.go
```diff
@@ -486,6 +486,57 @@ func listCmd(c *cli.Context) error {
...
+		err = unpinEverything(absPath, clusterName, identityPath)
+		if err != nil {
+			return cli.Exit(errors.Wrap(err, "unpin error"), 1)
+		}
+		err = killFollower(absPath)
+		if err != nil {
+			return cli.Exit(errors.Wrap(err, "error killing follower peer(ipfs-cluster-follow daemon)"), 1)
+		}
+	} else {
+		return cli.Exit(errors.Wrap(err, "ipfs-cluster-follow or ipfs daemon not exists"), 1)
```
hsanjuan, line 511: This would be more like `ipfs-cluster-follow does not seem to be running for this cluster`. ?

cmd/ipfs-cluster-follow/commands.go
```diff
@@ -486,6 +486,57 @@ func listCmd(c *cli.Context) error {
...
+		return cli.Exit("", 1)
+	}
+
+	// unpin cluster files and kill ipfs-cluster-follow process if cluster and ipfs daemon exist
+	_, err := os.Stat(filepath.Join(absPath, "api-socket"))
+	if err == nil {
+		err = unpinEverything(absPath, clusterName, identityPath)
+		if err != nil {
+			return cli.Exit(errors.Wrap(err, "unpin error"), 1)
+		}
```
hsanjuan, line 505: I think unpin should be behind a flag.

- `ipfs-cluster-follow myCluster stop`   should just kill the cluster peer
- `ipfs-cluster-follow myCluster stop --unpin --cleanup` should perform the unpin and cleanup actions.

The reason is that I wouldn't expect a stop command to be destructive by default (deleting pins).

cmd/ipfs-cluster-follow/commands.go
```diff
@@ -486,6 +486,57 @@ func listCmd(c *cli.Context) error {
...
+		cfgHelper.Manager().Shutdown()
+		mgr, err := cmdutils.NewStateManagerWithHelper(cfgHelper)
+		if err != nil {
+			return cli.Exit(errors.Wrap(err, "error creating state manager"), 1)
+		}
+		err = mgr.Clean()
+		if err != nil {
+			return cli.Exit(errors.Wrap(err, "error cleaning datastore"), 1)
+		}
```
hsanjuan, line 531: If we are removing the whole configuration folder, we might as well not spend time doing `mgr.Clean()` and we can save all of this lines.

cmd/ipfs-cluster-follow/commands.go
```diff
@@ -486,6 +486,57 @@ func listCmd(c *cli.Context) error {
...
+		err = os.RemoveAll(absPath)
+		if err != nil {
+			return cli.Exit(errors.Wrap(err, "fail to cleanup"), 1)
```
hsanjuan, line 534:
```suggestion
			return cli.Exit(errors.Wrap(err, "deleting " + absPath), 1)
```

cmd/ipfs-cluster-follow/main.go
```diff
@@ -271,6 +279,28 @@ as obtained from the internal state on disk.
...
+			{
+				Name:      "stop",
+				Usage:     "stop to follow the cluster and unpin everything",
+				ArgsUsage: "",
+				Description: fmt.Sprintf(`
+
+This command stops following %s and leaves %s. It also removes any
+persisted consensus data, including the current pinset (state).
+
+The next start of this follower will be like a first start to all effects.
```
hsanjuan, line 291: Adjust accordingly (it just stops by default but can delete everything).

hsanjuan (changes requested):

cmd/ipfs-cluster-follow/commands.go
loomts: sound great

cmd/ipfs-cluster-follow/commands.go
loomts: fix, but maybe `ipfs-cluster-follow myCluster stop --cleanup` will not unpin everything but only delete myCluster folder?

loomts:
I made a little change to make stop command logic clearer.
1. check `unpin` flag, then unpin everything if ipfs-cluster-follow daemon running, otherwise it will return.
2. attempts to kill the ipfs-cluster-follow process. But if the process is not running, it will continue.
3. check `cleanup` flag, then remove myCluster folder.

BTW, why this PR not be merged in v1.10, ~maybe this feature is not so much useful?~

hsanjuan:
v1.10 is still not released. This is still on my todo but not P0.

cmd/ipfs-cluster-follow/commands.go
```diff
@@ -486,6 +486,43 @@ func listCmd(c *cli.Context) error {
...
+
+	absPath, _, identityPath := buildPaths(c, clusterName)
+	if !isInitialized(absPath) {
+		printNotInitialized(clusterName)
+		return cli.Exit("", 1)
+	}
+
+	// if unpin flag is set, check if the ipfs-cluster-follow is running and unpin everything
+	if c.Bool("unpin") {
+		if _, err := os.Stat(filepath.Join(absPath, "api-socket")); err == nil {
```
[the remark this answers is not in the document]
hsanjuan: I also think it would be worth to let getClient return the client `endp`, and have an `localEndpoint(endp)` function that extracts the path of the unix socket from the multiaddress, or errors when not using a unix socket (in which case we cannot kill the process).

cmd/ipfs-cluster-follow/main.go
```diff
@@ -306,6 +338,27 @@ func socketAddress(absPath, clusterName string) (multiaddr.Multiaddr, error) {
...
+	ucred, err := syscall.GetsockoptUcred(int(connFile.Fd()), syscall.SOL_SOCKET, syscall.SO_PEERCRED)
+	if err != nil {
+		return err
+	}
+	followerP, err := os.FindProcess(int(ucred.Pid))
+	if err != nil {
+		return err
+	}
+	return followerP.Signal(syscall.SIGTERM)
+}
```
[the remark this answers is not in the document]
hsanjuan: Assuming this indeed doesn't work on Windows, another option would be to have `ipfs-cluster-follow run` directly write its pid into a file, which can be read and terminated directly, without having to resort to socket tricks.

The pid-file creation code would be part of `cmdutils` to be able to re-use it. And ofc needs to be removed when the process dies.

hsanjuan (changes requested):

loomts:
Thanks for your useful feedback, I will reconsider and fix it within a week!

cmd/ipfs-cluster-follow/main.go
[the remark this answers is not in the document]
loomts: You're right, it's not work on Windows, and even we get the pid of ipfs-cluster-follow on Windows, maybe we also cannot gracefully terminate process on Windows. See 
- https://stackoverflow.com/questions/55092139/gracefully-terminate-a-process-on-windows/62733194#62733194
- https://github.com/golang/go/issues/6720#issuecomment-66087740
- https://github.com/iwdgo/sigintwindows

1. Windows cannot directly send SIGTERM signal to process like Unix-like Systems.
2. We can use `taskkill /PID <pid> /F` or below code to terminate process easily, but it will directly close the process and we cannot do anything before closing processes
```go
func KillFollower(pid int) error {
	handle, _ := syscall.OpenProcess(syscall.PROCESS_TERMINATE, false, uint32(pid))
	defer syscall.CloseHandle(handle)
	return syscall.TerminateProcess(handle, 0)
}
```
3. The one way to gracefully terminate process is use `GenerateConsoleCtrlEvent`, but we must start ipfs-cluster-follow with `REATE_NEW_PROCESS_GROUP` flag to get the pid of "console process group".
4. Maybe we can only add a self close function on `Client` interface bypassing limitations of the Windows.

loomts:
@hsanjuan idk how to gracefully terminate process on Windows, would you mind that add shutdown api to restapi.go only for enable ipfs-cluster-follow stop command on Windows?(it can also make other unix Systems' implementation more elegant, without having to resort to socket tricks

PR #2126, opened 2024-12-12
Fix the error message during the initialization of tags informer

ikasas:
I fixed the error message during the initialization of the tags informer in ipfs-cluster-service and ipfs-cluster-follow because it was the error message for the numpin informer.


hsanjuan (approved):

hsanjuan:
Many thanks!

PR #2273, opened 2025-12-17
chore: execute goimports to format the code

stringscut:
execute goimports to format the code


hsanjuan:
cherry-picked here because actions are not running and I don't seem to be able to trigger them: https://github.com/ipfs-cluster/ipfs-cluster/pull/2285

PR #2324, opened 2026-06-29
ci: attach binaries to github releases

lidel:
> [!NOTE]
> @hsanjuan, would appreciate your feedback when you have a moment, no rush. The motivation: we need to move off `dist.ipfs.tech` for distribution, and the most future-proof path looks like each project owning its own binary pipeline rather than relying on the shared one. This PR adds that for ipfs-cluster; happy to rework it if you see things differently.
>
> Draft until [download-ipfs-distribution-action](https://github.com/ipfs/download-ipfs-distribution-action/) ships its fix to fetch from GitHub releases first (with `dist.ipfs.tech` as fallback); merge after that lands.

## Problem

ipfs-cluster binaries are published only to `dist.ipfs.tech` (via `ipfs/distributions`); the GitHub Release itself carries none. Anything pulling binaries from the release finds nothing, so the v1.1.6 assets had to be uploaded by hand.

## Fix

- Add `.github/workflows/release-binaries.yml`: a `workflow_dispatch` job that cross-compiles `ipfs-cluster-ctl`, `ipfs-cluster-follow` and `ipfs-cluster-service` for every platform and uploads any archive missing from a release. Idempotent, gated to tags `>= v1.1.6`.
- Document the artifact set in `RELEASE.md` and swap the `ipfs/distributions` step for running this workflow.

Run it against a tag after creating the release; re-runs only fill gaps. It does not change `dist.ipfs.tech` publishing or fire on its own.

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -1,0 +1,107 @@
+name: Release Binaries
+
+# Run manually (workflow_dispatch) against a release tag. Cross-compiles ipfs-cluster-ctl,
+# ipfs-cluster-follow and ipfs-cluster-service for every platform and attaches any archive
+# missing from the GitHub Release. Idempotent: an asset that already exists (including one
+# uploaded by hand) is left untouched, so a re-run only fills gaps. See RELEASE.md.
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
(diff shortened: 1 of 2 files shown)


hsanjuan (approved):

hsanjuan:
Can't hurt to try on next release I guess

PR #2325, opened 2026-07-02
refactor: use the built-in max/min to simplify the code

alaningtrump:
In Go 1.21, the standard library includes built-in [max/min](https://pkg.go.dev/builtin@go1.21.0#max) function, which can greatly simplify the code.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -1369,11 +1369,7 @@ func (m *Metric) SetTTL(d time.Duration) {
 // GetTTL returns the time left before the Metric expires
 func (m Metric) GetTTL() time.Duration {
 	expDate := time.Unix(0, m.Expire)
-	ttl := time.Until(expDate)
-	if ttl < 0 {
-		ttl = 0
-	}
-	return ttl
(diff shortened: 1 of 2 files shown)


alaningtrump:
Hi  @lidel, this is a small Go cleanup now that the repo targets a modern Go version: two manual bounds checks are replaced with built-in min/max in Metric.GetTTL and the IPFS pin-origin limit, with no intended behavior change. 

Could you take a quick look when convenient?

lidel:
@alaningtrump thanks, but I won't be reviewing this, until Sep 30 I only handle security fixes, and after that my (and Shipyard's) maintenance ends entirely: https://ipshipyard.com/blog/2026-the-end-of-ipfs-at-shipyard/

A non-Shipyard maintainer will have to pick this up. Who that is remains an open question for many projects: https://discuss.ipfs.tech/t/what-happens-to-ipfs-maintenance-and-public-infrastructure-after-september-30/20324

Keeping this open so it's ready for them.

hsanjuan (approved):

